PR #129, opened 2015-01-14
home5

matjazonline:
Hi Jean,
now webpage can be reached on /home5 path.
I will add password field back in.

[Review on Reviewable](https://reviewable.io/reviews/binary-com/binary-static/129)


matjazonline:
page  /home5_3  (https://binary-static-index1.herokuapp.com/home5_3)

PR #160, opened 2015-02-09
Remove sparkline component, and any code that is calling it and styles associated with it

borisyankov:


raunakkathuria:
Closing this as we are not going to remove this.

PR #163, opened 2015-02-10
/home5 page

matjazonline:


src/templates/haml/home5/index.html.haml
```diff
@@ -0,0 +1,450 @@
+- stash(
+-   title => l('Online Trading platform for binary options on Forex, Indices, Commodities and Smart Indices'),
+-   next_link => '/why-us',
+-   createacc => l('Create free account now'),
+- );
+- my $lang = stash->{language};
+%script{:src=>'//use.typekit.net/bfu3ywb.js'}
```
raunakkathuria: we would need to remove this when we release as we already have typekit loaded on binary.com

src/templates/haml/home5/index.html.haml
```diff
@@ -0,0 +1,450 @@
...
+              wait = true;              // Prevent future invocations
+              setTimeout(function () {  // After a period of time
+                  wait = false;         // And allow future invocations
+              }, limit);
+          }
+      }
+  }
+  //var elems=null
+
+  var resetH=function(){
```
raunakkathuria: can we load javascript at end of content? else it will block the rendering..

ashkanx:
Issue with footer css. headers' color has been changes to #2a3052.
footer

src/templates/haml/home5/index.html.haml
```diff
@@ -0,0 +1,450 @@
...
+        //$(window).resize(throttle(resetH,500))
+      })
+    }
+  }
+  window._waitJq()
+
+  }())
+
+:css
+  .static_full .section:first-child{padding-bottom:0;}
```
raunakkathuria: we generally don't use css inside template but as this is only for ab testing we can keep and it will go away when we remove home5

raunakkathuria:
Can we update the text of create account button to "Create Free Account" as of the now the account form at top and bottom have inconsistent button design (as shown in first and second image)  
demo

raunakkathuria:
There is lots of white space in between quote and name. Can we increase box length of quote 
quote

raunakkathuria:
We need to compress the svg files, I will do it when we are ready to release it. Just adding a note so I will remember

src/templates/haml/home5/index.html.haml
borisyankov: We should actually embed the TypeKit code in the haml and .tt and remove it from GTM. I have done this in the new design so no rush I guess.

src/templates/haml/home5/index.html.haml
```diff
@@ -0,0 +1,594 @@
...
+                %input{:type => 'hidden', :name => 'l', :value => $lang}
+              %p{:style=>'margin:0px;padding:0px;line-height: 1.5em;font-weight: bold;'}= l('Get your free $10k virtual money account')
+        .bottom-dark-box.acc-input-security{:style=>'margin:0 0 0 -9px;overflow:hidden;bottom:0;padding: 0px 0;background:#15212d;font-size: 72%;width:100%;text-align:center;color:#fff'}
+          %p{style=>'color:inherit; padding: 1px 8px;line-height: 1.3em;font-size: 0.9em;'}
+            %img.acc-input-security-icon{:src=>url_for('images/pages/home5/security-w.svg'),:style=>'margin: -6px 5px 0;max-width:13px;'}
+            = l('we use highest security and privacy measures')
+  .seen-on.section-margin
+    .grd-container
+      .grd-grid-6.grd-hide-tablet.grd-hide-phablet.grd-hide-mobile{:style=>'padding-top:8px'}
+        %h4{:style=>'margin:53px 0 28px;line-height:1.7em;padding: 0 0px;'}= '1st in the market and trusted by over 800.000 users'
```
raunakkathuria: Can we replace '.' with ',' so 800.000 will be 800,000 as we follow this convention

raunakkathuria:
When we hover over 'Create free Account Now' button the hover color changes to transparent or orange in this case. Was it intentional?

PR #166, opened 2015-02-11
no attorney

kavehmz:


raunakkathuria:
We have some javascript also for granter_loginid in statement and portfolio pages. We need to delete that also..

PR #172, opened 2015-02-17
Add TypeKit to head to prevent FOUT

borisyankov:
This adds the TypeKit loading code to the header, as discussed with JY.

Raunak, I don't have write access to the GTM, but on this change going live, we need to remove the corresponding entry.


raunakkathuria:
I am merging it in, will remove the GTM tag after the release is done

PR #176, opened 2015-02-24
replace home2

ashkanx:


src/sass/common/footer.scss
```diff
@@ -3,7 +3,7 @@
     background: $BIN_BLUE;
     h1 {
       font-size: 80%;
-      text-transform: uppercase;
+      // text-transform: uppercase;
```
raunakkathuria: no need to comment the old code, you can remove that

src/sass/pages/home2.scss
```diff
@@ -1,52 +1,241 @@
...
-.xtopcta{width:80%} 
-.xtestimonial h1{padding:10px;}
-.xhead{padding:10px;}
-.xcta{position:initial;}
-.xcta .xreal{margin-right:0px;}
-.xcta .xreal span{font-weight: 700;border-radius:0;padding: 20px 20px; width:60%;margin:10px 0;}
-.xcta .xvirtual span{background-color:#2a3052;font-weight: 700;border-radius:0;padding: 20px 20px; width:60%}
-}
-
+.home2-content {
```
raunakkathuria: Why we have home2.scss? we need to remove it i think

raunakkathuria:
We need to compress png and svg files before we upload

borisyankov:
Raunak, make sure you don't compress .png files twice. If the difference in size after compressing is small, it might be a recompress. SVGs - I haven't found big difference in sizes there.

src/templates/haml/home4/index.html.haml
```diff
@@ -88,9 +88,9 @@
       .grd-grid-12
         .ystart
           .button.pjaxload.pjaxload-focus.login-button
-            %span Daftar Akun Demo
+            %span= l('Daftar akun demo')
```
raunakkathuria: we dont need to add this in translation as its in Indonesian

raunakkathuria:
Getting this in console ""Invalid App Id: Must be a number or numeric string representing the application id." for facebook id I think

PR #184, opened 2015-02-25
Remove buggy and soon not needed code for footer

borisyankov:
The code that is supposed to move the footer to the bottom of the screen is not currently working.
I have removed it.
There is a very simple approach with flexbox that will implement it in pure css with just a few lines of code.


raunakkathuria:
Can we release this individually cos the code was to make footer sticky at bottom?

borisyankov:
That is it. The code to make it sticky didn't work, and doing it via js was overkill anyways :)

PR #187, opened 2015-02-26
Add responsive behavior for main text and footer

borisyankov:
Moved from redesign fork to main project.


src/sass/media-queries.scss
```diff
@@ -1,26 +0,0 @@
...
-                              }
-                          }
-                      }
-                  }
-              }
-          }
-      }
-  }
-
-  #sell_popup_container {
```
raunakkathuria: have we fixed this somewhere else? It was added so that we show our sell popup as 100% on mobile and 7 -inch tablet

borisyankov: Yeah, none of the styles are needed but I have forgotten to include the non-media-query #sell_popup_container style. Will redo the PR, but apart from that, these are not needed.

raunakkathuria:
Can we instead of creating pull request directly to master create it for some other branch so that it can be tested on QA?

borisyankov:
It is in a branch on the main repo, so the 'responsive' branch can be tested on QA, I think?

PR #188, opened 2015-02-27
pjax-lib upgrade + all custom changes done to old version reapplied to new one

borisyankov:
Removed a 'indicate-loading' to not grey out the content while new one is loaded.

### DIFF
--- a/src/sass/common.scss
+++ b/src/sass/common.scss
@@ -179,7 +179,7 @@ select.input {
 label {
     line-height: 25px;
 }
-// used for form label in grid structure 
+// used for form label in grid structure
 .form_label {
     text-align: right;
     line-height: 25px;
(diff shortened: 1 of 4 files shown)


src/sass/common.scss
```diff
@@ -331,15 +331,6 @@ fieldset .error-msg {
     position: absolute;
     top: -9999px;
 }
-.indicate-loading {
```
raunakkathuria: for this we need to remove from js also src/javascript/binary/pages/bet/bet_form.js

borisyankov: Yeah, I did notice the several lines in bet_form.js and also in pjax.js but decided to leave them there if we decide to actually add some styling in the future. No harm.

raunakkathuria: Lets remove it completely if we need it in future we can reapply them back

PR #200, opened 2015-03-05
Refactor list styles, only one style remains

borisyankov:
The class name 'bullet' is kept, because it is universal, though the style is actually of a check mark.


raunakkathuria:
Why translations files are added in pull request?

borisyankov:
Because of the <ul class='checked'> line being included in them.
Should we rework the translation files or?

raunakkathuria:
Can you remove the translation file, as it is automated process they will added themselves

borisyankov:
Sure.

PR #201, opened 2015-03-06
Refactor text color.

borisyankov:
All random text colors are now one single color. As they should.


raunakkathuria:
We decided not to go with this and other pull request?

PR #223, opened 2015-03-18
Fix digit tick contracts processing, prevent extra ticks display

raunakkathuria:


src/javascript/binary/pages/bet/bet_price.js
```diff
@@ -225,6 +224,7 @@ var BetPrice = function() {
                                         if ($self.applicable_ticks.length === how_many_ticks) {
                                             $self.evaluate_digit_outcome();
                                             $self.reset();
+                                            break; // need to break the loop else it will keep on processing the extra tick
```
mwmiller: Nice!

PR #224, opened 2015-03-18
Implement a small fix to a fix, to actually fix 'Cannot read property '0' of undefined'

borisyankov:
|| is needed instead of && :)

### DIFF
--- a/src/javascript/binary/livechart.js
+++ b/src/javascript/binary/livechart.js
@@ -281,9 +281,10 @@ LiveChartTick.prototype.process_data = function(point) {
             epoch: parseInt(point[1]),
             quote: parseFloat(point[2])
         };
-        if (!this.chart || !this.chart.series) {
-            return;
-        }
+        
+        if (!this.chart) return;
+        if (!this.chart.series) return;
+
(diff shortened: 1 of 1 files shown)


src/javascript/binary/livechart.js
```diff
@@ -281,7 +281,7 @@ LiveChartTick.prototype.process_data = function(point) {
             epoch: parseInt(point[1]),
             quote: parseFloat(point[2])
         };
-        if (!this.chart && !this.chart.series) {
+        if (!this.chart || !this.chart.series) {
```
raunakkathuria: what difference it will make, i don't its better than before.. Say this.chart is defined and this.chart.series is not defined  then it will not return and this.chart.series[0] will give error then. And we use these conditions at lot of places we need to update them also

borisyankov: Totally better ;)
The nots may be confusing you, but currently if this.chart is undefined it
still continues evaluating second part and throws there. Or is the correct
operator. I've tested it.

jy-binary: Add a comment for future reference?

PR #235, opened 2015-03-24
home5 page to check translations

matjazonline:


src/templates/haml/home5/index.html.haml
```diff
@@ -104,11 +106,26 @@
...
+      }
+      .second-label{
+        margin:4px 0 0 0;
+      }
+      .home-content #banner{
+
+      }
 .home-content.paragraph-style
-  #banner.section-margin{:style=>'background-position-y: 68%;background-position-x: -272px;'}
+  #banner.section-margin{:style=>'background: url(../images/pages/home5/banner-background_1.png);background-repeat: no-repeat;background-position: 50% 0%;'}
```
raunakkathuria: I don't think url will work. We need to use url_for I think. Can you check if we have used url instead of url_for?

PR #245, opened 2015-04-14
sy/email_login_ui

shuwnyuantee:


config.json
```diff
@@ -2,7 +2,7 @@
     "website_name": "Binary.com",
     "display_languages": "EN,ID,RU,ES,FR,PT,DE,ZH_CN,PL,AR",
     "static": {
-        "url": "https://static.binary.com/",
+        "url": "https://shuwnyuantee.github.io/binary-static/",
```
raunakkathuria: need to make it static.binary.com ;)

src/javascript/binary/base/page.js
```diff
@@ -30,6 +30,33 @@ var gtm_data_layer_info = function() {
...
+        this.is_logged_in = true;
+
+        if(loginid_list !== null && typeof loginid_list !== "undefined") {
+            var loginid_array = [];
+            var loginids = loginid_list.split('+').sort();
+
+            for (var i = 0; i < loginids.length; i++) {
+                var real = 0;
+                var items = loginids[i].split(':');
+                if (items[1] == 'R') {
```
raunakkathuria: we need to use === (strong comparison) instead of == else test will fail

src/javascript/binary/base/page.js
```diff
@@ -335,8 +363,27 @@ Header.prototype = {
...
-        if (this.client.is_logged_in) {
-            $("#client_loginid").html(this.client.loginid);
+        if (this.user.is_logged_in && this.client.is_logged_in) {
+            var loginid_select;
+            var loginid_array = this.user.loginid_array;
+            for (var i=0;i<loginid_array.length;i++) {
+                var curr_loginid = loginid_array[i].id;
+                var real = loginid_array[i].real;
+                var selected = '';
+                if (curr_loginid == this.client.loginid) {
```
raunakkathuria: same here === instead of ==

config.json
shuwnyuantee: sure, will revert this before merge. Now still need it in testing :)

PR #254, opened 2015-04-24
Pull request to add new highcharts application in binary.com

arnabk:


src/templates/haml/charting/index.html.haml
```diff
@@ -31,3 +31,11 @@
       %h4
         %a.pjaxload{:href => url_for('/charting/livechart')}= l('Live Charts')
       %p= l('This is our \'next generation\' charting tool.')
+  .grd-grid-6.grd-parent
+    .grd-grid-8.grd-with-padding
+      %a.pjaxload{:href => url_for('//highcharts.binary.com'), :target => '_blank'}
```
raunakkathuria: for external link you can directly specify "//highcharts.binary.com" instead of url_for. url_for is used for binary.com url. I have fixed it, as of now it is broken on binary.com. If possible I will do a release

PR #256, opened 2015-04-28
Adding 4 new landing pages, remove home3/4/5

ashkanx:


raunakkathuria:
are these png images compressed?

borisyankov:
I think we doing it wrong ;)
We should have kept the .png and .svgs in the original state, and do the compression in the build process.

raunakkathuria:
@borisyankov ok then we will find grunt plugin for this so that only compressed images go to gh-pages

ashkanx:
Sorry Boris, Already compressed  :)
Raunak will talk to you about this

PR #257, opened 2015-04-29
template changes for corporate action and smart index

yngshan:


src/templates/toolkit/portfolio/tradecorporate_action.html.tt
```diff
@@ -0,0 +1,52 @@
+<p align="left"> The following table reflects the corporate actions that happened during contract period. </p>
```
raunakkathuria: add localize here also

PR #304, opened 2015-05-25
enable MF acc transfer

shuwnyuantee:


src/javascript/binary/base/page.js
```diff
@@ -629,7 +629,7 @@ Contents.prototype = {
                 }
 
                 // temporary only show for internal staff
```
jy-binary: remove the comment

shuwnyuantee: ok

PR #312, opened 2015-05-28
View balances link

shuwnyuantee:
- add "View balances" link
- homepage VR acc opening, js to throw error when both password are empty


src/sass/common/reality-check.scss
```diff
@@ -21,7 +21,7 @@
     background-color: #fff;
     border-radius:    .3em;
 }
-#reality-check > div > div > div:first-child {
+.lightbox > div > div > div:first-child {
```
tfoertsch: perhaps we now want to identify the elements inside the .lightbox > div > div box differently. lightbox is a class and hence can contain various things.

Maybe it's a good idea to give the parent of this box separate ids for both, reality-check and this balance box.

shuwnyuantee: Thanks, I added id to both the div in reality-check & balances. Recognise straight by these id is more straight forward.

src/templates/haml/user/balances.html.haml
```diff
@@ -0,0 +1,22 @@
...
+    %table
+      %tr
+        %th= l('Login ID')
+        %th= l('Currency')
+        %th= l('Balance')
+      - for my $balance (@{$client_balances}) {
+          %tr
+            %td= $balance->{loginid}
+            %td= $balance->{currency}
+            %td= $to_monetary_number_format->($balance->{balance})
```
arcofdescent: Its a good practice to always right align numbers, especially currency amounts.

shuwnyuantee: @arcofdescent Thanks for suggestion, will improve it this way in my next release

PR #322, opened 2015-06-09
add cashier withdrawal request template

raunakkathuria:


src/templates/toolkit/email/resend_verification.html.tt
```diff
@@ -1,6 +1,6 @@
 [% l("Dear Valued Customer,
 
-In order to verify your email address, please click on following link:
+In order to verify, please click on following link:
```
jy-binary: Please change this to:

"Please click on the following link:"

PR #346, opened 2015-07-07
remove logged into bo check

einhverfr:


einhverfr:
This doesn't look safe

PR #350, opened 2015-07-15
spread

junbon:


junbon:
- Added spreads display, streaming, and info update into javascript
- Added template for spreads button, analyse spread contract etc

kavehmz:
JB for future projects lets release in Steps.
There is no way that we can't do a step by step release.

src/templates/toolkit/bet/explanation/winning/spreads.html.tt
```diff
@@ -0,0 +1,6 @@
+<p> [% l('Spread contracts payout based on the movement of the index relative to the entry level of the contract.') %] </p>
+<p> [% l("'<strong>Long</strong>' contracts generate profits when the index rises and loses when it falls.") %] </p>
```
tfoertsch: ... and **losses** when it falls ...

PR #352, opened 2015-07-18
Pull request to integrate tradingview.binary.com

arnabk:


arnabk:
A preview could be viewed at http://arnabk-binary-static.heroku.com/

PR #364, opened 2015-07-29
for forgot password

fayland:


src/templates/haml/user/lost_password_email_sent.html.haml
```diff
@@ -2,5 +2,6 @@
 .grd-grid-12
   %h1= stash->{title}
   %p= l("Thank you. You're one step closer to being done!")
+  %p= l("You will receive an email shortly to the specified email address if your account exists with [_1]", stash->{website_name})
```
raunak-binary: add a full stop at end of statement

PR #375, opened 2015-08-04
New pages: Bulk Trade Facilities & Partners

ashkanx:


src/templates/haml/static/bulk_trader_facility.html.haml
```diff
@@ -0,0 +1,88 @@
...
+- );
+.grd-container.grd-default-container
+  .bulk-trader-facility.static_full
+    .grd-grid-3.grd-hide-mobile
+      .sidebar
+        %ul#sidebar-nav
+          %li#sidebar-link-section-1.selected
+            %a{:href => "#bulk-trader-facility"}= l('Bulk trader facility')
+          %li#sidebar-link-section-2
+            %a{:href => "#client-guide"}= l('[_1] client\'s guide', stash->{website_name})
```
raunakkathuria: better to use double quotes here like `l("[_1] client's guide", stash->{website_name})` so there is no need to put slash in front of single quote.

src/templates/haml/static/bulk_trader_facility.html.haml
```diff
@@ -0,0 +1,88 @@
...
+          %li#sidebar-link-section-3
+            %a{:href => "#account-manager-guide"}= l('Account manager\'s guide')
```
raunakkathuria: same here better to use double quotes

src/templates/haml/static/bulk_trader_facility.html.haml
```diff
@@ -0,0 +1,88 @@
...
+              %p!= l("<a href='#account-manager-guide'>Account managers</a> can now place trades for their clients with [_1]’s bulk trader facility.", stash->{website_name})
+              %p!= l("This is very helpful for <a href='#client-guide'>clients</a> who don’t always have enough time, or expertise, to follow market trends.")
+              %p!= l("It’s also a plus for account managers, who are paid a <a href='#''>commission</a> when they refer new clients to [_1].", stash->{website_name})
+            .clearfix
+      .section
+        %a#client-guide
+        .section-content
+          #section-2
+            .grd-grid-12
+              %h1= l('[_1] client\'s guide', stash->{website_name})
```
raunakkathuria: same double quotes thing

PR #388, opened 2015-08-12
New Footer: Add partner section

ashkanx:


kavehmz:
Why merge to binary-com:translations ?

PR #394, opened 2015-08-17
Japan changes

negar-binary:
hide things that clients in japan aren't supposed to see.


src/javascript/binary/pages.js
```diff
@@ -30,7 +30,11 @@ var trading_times_init = function() {
                   success: function(html){
                             trading_times.replaceWith(html);
                             trading_times = $("#trading-tabs");
-                            trading_times.tabs();
+			    if (page.language() === 'JA') {
```
jy-binary: can be done simpler?

src/templates/haml/static/tour.html.haml
```diff
@@ -49,7 +50,7 @@
         %h1= l('Study our live charts')
         %p= l('Frame your trading strategy using [_1]\'s powerful live charts. Instantly view the trading patterns of forex, indices, commodities and stocks.', stash->{website_name})
         %a.pjaxload.button{:href => url_for('/charting/livechart')}
-          %img.responsive#new-trader-icon{:src => url_for('images/pages/tour/live-chart-snapshot.svg')}
+          %img.responsive#new-trader-icon{:src => url_for('images/pages/tour/live-chart-jp-small.png')}
```
jy-binary: wrong name

src/javascript/binary/pages.js
negar-binary: @jy-binary we need to only display forex for trading times so all other tabs should be disabled for japan. not sure how else to do this.

src/templates/haml/static/tour.html.haml
negar-binary: @jy-binary sorry, did you mean the standard for file naming is wrong? because that's the name of the image we have newly added with Manros's help for our live chart in tour page

src/javascript/binary/pages.js
borisyankov: @negar-binary Have a look at how it will be done with pure css/sass here:
https://github.com/borisyankov/binary-static/blob/newdesign/src/sass/_access.scss

negar-binary: @borisyankov @jy-binary I tried adding a class in the html to hide tabs (something like <li [% IF market.tabid == 'forex' %] class='hide-ja' [% END %]), but without the javascript modification it doesn't work. this is the solution that raunak suggested and it seems to be the only thing that's working.

src/templates/haml/static/tour.html.haml
jy-binary: we don't need a japan-specific image here; no need for "jp-" in the file name.

PR #400, opened 2015-08-20
proper font-size for bet description, remove strong css

raunakkathuria:


src/sass/pages/trading.scss
```diff
@@ -391,9 +391,7 @@ a.spread_250, a.spread_260, a.spread_buy, a.spread_sell {
   width: 55.5%;
   float: right;
   padding: 0 3px 3px 0;
-  strong:first-child {
-      font-size: 130%;
-  }
+  font-size: 1.2em;
```
borisyankov: I'd say both approaches = not ideal :)
For the newdesign you would use $FONT_SIZE_L (to be one of the predefined sizes)

raunakkathuria: this is not for new design, its just for fixing current trading page when we remove strong from the description

PR #403, opened 2015-08-21
hide spread button on buy.

junbon:


src/javascript/binary/pages/bet/bet_price.js
```diff
@@ -84,8 +84,7 @@ var BetPrice = function() {
                     e.preventDefault();
                 }
                 var target = $(e.target);
-                var button = target.parents('a[class^="spread"]');
-                that.spread.disable(button);
+                that.hide_buy_buttons();
```
raunakkathuria: just want to verify that whether this is called on click of purchase button or before sending request?

junbon: This is called on click of purchase since the ajax request is in the buy_bet function.

PR #421, opened 2015-08-27
raunak/payout_change_fix

raunakkathuria:
- update description on fly rather than hiding price boxes

### DIFF
--- a/src/javascript/binary/pages/bet/bet_price.js
+++ b/src/javascript/binary/pages/bet/bet_price.js
@@ -623,7 +623,6 @@ var BetPrice = function() {
                     return prices;
                 },
                 prices_from_form: function () {
-                    
                     var prices = [],
                         order_forms = $('.orderform'),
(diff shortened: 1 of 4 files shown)


src/javascript/binary/pages/bet/bet_price.js
```diff
@@ -762,9 +760,17 @@ var BetPrice = function() {
...
-                            var payout_element = $('strong:first-child', $(this));
-                            payout_element.html(payout.value);
-                            price_moved(payout_element, old_payout.raw, payout.raw);
+                            var elm = $(this);
+                            if (elm) {
+                                var desc = elm.text();
+                                if (desc) {
+                                    desc = desc.trim();
+                                    if(/^([A-Z]{3}) \d+\.\d+/g.test(desc)) {
+                                        desc = desc.replace(/\d+\.\d+/g, payout.value);
```
mwmiller: Why are these `/g` modifiers here?  This seems exceptionally dangerous.

PR #426, opened 2015-08-28
Remove partnerapi page and links. Redirect to developers.binary.com

borisyankov:


jy-binary:
Please don't release this yet. v1 is the official WSAPI version at the moment; we shouldn't mention v2 anywhere until it is more stable

PR #435, opened 2015-09-03
Ashkan/Change the Font for Vietnamese language

ashkanx:


src/sass/language_base.scss
```diff
@@ -16,4 +18,8 @@
 		}
 	}
 }
+/*Vietnamese Language */
+.VI{
```
raunakkathuria: can we add body to this like `body.VI`, should be same for `JA` also

jy-binary:
Looks good. Please go ahead and launch the language too.

PR #447, opened 2015-09-08
Ashkan/new_home

ashkanx:
New home page after Negar's A/B testing


src/templates/haml/home/index.html.haml
```diff
@@ -38,34 +44,34 @@
...
+              %h3= l('Your funds are safe')
+              %p= l('With Binary.com, your money is secure. We segregate and hold client funds in secure AAA-rated banks.')
+        .grd-grid-6.grd-grid-mobile-12.grd-grid-phablet-12
+          %ul
+            %li.xlist3
+              %h3= l('Trade your way')
+              %p= l("Don't limit yourself! Whatever your market view, Binary.com allows you to put your strategy to work and profit from your predictions.")
+            %li.xlist4
+              %h3= l('Go with the leader')
+              %p= l('Binary.com is the award-winning industry pioneer in online options trading.')
```
raunakkathuria: just adding a note for future that instead of Binary.com it should be stash->{website}

PR #458, opened 2015-09-10
Negar/input validation

negar-binary:
validate the inputs on new trading page


src/javascript/binary/pages/trade/event.js
```diff
@@ -266,5 +270,6 @@ var highBarrierElement = document.getElementById('barrier_high');
 if (highBarrierElement) {
     highBarrierElement.addEventListener('input', debounce( function (e) {
         processPriceRequest();
+        submitForm(document.getElementById('websocket_form'));
```
borisyankov: These lines of code are repeating themselves (the whole ifs).
We should have them in a single function called few times.

borisyankov:
Turns out Safari and iOS do not show the error messages for the HTML5 input validation.
I tested and confirmed that. Need to think of a better/more consistent way to do it.
The :invalid, :valid, and :required pseudo-classes will provide a good way to style the inputs.

negar-binary:
@borisyankov ah :\ thanks for checking. will close this PR for now.

PR #463, opened 2015-09-11
Negar/flash

negar-binary:


jy-binary:
I see lots of .svg files added, but I don't see the flash files being deleted?

ashkanx:
we are still waiting for your approval before we remove all files:
A test page is here: 
https://www.binaryqa12.com/d/trade.cgi?market=forex&time=5t&form_name=risefall&expiry_type=duration&amount_type=payout&H=S0P&currency=USD&underlying_symbol=WLDAUD&amount=100&date_start=now&type=CALL&L=S0P&l=EN

borisyankov:
JY has a point. I think it is technically correct to delete the SVGs in this commit, since we are replacing them. The file deletion can always be reverted if it was an error.

PS. We need to do something about this font in the SVG, needs to use the page Roboto font.

ashkanx:
@borisyankov why delete SVGs? we are gonna replace all flash files with these SVGs.
Negar is still working on this branch. it's not ready to merge yet.

borisyankov:
I totally meant the SWFs :)

PR #472, opened 2015-09-14
Negar/input validation

negar-binary:
check validation of input for new trading page


borisyankov:
Looks good.

negar-binary:
@borisyankov ahh wait I need to change the barrier flexibility since in Randoms it needs a '+' at beginning but in for example forex GBP/USD the barriers don't need a '+' or '-'. can I create another PR for that one change too :) sorry. Wasn't expecting the merge so fast :P

PR #475, opened 2015-09-14
binary-maksym/japan-ui

binary-maksym:


src/javascript/binary_japan/contract.js
```diff
@@ -1,82 +1,108 @@
...
+            periods[currentCategory][currentContract['underlying_symbol']][period] = {
+                available_barriers: currentContract['available_barriers'],
+                barrier: currentContract['barrier'],
+                high_barrier: currentContract['high_barrier'],
+                low_barrier: currentContract['low_barrier'],
+                barriers: currentContract['barriers'],
+                date_start: currentContract['trading_period']['date_start'],
+                date_expiry: currentContract['trading_period']['date_expiry'],
+                duration: currentContract['trading_period']['duration']
+            };
```
borisyankov: Are you doing a full object copy here?
We should gradually start modernizing our code base. Object.assign is available in ES6, and we might start using it soon via polyfill.

src/javascript/binary_japan/symbols.js
```diff
@@ -0,0 +1,20 @@
+if(typeof is_japan === 'function'){
+	Symbols._details = Symbols.details.bind({});
```
borisyankov: Be careful with 'Symbols' name. We should reconsider the naming, since Symbol is a fundamental object in ES6 and we might mess stuff.

src/javascript/binary_japan/symbols.js
```diff
@@ -0,0 +1,20 @@
...
+
+	Object.defineProperties(Symbols,{
+		details:{
+			value:function(data){
+				var active_symbols = [];
+
+				data.active_symbols.forEach(function(symbol){
```
borisyankov: We've used the forEach + push in new array a lot in the past.
If you are not familiar, read on the map() method of Array (it is supported in ES5 thus all browsers we support)

src/javascript/binary_japan/symbols.js
```diff
@@ -0,0 +1,20 @@
...
+					if(symbol.market==='forex' && symbol.submarket==='major_pairs'){
+						active_symbols.push(symbol);
+					}
```
borisyankov: And we would have combined map() + filter() to achieve this in a more functional way.

src/javascript/binary_japan/contract.js
binary-maksym: Yep, because this object contains only one similar function. 
The other stuff is different

src/javascript/binary_japan/symbols.js
binary-maksym: I didnt invent this object. Just took from Raunak's code

src/javascript/binary_japan/symbols.js
binary-maksym: Nice) I am familiar to it's principles, Perl has the same function. Now will use it in JS

PR #504, opened 2015-09-29
binary-maksym/new-layout

binary-maksym:
Bugfixes
digits added
purchase window update


binary-maksym:
digits added
purchase window update

PR #513, opened 2015-10-03
binary-maksym/new-layout

binary-maksym:


mwmiller:
@borisyankov @mm-binary  We seriously need better history before these things get merged.  It will be impossible to understand what has happened and why.

PR #524, opened 2015-10-06
binary-maksym/new-layout

binary-maksym:
Fixes
proposal fix


raunakkathuria:
just a note, next time please add description related to what changes are and keep different branch name

PR #529, opened 2015-10-07
binary-maksym/new-trading-page

binary-maksym:
1 Sort durations fix
2 Format account balance as per current format
3 Delete prediction field from japan ui
4 Add loading icon on intial load
5 Updated amount parameter as WS were changed
6 Market's selectbox fixed (mistakenly some submarkets were inactive)


src/javascript/binary/pages/trade/price.js
```diff
@@ -37,7 +37,7 @@ var Price = (function () {
             prediction = document.getElementById('prediction');
 
         if (payout && payout.value) {
-            proposal['amount_val'] = parseFloat(payout.value);
+            proposal['amount'] = parseFloat(payout.value);
```
raunakkathuria: these all have been fixed and is live in master. merge latest master as there will be conflicts

PR #538, opened 2015-10-08
binary-maksym/initial_logo_fix

binary-maksym:
https://trello.com/c/oZPQwuVA/108-initial-loading-logo-make-smaller-and-gray


borisyankov:
That is an incorrect approach.
We can just change a style or a class for the SVG to change it's color.
We shouldn't use two resources.

I am also questioning the whole idea of having the logo in gray ;)

binary-maksym:
It's JY's idea to have initial logo in grey color. It's impossible to change svg color while using it as background image

raunakkathuria:
Agree with @borisyankov. we will just make the logo little smaller

PR #561, opened 2015-10-13
binary-maksym/ie-fix

binary-maksym:


raunakkathuria:
just adding a note, don't do these types of fix for IE. We need to make flexbox work on it. It is supported by IE

binary-maksym:
It's menu bar hight. Not flexbox

PR #567, opened 2015-10-15
Negar/careers

negar-binary:
add pdf of employee handbook to careers page for download and fix padding of footer for Indonesian language


src/sass/pages/careers.scss
```diff
@@ -0,0 +1,25 @@
+.career {
+	.xquote {
+    	background-color: #F2F2F2;
+    	padding: 20px 20px;
+    	margin: 30px 0;
+		div {
```
raunak-binary: better to keep consistent tab space as sass depends on intendation

src/templates/haml/static/careers.html.haml
```diff
@@ -23,10 +23,25 @@
...
+            %img.responsive{:src => url_for('images/pages/careers/corporate-culture.svg')}
+          %div.text
+            %a{:href => 'http://blog.wobb.my/project/binary/', :target => '_blank'}= l('Learn more about our corporate culture')
+    .grd-grid-6.grd-grid-mobile-12.grd-grid-phablet-6
+      %div.xquote
+        %div
+          %div.picture
+            %img.responsive{:src => url_for('images/pages/careers/handbook.svg')}
+          %div.text
+            %a{:href => url_for('/templates/haml/static/download/HB-employee-FA.pdf'), :download => true, :target => '_blank'}= ('Employee handbook')
```
raunak-binary: the localize tag is missing it should be `l('Employee handbook')`

src/sass/pages/careers.scss
negar-binary: strange. in my local file the tabs are correctly spaced.

src/templates/haml/static/careers.html.haml
negar-binary: ahh thanks :)

PR #575, opened 2015-10-19
Negar/input validation

negar-binary:
change input type to number and remove pattern as invalid message was incorrectly showing in IE
add new about us page (to replace Group Information; Group Information page can now be accessed through the link on the bottom of the new about us page)


src/templates/toolkit/bet/static.html.tt
```diff
@@ -149,8 +149,8 @@
                         <div class="row-inner big-col">
                            <div class="col-inner">
                               <select id="currency" class="small_width_input"></select>
-                              <input id="amount" type="text" value="10" class="medium_width_input" pattern="^([1-9]|[1-9][0-9]|[1-9][0-9][0-9]|[1-9][0-9][0-9][0-9]|[1-9][0-9][0-9][0-9][0-9]|[1][0][0][0][0][0])$" title="100,000 maximum"/>
-                              <input id="amount_per_point" type="text" value="1" class="medium_width_input" pattern="^([1-9]|[1-9][0-9]|[1-9][0-9][0-9]|[1-9][0-9][0-9][0-9]|[1-9][0-9][0-9][0-9][0-9]|[1][0][0][0][0][0])$" title="100,000 maximum"/>
+                              <input id="amount" type="number" value="10" class="medium_width_input" min="1" max="100000"/>
```
raunakkathuria: amount can be decimal as well, can you verify this from our current trading page

src/templates/toolkit/bet/static.html.tt
```diff
@@ -169,7 +169,7 @@
                                 <label for="stop_loss" id="stop_loss_label"></label>
                             </div>
                             <div class="big-col">
-                                <input id="stop_loss" type="text" value="10" pattern="^([1-9]|[1-9][0-9]|[1-9][0-9][0-9]|[1-9][0-9][0-9][0-9]|[1-9][0-9][0-9][0-9][0-9]|[1][0][0][0][0][0])$" title="100,000 maximum"/><label></label>
```
raunakkathuria: please check with quants whether stop_loss and stop_profit can be decimal, i think it can be

negar-binary:
@raunakkathuria you're right, thanks :)

src/templates/toolkit/bet/static.html.tt
```diff
@@ -95,7 +95,7 @@
                            </label>
                         </div>
                         <div class="big-col">
-                           <input id="barrier_high" type="text" name="H" pattern="^\+\d+(\.\d{1,3})?|\d+(\.\d{1,3})?$" title="0.001 format" />
+                           <input id="barrier_high" type="text" name="H" pattern="^\+\d+(\.\d{1,4})?|\d+(\.\d{1,4})?$" title="maximum of 4 decimal points allowed" />
```
raunakkathuria: if it means always starts with `+` then it may fail because sometimes we show absolute value as well

src/templates/toolkit/bet/static.html.tt
```diff
@@ -115,7 +115,7 @@
                            </label>
                         </div>
                         <div class="big-col">
-                           <input id="barrier_low" type="text" name="L" pattern="^\-\d+(\.\d{1,3})?|\d+(\.\d{1,3})?$" title="0.001 format" />
+                           <input id="barrier_low" type="text" name="L" pattern="^\-\d+(\.\d{1,4})?|\d+(\.\d{1,4})?$" title="maximum of 4 decimal points allowed" />
```
raunakkathuria: if it means always starts with `-` then it may fail because sometimes we show absolute value as well

PR #578, opened 2015-10-19
majid/portfoliows

majid-binary:
Not tested as for an unknown reason I cannot log in to binaryqa29.com. Will test when login issue is resolved.


src/javascript/binary/pages/portfiliows.js
```diff
@@ -0,0 +1,159 @@
...
+
+    var pageLanguage = page.language();
+    if(!pageLanguage) pageLanguage = "EN";
+
+    var init = function() {
+        TradeSocket.init({"thendo":"updateBalance"});
+    };
+
+    
+    var updateBallance = function(data) {
```
jy-binary: only one 'l' in "Balance"

src/javascript/binary/pages/portfiliows.js
```diff
@@ -0,0 +1,159 @@
...
+        $("#portfolio-table").removeClass("dynamic");
+
+    };
+
+
+    /*** utility functions ***/
+
+    // Until we can publish language specific json objects to each gh-pages branch
+    var uiTexts = {
+        "EN": {"view":"View"},
```
jy-binary: I don't think this should be done like this. All our language translations should be in the language files, otherwise we'll never remember which places translations are peppered around the code. Please check with Raunak how to refactor this.

src/javascript/binary/pages/portfiliows.js
```diff
@@ -0,0 +1,159 @@
...
...let-6">!longcode!</div><div class="grd-grid-5 grd-grid-mobile-5 grd-grid-phablet-4"><div class="grd-grid-4 grd-grid-mobile-12 grd-grid-phablet-12 right-aligned">!currency! <strong>!buy_price!</strong></div><div class="grd-hide grd-show-mobile grd-grid-mobile-12 grd-show-phablet grd-grid-phablet-12"><br>Indicative :</div><div class="grd-grid-4 grd-grid-mobile-12 grd-grid-phablet-12 right-aligned">!currency! 0.00</div><div class="grd-hide grd-show-mobile grd-grid-mobile-12 grd-show-phablet grd-grid-phablet-12"><br></div><div class="grd-grid-4 grd-grid-mobile-12 grd-grid-phablet-12 center-aligned"><div><span class="button"><button class="button open_contract_details" onclick="return false;" contract_id="5266199068" controller_action="sell" currency="!currency!" l="!lang-code!" payout="!payout!" purchase_price="!buy_price!" purchase_time="2015-10-02 08:24:13" qty="1" recomputed_price="0" sell_channel="P_!contract_type!-1443774253_!symbol!_1444424400_14958600_14065600_c_!currency!_!lang-code!" shortcode="!contract_type!_!symbol!_100_1443774253_1444424400F_14958600_14065600" url="https://www.binary.com/trade/analyse_contract?l=!lang-code!">:view:</button></span></div></div></div></div>';
```
jy-binary: why the hard-coded sample data like contract_id="5266199068" ?

src/javascript/binary/pages/portfiliows.js
majid-binary: Opps, will fix. Thanks.

src/javascript/binary/pages/portfiliows.js
majid-binary: I could not find somewhere this has been an implementation (for dynamic content). I raised this question [here](https://docs.google.com/a/regentmarkets.com/document/d/18tJEk6pxARb7Bm8ZF5Oc3jan7--FwmcwkCBMiLr5urs/edit?disco=AAAAAiIiI8U) too. So this is just a suggested method and sure there will be better methods. Once the language files become available in-page we could use that object in place of uiText.

src/javascript/binary/pages/portfiliows.js
majid-binary: The API (v2/v3) still does not return some required data. Once it does, the code should be updated to consume the new data. Until then, the View button would not be functional. For now, what is not returned has some dummy data hard-coded.

src/javascript/binary/pages/trade/message.js
```diff
@@ -6,31 +6,79 @@ var Message = (function () {
...
+
+                var thendo = "unknown";
+                if("passthrough" in response.echo_req && "thendo" in response.echo_req.passthrough) {
+                    thendo = response.echo_req.passthrough.thendo;
+                }
+
+                switch(thendo) {
+                    case "updateBalance":
+                        PortfolioWS.updateBallance(response);
+                        TradeSocket.send({"portfolio": 1}); // It seems the spawn property has been removed from the API (it is still in the Playground but even there it results in an error)
```
jy-binary: remove comments like this; if it's removed, then no need for a comment.

src/sass/pages/portfoliows.scss
```diff
@@ -0,0 +1,9 @@
+a.dynamic, span.dynamic, div.dynamic, p.dynamic, button.dynamic {
+	display: none!important;
+}
+
+span.loading:before {
+  content: "Loading ...";
```
jy-binary: translations handling?

majid-binary: Will remove it as the same loading effect used on https://www.binary.com/trading?l=EN will be used. In general I guess it would be to cumbersome to deal with translation in sass/css.

src/javascript/binary/pages/portfiliows.js
raunakkathuria: @majid-binary as explained. Please make use of text.localize

src/javascript/binary/pages/portfolio.js
```diff
@@ -147,7 +147,7 @@ pjax_config_page('portfolio', function() {
             $('#currencyfrom').keyup(function(event) { currencyConvertorCalculator(event.target); });
             $('#currencyfromvalue').change(function(event) { checkCurrencyAmountFormat(event.target); });
             $('#currencyfromvalue').keyup(function(event) { checkCurrencyAmountFormat(event.target); });
-            Portfolio.update_indicative_prices();
+            //Portfolio.update_indicative_prices();
```
raunakkathuria: use `proposal_open_contract` to get prices

src/sass/pages/portfoliows.scss
raunakkathuria: Please don't add any content in sass/css. They should be only used for stylings

config.json
```diff
@@ -1,6 +1,6 @@
 {
     "website_name": "Binary.com",
-    "display_languages": "EN,ID,RU,ES,FR,PT,DE,ZH_CN,PL,AR,ZH_TW,VI,IT",
+    "display_languages": "EN,ID,RU,ES,FR,PT,DE,ZH_CN,PL,AR,ZH_TW,VI",
```
jy-binary: why are you removing italian in this branch?

src/javascript/binary/pages/trade/message.js
```diff
@@ -3,35 +3,82 @@
...
+        
+        switch(msg_type) {
+
+            case "error":
+                $(".error-msg").text(response.error.message);
+                break;
+
+            case "authorize":
+
+                var thendo = "unknown";
```
jy-binary: what does "thendo" stand for?

src/javascript/binary/pages/portfiliows.js
```diff
@@ -0,0 +1,181 @@
...
+    if(!pageLanguage) pageLanguage = "EN";
+
+    var rowTemplate;
+    var retry;
+
+    var init = function() {
+        // get the row template and then discard the node as it has served its purpose
+        rowTemplate = $("#portfolio-dynamic tr:first")[0].outerHTML;
+        $("#portfolio-dynamic tr:first").remove();
+        TradeSocket.init({"thendo":"updateBalance"});
```
qingweibinary: There's a balance api , or is there any reason you're not using ? 

=)

majid-binary: Authorize does return a balance. Thought it would be better to use the data we have rather than requesting it again.

src/javascript/binary/pages/portfiliows.js
```diff
@@ -0,0 +1,181 @@
...
+
+    /**
+     * Amounts received from the API could be integer or decimal numbers.
+     * In case we have an integer, like 73, we want to display a decimal
+     * with two fractions, i.e. 73:00
+     * This function does that.
+     * Adapted from http://stackoverflow.com/a/14428340
+     * Kudos to: [VisioN](http://stackoverflow.com/users/1249581)
+    **/
+    var fixCurrency = function(n, c) {
```
qingweibinary: I've been using 

    Number(parseFloat(n)).toFixed(2)

does this work for you?

src/javascript/binary/pages/portfiliows.js
qingweibinary: I am not sure if it's a good idea to overload authorize api with other information, current implementation some what violates SRP, if later we want to remove the balance from authorize response, it will break your code

@raunak-binary @jy-binary 
any insights?

src/javascript/binary/pages/portfiliows.js
borisyankov: The actual code might be a bit more complicated.
Just .toFixed(2) is not enough. There is a 'pip' size value somewhere, that is specific to each asset type, if pip size is 0.0001 it should be .toFixed(4) if it it 0.01 it should be .toFixed(2). (pip = how many digits matter.

src/javascript/binary/pages/portfiliows.js
borisyankov: Not sure if it violates SRP, there might be valid reasons to return data that is available via other means. In this case I think it is rather random ;)
I too am using this value to initialize the balance value within binary-next-gen, but will redo it with balance streaming and guess no need to bother with this entry (and we should just take it out)

src/javascript/binary/pages/portfiliows.js
raunakkathuria: @borisyankov pip size is for barriers, if we check amount its always rounded off to two decimal places, i don't if its amount or barrier you are talking, Adding it as reference

borisyankov: Nope, I definitely mean for every value. If you ask me how much is the gold today, I need to take into account the pip. 
For example:
1 bitcoin is 443.91 USD
1 USD is 0.0023 Bitcoin

src/javascript/binary/pages/portfolio.js
majid-binary: Done. Thanks.

src/javascript/binary/pages/trade/message.js
majid-binary: Removed.

src/sass/pages/portfoliows.scss
majid-binary: Removed CSS :before content.

src/javascript/binary/pages/portfiliows.js
majid-binary: Used text.localize in later commits.

src/javascript/binary/pages/portfiliows.js
majid-binary: Removed hard-coded values and used call to proposal_open_contract to get the required values in the most recent commits.

config.json
majid-binary: This resulted from uploading an old version of config.json. Changed back in the most recent commit.

src/javascript/binary/pages/trade/message.js
majid-binary: This is the same use-case explained in response to the same question on my other PR here:
https://github.com/binary-com/binary-static/pull/597#discussion_r43570631

src/javascript/binary/pages/portfiliows.js
majid-binary: For this I will wait for when a consensus reached on this and then will make changes accordingly.

PR #580, opened 2015-10-20
Qingwei/view balance ws

qingweibinary:
The view balance popup are depending on Websocket data now, and responsiveness is added


src/javascript/binary/static/viewbalance/viewbalance.data.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * Created by qingwei on 19/10/2015.
```
borisyankov: What IDE are you using that adds this?
You should try Atom ;)

borisyankov:
Done with review. Looks good. :+1:

src/javascript/binary/static/viewbalance/viewbalance.data.js
qingweibinary: Webstorm, argh... I'll try disable this...

src/javascript/binary/static/viewbalance/viewbalance.ui.js
```diff
@@ -0,0 +1,58 @@
...
+
+        var $span = $("<span></span>", {class: "button"});
+        var $button = $("<button></button>", {
+            class: "button",
+            id: "close-balances",
+            text:Content.localize().textContinueTrading
+        });
+        $span.append($button);
+
+        $popupDiv.append($balTitle).append($br).append($table).append($span);
```
raunakkathuria: can't we use documentFragment here instead of appending multiple times, if jquery internally uses it then its fine

borisyankov: This is a good solution, as the actual DOM manipulation happens after it, and is not concerned with many elements.
I had to optimize these things before, but it is needed once you reach few thousands.
Ironically, passing a pure string representation of the html/dom is the fastest, but least maintainable approach.

qingweibinary: I do not think jquery use it underlying, but browser's js engine might do batch update for us (not guaranteed)

I'll change it

qingweibinary: @raunak-binary 

I was wrong, JQuery does use Document Fragment, when you passed in an array but not in loops, so not relevant in the code here

But I think if DocFragment helps because it delay reflow, a detached DOM (which is used here) should serve the same purpose, because reflow only happens when we append to the view

One of the answer here have similar opinion, what do you think?
http://stackoverflow.com/questions/14615484/does-jquery-use-create-document-fragment-inside-each-loops

I am reluctant to change as documentFragment API does not accept JQuery object out of the box, haha

borisyankov: I am totally fine with this solution. 
It is fast enough + maintainable. A good balance.

raunakkathuria: Ok no problem then

src/javascript/binary/static/viewbalance/viewbalance.data.js
```diff
@@ -0,0 +1,46 @@
+
```
qingweibinary: I am reimplementing the wheel, reason being that sharing TradeSocket is not a good idea because it get closed every now and then but balance needs to be always open once authorized, currently this is a temporary fix, would move to binary-live-api

ashkanx:
MLT & MF accounts balance aren't displayed
image

Here is the link to issue: https://github.com/binary-com/websockets/issues/22

src/config/locales/pt.po
```diff
@@ -13896,6 +13896,7 @@ msgstr "Prêmio de %1 %2 se %3 sair fora de %7 e %6 até %5 depois de %4."
 
 #~ msgid "Brazilian Index"
 #~ msgstr "Índice Brasileiro"
+>>>>>>> upstream/master
```
jy-binary: merge problem!!

qingweibinary: fixed

src/javascript/binary/pages/trade/content.js
```diff
@@ -71,7 +71,11 @@ var Content = (function () {
             textSaleDate: text.localize('Sale Date'),
             textSalePrice: text.localize('Sale Price'),
             textProfitLoss: text.localize('Profit/Loss'),
-            textTotalProfitLoss: text.localize('Total Profit/Loss')
+            textTotalProfitLoss: text.localize('Total Profit/Loss'),
+            textLoginID: text.localize('Login ID'),
+            textCurrency: text.localize('Currency'),
+            textBalances: text.localize('Balances'),
+            textContinueTrading: text.localize('Continue Trading')
```
jy-binary: have we not removed the lightbox? If so why do we still need the "Continue Trading" message?

qingweibinary: yes, will remove

PR #597, opened 2015-10-22
majid/passwordws

majid-binary:


src/javascript/binary/pages/client/change_passwordws.js
```diff
@@ -0,0 +1,130 @@
...
+	var bResult = true; 
+	var input_element_oldpassword = document.getElementById('oldpassword');
+	var error_element_erroroldpassword = document.getElementById('erroroldpassword');
+	document.getElementById('erroroldpassword').innerHTML = '';
+	if (input_element_oldpassword && error_element_erroroldpassword) {
+		var regexp;
+		var bInputResult = true;
+		regexp = new RegExp('^.+$');
+		if (bInputResult && !regexp.test(input_element_oldpassword.value)){
+			error_element_erroroldpassword.innerHTML = 'Please enter a password.';
```
jy-binary: translation handling?

majid-binary: Will be using the standard `text.localize('Text');` as used elsewhere on Binary.com pages. By the way, the content of line 41 downward is a copy of inline javascript (inside form's `onsubmit`) from [the existing password change page](https://www.binary.com/user/change_password?l=EN); and will be refactored.

src/javascript/binary/pages/client/change_passwordws.js
```diff
@@ -0,0 +1,144 @@
...
+		// Max length
+		if(new_pass.length > 25) {
+			$form.find("p[data-error='too-long']").removeClass("hidden");
+			isValid = false;
+		} else {
+			$form.find("p[data-error='too-long']").addClass("hidden");
+		}
+
+		// Invalid characters
+		var regexp = new RegExp('^[ -~]+$');
```
jy-binary: I think we should restrict the password to alphanumeric. Allowing weird characters is unnecessary. We already had a report in hackerone.com about a vulnerability involving passwords that included "

src/javascript/binary/pages/trade/message.js
```diff
@@ -3,33 +3,97 @@
...
+                }
+
+                // still unknown?
+                console.log("An error occured.", response);
+                break;
+
+            case "authorize":
+
+                var thendo = "unknown";
+                if("passthrough" in response.echo_req && "thendo" in response.echo_req.passthrough) {
```
jy-binary: what does the word "thendo" stand for?

src/javascript/binary/pages/trade/message.js
```diff
@@ -3,33 +3,97 @@
...
+
+        var response;
+
+        try {
+            response  = JSON.parse(msg.data);
+            if("object" !== typeof response || !("msg_type" in response)) {
+                throw new Error("Response from WS API is not well formatted.");
+            }
+        } catch(e) {
+            throw new Error("Response from WS API is not well formatted.");
```
borisyankov: That is very weird code. Throw + recatch.
Also I don't think too strict format checking does any good, if you have no concrete plan to process the error.

src/javascript/binary/pages/trade/message.js
majid-binary: On Portfolio page when the response to authorize arrives we want to update the balance figure. I thought other pages might need to perform their own processing upon authorization. To enable each caller perform its own desired actions, the caller can send an object in the passthrough like `{"thendo":"myThing"}`. So, thendo is understood to be like (1) first authorize, (2) then do myThing.

src/javascript/binary/pages/trade/message.js
majid-binary: The thinking here has been, (1) the response to API request might be be totally screwed and not a valid JSON at all. If that is the case, `response  = JSON.parse(msg.data);` will cause a Fatal error (hence the `try` block and the throw in the catch block), or (2) it is valid JSON and we succeed in parsing it into an object, but there is no `msg_type` key in it. The code that follows relies on this key to exist, if it does not, we again complain that the response is not well formatted and terminate further execution.

But anyway, if this is being paranoid, and we don't need these checks, I'll remove them.

src/javascript/binary/pages/client/change_passwordws.js
majid-binary: Alphanumeric as in `/^[a-z0-9]+$/i`? 
We do not proactively force strong passwords. With this restriction, wouldn't the ability to set strong passwords compromised?

src/javascript/binary/pages/trade/message.js
```diff
@@ -3,33 +3,97 @@
...
+        switch(msg_type) {
+
+            case "error":
+
+                // change_password
+                if("change_password" in response.echo_req) {
+                    // e.g.: {"echo_req":{"old_password":"1111","new_password":"zzzzzzzz","change_password":"1"},"version":3,"error":{"code":"InputValidationFailed","message":"Input validation failed  - $.old_password: does not match the regex pattern ^.{5,25}$"},"msg_type":"error"}
+                    PasswordWS.apiResponse(response);
+                    break;
+                }
```
majid-binary: Some error responses are stateless. When we receive a response and trigger the `process` function we do not know what part of the code the request originated from (what API method was called). For non-error responses we look at the `msg_type` sub-key to dispatch the data to the right handler. However, in case of an error, `msg_type` itself becomes `error` and our only clue is lost. We cannot assume that for all API methods we could use the same error handler. Each method should be able to decide how to react to errors.

This could be solved with two approaches.
    - Switching to live-api, where the method calls are `thenable`
    - Having API return the method name when an error occurs
For now though, we have to do it as done here.

src/javascript/binary/pages/client/change_passwordws.js
borisyankov: The API should sanitize the input. Restricting characters for passwords is a known security anti-pattern.

src/javascript/binary/pages/trade/message.js
borisyankov: We should certainly be moving to live-api soon ;)

src/javascript/binary/pages/trade/message.js
jy-binary: oh, it's "then do" ? That wasn't clear at all! Suggest you rename it to then_do

PR #598, opened 2015-10-23
download folder

negar-binary:
put pdf in new download folder


raunakkathuria:
Can we keep Japanese pdf to `download/japan` folder just to distinguish

negar-binary:
@raunakkathuria done.

PR #609, opened 2015-10-26
Negar/embed chart

negar-binary:
replace livechart with highchart
remove livechart page and unneeded livechart code
add tradingview to footer
(deployed to qa17)


raunakkathuria:
@negar-binary just a note for future, use descriptive message as commit message

src/javascript/binary/chart.js
```diff
@@ -0,0 +1,4 @@
+var chart;
```
raunakkathuria: why these extra file for these variables only?

negar-binary: @raunakkathuria there was a file called livechart and I removed everything else in it and changed the name to chart, but I think we need these vars as they are used elsewhere. should I move them to another file?

raunakkathuria:
@borisyankov can you also review this, I am reviewing backend code for this

src/javascript/binary/chart.js
borisyankov: Yeah, certainly need to be declared wherever they are used.

negar-binary: @borisyankov @raunakkathuria done

PR #621, opened 2015-10-28
Qingwei/statementws adapt apiv3

qingweibinary:
Fix typo that leads to incorrect variable name


borisyankov:
There is a thing we don't enforce enough, lets do it with this PR though :)
Please rebase into a single commit.

PR #627, opened 2015-10-28
Qingwei/apiv3 migration

qingweibinary:


src/templates/toolkit/account/statementws.html.tt
```diff
@@ -7,7 +7,7 @@
 
     <div id="date-nav-panel">
         <div>
-            <button class="button" id="newer-date">Next Day</button>
```
raunakkathuria: need to use translation for this `text.localize`. You can do it later but keep this in mind

PR #636, opened 2015-10-29
Qingwei/profit table ws

qingweibinary:
new profit table using wsapi

(I would be glad if someone point me to solution of this ugly commit history, either to fix it or to avoid it in the future)


qingweibinary:
Please do not merge it at the moment, route on bom-web are not submitted yet

src/javascript/binary/static/profittable/profittable.ui.js
```diff
@@ -0,0 +1,122 @@
...
+
+    var profitTableID = "profit-table";
+    var cols = ["buy-date", "ref", "contract", "buy-price", "sell-date", "sell-price", "pl"];
+    var header = ["Purchase Date", "Ref.", "Contract", "Purchase Price", "Sale Date", "Sale Price", "Profit/Loss"];
+    var footer = ["Total Profit/Loss", "", "", "", "", "", ""];
+
+    function createEmptyTable(){
+        function mergeRows(start, end, $row){
+            var texts = $row.children().map(function (i, obj) {
+                if (i >= start && i <= end) {
```
borisyankov: You can use .filter() here.
As in .map(...).filter(...)

src/javascript/binary/static/profittable/profittable.ui.js
```diff
@@ -0,0 +1,122 @@
...
+
+        var subTotalType = (total >= 0 ) ? "profit" : "loss";
+        $("#pl-day-total > .pl").removeClass("profit").removeClass("loss");
+        $("#pl-day-total > .pl").addClass(subTotalType);
+    }
+
+    function createProfitTableRow(transaction){
+        var buyDate = moment.
+            utc(transaction["purchase_time"] * 1000).
+            locale("en").
```
borisyankov: Should we be considering the current language for localization of the date?
I don't think we are doing it now, but may be we should?

src/javascript/binary/static/profittable/profittable.ui.js
qingweibinary: Exactly, will fix =)

src/javascript/binary/static/profittable/profittable.ui.js
qingweibinary: Nice one, this is actually a mistake I made when refactoring other parts, will fix

src/javascript/binary/pages/trade/message.js
```diff
@@ -46,6 +44,7 @@ var Message = (function () {
                 }
             } else if (type === 'error') {
                 $(".error-msg").text(response.error.message);
+                console.log("Request failed", response.error);
```
raunakkathuria: remove the console log

src/javascript/binary/pages/trade/content.js
```diff
@@ -208,10 +208,19 @@ var Content = (function () {
         localizeTextContentById("end-of-table");
     };
 
+    var profitTableTranslation = function(){
+        var titleElement = document.getElementById("profit-table-title").firstElementChild;
+        var title = titleElement.textContent;
+        titleElement.textContent = text.localize(title);
+
+        localizeTextContentById("err");
```
raunakkathuria: why again localizeTextContentById, why not content.localize()

src/javascript/binary/static/profittable/profittable.ui.js
```diff
@@ -0,0 +1,100 @@
+
+var ProfitTableUI = (function(){
+    "use strict";
+
+    var profitTableID = "profit-table";
+    var cols = ["buy-date", "ref", "contract", "buy-price", "sell-date", "sell-price", "pl"];
+    var header = ["Purchase Date", "Ref.", "Contract", "Purchase Price", "Sale Date", "Sale Price", "Profit/Loss"];
```
raunakkathuria: better to make it like this `var header = [Content.localize().textPurchaseDate` as explained earlier also, we should have centralized texts for new javascripts. Old js is scattered we should not proceed with same way as old one

src/javascript/binary/static/profittable/profittable.ui.js
```diff
@@ -0,0 +1,100 @@
...
+    var footer = ["Total Profit/Loss", "", "", "", "", "", ""];
```
raunakkathuria: same Content.localize

src/javascript/binary/static/profittable/profittable.ui.js
```diff
@@ -0,0 +1,100 @@
...
+
+        $("#pl-day-total > .pl").text(Number(total).toFixed(2));
+
+        var subTotalType = (total >= 0 ) ? "profit" : "loss";
+        $("#pl-day-total > .pl").removeClass("profit").removeClass("loss");
+        $("#pl-day-total > .pl").addClass(subTotalType);
+    }
+
+    function createProfitTableRow(transaction){
+        var buyDate = moment.
```
raunakkathuria: why we need replace? moment provide multiple formats itself

src/templates/toolkit/account/profit_tablews.html.tt
```diff
@@ -0,0 +1,7 @@
+<div id="profit-table-ws-container">
+    <div class="page-title" id="profit-table-title">
+        <h1>Profit Table</h1>
```
raunakkathuria: localize

src/javascript/binary/pages/trade/content.js
qingweibinary: Forgotten, will fix

src/javascript/binary/pages/trade/message.js
qingweibinary: OK

src/javascript/binary/static/profittable/profittable.ui.js
qingweibinary: because format("YYYY-MM-DD\nHH:mm:ss") does not work as expected, it output to

2015-09-0916:45:22 

instead of 

2015-09-09
16:45:22

borisyankov: I am curious why we are not localizing these things ;)
We haven't been localizing them up till now too, not just for this new code.

borisyankov: BTW, an important approach, with a lot of libs is to extract them into a reusable module like what I am doing here:
https://github.com/binary-com/binary-next-gen/blob/master/src/_utils/DateUtils.js

That is for two reasons:
First, reusability. Even moment.format('YYYY-MM-DD') might need changing, and better have a more 'abstracted' name.
Second - if we decide to change the library itself, it is much, much easier to do so.

raunakkathuria: @qingweibinary we can use same date to get 2015-09-09 and 16:45:22 separately using different format. no need for implementing our own stuff

qingweibinary: So, as a conclusion, should I abstract it into a module?

borisyankov: Not at this moment (no pun intended :) )

src/javascript/binary/static/profittable/profittable.ui.js
qingweibinary: Got it

src/templates/toolkit/account/profit_tablews.html.tt
qingweibinary: Sorry, but what do you mean by this?

raunakkathuria: please use Content.localize() from javascript to localize Profit table

src/javascript/binary/pages/trade/content.js
```diff
@@ -193,25 +207,21 @@ var Content = (function () {
...
-        var element = document.getElementById(id);
-        var textContent = element.textContent;
-        element.textContent = text.localize(textContent);
-    }
-
     var statementTranslation = function(){
         var titleElement = document.getElementById("statement-title").firstElementChild;
-        var title = titleElement.textContent;
-        titleElement.textContent = text.localize(title);
+        titleElement.textContent = localize.textStatement;
```
raunakkathuria: shouldn't it be `Content.localize().textStatement`

src/javascript/binary/static/profittable.js
```diff
@@ -0,0 +1,14 @@
+
+pjax_config_page("profit_tablews", function(){
+    return {
+        onLoad: function() {
+            Content.populate();
```
raunakkathuria: you don't need to populate it as your ids are in your js only. Populate update the static content on trading page

src/javascript/binary/pages/trade/content.js
qingweibinary: localize is in scope thus it is working

src/javascript/binary/static/statement.js
```diff
@@ -1,6 +1,7 @@
 pjax_config_page("statementws", function(){
     return {
         onLoad: function() {
+            Content.populate();
```
raunakkathuria: can you check if we can use it before calling Content.populate?

src/javascript/binary/static/profittable.js
qingweibinary: well, there's some terrible dependency problem
1. since we want to have all text together, they sits in the Content.localize object
2. Content.localize object is initialized when populate is called, it cannot be initialized when script is executed, because by then the global text might not be available
3.  so to ensure the text in Content.localize is accessible, populate must be called after global text is available and before my page use them, which happen to be here

Any suggestion is welcome =)

src/javascript/binary/static/statement.js
qingweibinary: if I do not called it here, all text I need is not available

raunakkathuria: Ok :+1: just wanted to confirm

src/javascript/binary/static/profittable.js
raunakkathuria: I will try to separate it then later. Thanks

qingweibinary: IMO, we should split all text into their corresponding domain (statement, trade, profit table)
so that we dont share the localize object? since all text is in text.json after all

PR #643, opened 2015-10-30
binary-maksym/fixes

binary-maksym:
decrease menu size to prevent 2 lines
show loading logo for spreads analyse
fix broken sell button in franch


raunakkathuria:
its better to add it here for languages where it goes to second line rather than dependent on js,`src/sass/language_base.scss`

binary-maksym:
Thanks, will fix

PR #644, opened 2015-10-30
Nuru/bet tr fr

nuruddeensalihu:
Hi,

Contained are the changes i made in binary.com. Its regarding a task for replacing the start trading page with beta trading interface for our french language in binary.com. Should there be anything with regard to this , please do let me know and would be glad to look into it.


config.json
```diff
@@ -2,7 +2,7 @@
     "website_name": "Binary.com",
     "display_languages": "EN,ID,RU,ES,FR,PT,DE,ZH_CN,PL,AR,ZH_TW,VI,IT",
     "static": {
-        "url": "https://static.binary.com/",
```
raunakkathuria: don't push this change

raunakkathuria:
Please don't commit so many just for minor change, its ok as of now but don't do it for future

src/javascript/binary/base/page.js
```diff
@@ -787,7 +799,15 @@ Page.prototype = {
         var that = this;
         $('#language_select').on('change', 'select', function() {
             var language = $(this).find('option:selected').attr('class');
-            document.location = that.url_for_language(language);
+            var loc = document.location; // quick access
```
raunakkathuria: you can use this `page.url.url_for('/trading')`

borisyankov:
@raunakkathuria I already asked @qingweibinary yesterday about rebasing few commits, so may be we start enforcing this, and may be now is a good time? :)

config.json
borisyankov: This is the most changed line in this project ;)

src/javascript/binary/base/page.js
nuruddeensalihu: Hi Raunak, The url_for when used returned the url and the language of the current page. When used it would return something like 

```/trading?l=EN```  even when the language is changed to a new language from the drop down . This is because unlike the ```url_for_language``` which  accept the selected language as a parameter, the url_for does not accept the selected language and return the language of the existing page before post back as a result. I thought i had to made some change to the url_for if i want it to accomodate my requirement, which add the moment i noticed is being referenced in many places.

PR #657, opened 2015-11-02
binary-maksym/initial load fix

binary-maksym:


src/javascript/binary/base.js
```diff
@@ -1,12 +1,13 @@
 // for IE (before 10) we use a jquery plugin called jQuery.XDomainRequest. Explained here,
```
raunakkathuria: why this is used? we don't support IE < 10

PR #678, opened 2015-11-06
added view button on profit and statement table

qingweibinary:


src/javascript/binary/pages/bet/bet_sell.js
```diff
@@ -738,7 +738,7 @@ var BetSell = function() {
         },
         register: function () {
             var that = this;
-            $('#profit-table, #portfolio-table, #bet_calculation_container, #statement-table, #contract_confirmation_container').on('click', '.open_contract_details', function (e) {
```
raunakkathuria: don't remove `#statement-table` as it is used in backoffice

qingweibinary: I see, ok will fix

PR #683, opened 2015-11-09
patched autocomplete off to page

qingweibinary:


src/javascript/binary/base.js
```diff
@@ -222,3 +222,9 @@ if (!/backoffice/.test(document.URL)) { // exclude BO
         LocalStore.set('active_loginid', match);
     });
 }
+
+
+//this method is to prevent caching on sensitive data
+pjax_config_page("new_account/real", function(){
+    $("#secretanswer").attr("autocomplete", "off");
```
borisyankov: Should we set this in the html directly?

qingweibinary: I'd wished to, but this form is generated by perl entirely
using our in-house perl module called ***HTML::FormBuilder***, and apparently it does not support adding attribute and not much people  have related knowledge on the module, so the effort to get this up would be considerably higher than JS approach.

Is there any security concerns if we use JS? If yes, I would try to fix it from HTML

borisyankov: One more reason not to generate complex html with server-side code ;)

It is fine as it is then. No concerns.

qingweibinary:
Do Not Merge At the moment, some ordering is not correct

qingweibinary:
OK, it's ready for review now ~ fixed the mistake in pjax

src/javascript/binary/base.js
```diff
@@ -222,3 +222,13 @@ if (!/backoffice/.test(document.URL)) { // exclude BO
         LocalStore.set('active_loginid', match);
     });
 }
+
+
+//this method is to prevent caching on sensitive data
+pjax_config_page("new_account/real", function(){
```
jy-binary: this looks like a convoluted way to add a tag.. Why can't you put the tag straight into the html?

qingweibinary: the current html form is generated by a perl module, which does not seems to support tag addition, to make it work require some changes in the perl module which I have no knowledge of, current way of doing it should not possess any security concerns, 
but you're right, it is confusing, so do you want me to fiddle with the perl script or you are OK for this at the moment?

PR #690, opened 2015-11-10
Mohammad/cashier_changes

binary-hamedanchi:
Some changes in the Cashier page

### DIFF
--- a/src/sass/pages/cashier.scss
+++ b/src/sass/pages/cashier.scss
@@ -105,11 +105,6 @@
     width: 111px;
     height: 38px;
 }
-.button_cashier {
-    span {
-        min-width: 75px;
-    }
-}
 .no_side_padding {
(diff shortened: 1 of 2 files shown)


src/sass/pages/cashier.scss
```diff
@@ -105,7 +105,15 @@
     width: 111px;
     height: 38px;
 }
-
+.button_cashier {
```
jy-binary: do we really need to add custom styling to the button? I would be better to keep the number of CSS tweaks down to a minimum. Ideally the entire site uses only what is listed on https://www.binary.com/styles without any further tweaks anywhere.

binary-hamedanchi: It was just to make the buttons ('Deposit' & 'Withdraw') having the same width.
I removed the class.

PR #702, opened 2015-11-12
majid/portfolio_ws

majid-binary:


src/javascript/binary/pages/trade/socket.js
```diff
@@ -40,17 +40,26 @@ var TradeSocket = (function () {
         }
```
qingweibinary: tradesocket is changed, so you probably need to adapt 

Related PR:
https://github.com/binary-com/binary-static/pull/707

src/javascript/binary/pages/portfiliows.js
```diff
@@ -0,0 +1,209 @@
...
+
+    /**
+     * Amounts received from the API could be integer or decimal numbers.
+     * In case we have an integer, like 73, we want to display a decimal
+     * with two fractions, i.e. 73:00
+     * This function does that.
+     * Adapted from http://stackoverflow.com/a/14428340
+     * Kudos to: [VisioN](http://stackoverflow.com/users/1249581)
+    **/
+    var fixCurrency = function(n, c) {
```
jy-binary: this is a common utility function. Can you move it to src/javascript/binary/static/common/string_util.js ?

PR #722, opened 2015-11-16
client side redirect for unauth user

qingweibinary:


src/javascript/binary/static/profittable.js
```diff
@@ -3,6 +3,10 @@
 pjax_config_page("profit_table", function(){
     return {
         onLoad: function() {
+            if (!getCookieItem("login")) {
```
raunakkathuria: why we need this? we can make it `auth_r` call in bom-web itself

PR #733, opened 2015-11-17
preserve decimal place

qingweibinary:


src/javascript/binary/pages/bet/tick_trade.js
```diff
@@ -153,7 +153,7 @@ var TickDisplay = function() {
                         if (data[i][0] === 'tick') {
                             var tick = {
                                 epoch: parseInt(data[i][1]),
-                                quote: parseFloat(data[i][2])
+                                quote: parseFloat(data[i][2]).toFixed(data[i][2].split('.')[1].length)
```
raunakkathuria: i don't know if its valid in our case but if we don't send it as decimal value it will fail (just added a note, don't think we ever send it without decimal places)

qingweibinary: True, I will handle it

jy-binary: How are we rounding? Note that every underlying has its own specific "number of decimals" setting. Do we have an API call that returns this? If not we need to add one, and then do the rounding properly.

borisyankov: Yes, we have info for 'pips' from ActiveSymbols call.
And I have created a function already for binary-next-gen to format based on it.

qingweibinary: this charting solution is not using Websocket, the websocket version is under development by Negar

PR #746, opened 2015-11-18
negar/limitws

negar-binary:
add new static limit page using ws


src/javascript/binary/header.js
```diff
@@ -0,0 +1,14 @@
+var fc_CSS=document.createElement('link');
+fc_CSS.setAttribute('rel','stylesheet');
+var isSecured = (window.location && window.location.protocol == 'https:');
+var lang = document.getElementsByTagName('html')[0].getAttribute('lang');
+var rtlLanguages = ['ar','he'];
+var rtlSuffix = (rtlLanguages.indexOf(lang) >= 0) ? '-rtl' : '';
+fc_CSS.setAttribute('type','text/css');
+fc_CSS.setAttribute('href',((isSecured)? 'https://d36mpcpuzc4ztk.cloudfront.net':'http://assets1.chat.freshdesk.com')+'/css/visitor'+rtlSuffix+'.css');
```
jy-binary: what is this freshdesk stuff?
We shouldn't put it here; it needs to go via Google Tag Manager

PR #754, opened 2015-11-19
Nuru/reality check

nuruddeensalihu:
Hi,

Attached contained the changes that would limit our reality check textbox to accepy number/numeric values only . This was as demonstrated on the usability testing screen . The textbox should accept numerical values only.

Regard;


jy-binary:
this is ridiculous!!

PR #755, opened 2015-11-19
binary-maksym/timout fix

binary-maksym:


src/javascript/binary/socket.js
```diff
@@ -62,16 +62,18 @@ var BinarySocket = (function () {
...
+            if(data.contracts_for || data.proposal){
+                data.passthrough.req_number = ++req_number;
+                timeouts[req_number] = setInterval(function(){
+                    if(typeof reloadPage === 'function'){
+                        var r = confirm("The server didn't respond to the request:\n\n"+JSON.stringify(data)+"\n\nReload page?");
+                        if (r === true) {
+                            reloadPage();
+                        } 
+                    }
+                }, 7*1000);
```
majid-binary: Multiple issues IMHO:
 1. Why `setInterval`? We should just `setTimeout` here and then no need to `clearTimeouts`.
 2. Other UIs increment the next check's interval each time it fails (e.g. Gmail first retries in 5 seconds, if that fails it then tries in 15 seconds, ...). Here we do it every time after 7 seconds.
 3. Other UI's do not present a `confirm`, they show a disconnected message (a `div`) and a count-down timer. On timeout, they do retry (to re-connect, not relaod). That is superior.

borisyankov: I agree with 1)
I agree with 2) but we need to be even more aggressive, and then delay the timeout. We can have a notice for 'offline' work.

We reverted the Trade page to the old one, and there might be reworkings on the back end too.
If this will take time, we can just replace this with binary-live-api that will handle this.

binary-maksym:
This message actually is not going to be the regular message. This message was added because we still have problems with our websocket api and some requests just break it. api just hangs and only reseting all form setting to defaults and reloading page helps. when backend team solves all api issues this message will not appear any more.

I used setinterval instead of settimeout and confirm instead of alarm because sometimes there are requests that are not hanging, but response comes just after timeout is reached

PR #763, opened 2015-11-24
majid/portfolio_ws

majid-binary:


raunakkathuria:
@qingweibinary can we use your flex table function in this?

src/javascript/binary/pages/portfiliows.js
```diff
@@ -0,0 +1,199 @@
...
+    /**
+     * Updates portfolio table
+    **/
+    var updatePortfolio = function(data) {
+
+        /**
+         * Check for error
+        **/
+        if("error" in data) {
+            throw new Error("Trying to get portfolio data, we got this error", data.error);
```
raunakkathuria: How we are handling the localize for these error messages?

qingweibinary:
I believe yes, just that to make the table generic enough, it does not provide much functionality out of the box

src/javascript/binary/static/common/string_util.js
```diff
@@ -31,12 +31,34 @@ var StringUtil = (function(){
         return momentObj.toString();
     }
 
+    /**
+     * Format currency
+     * formatCurrency(20, "USD") -> "USD 20.00"
+     * formatCurrency("10000", "GBP") -> "GBP 10,000.00"
+     * formatCurrency(10.027, "EUR") -> "EUR 10.027"
+    **/
+    function formatCurrency(n, c) {
```
jy-binary: to be removed and use Negar's function instead

src/javascript/binary/pages/portfiliows.js
majid-binary: This is not shown in the interface, just in console. Do we need localization for it?

PR #766, opened 2015-11-24
negar/virtualws

negar-binary:
create new form and page for virtual account creation through ws


src/sass/static/virtual_acc_opening.scss
```diff
@@ -0,0 +1,36 @@
+#virtual-form{
+	text-align: center;
```
jy-binary: why so much custom CSS? Please try to use only our standard styling from https://www.binary.com/styles

src/templates/haml/home2/styles.html.haml
```diff
@@ -0,0 +1,519 @@
+- stash(
```
jy-binary: why is this file duplicated?

src/templates/haml/home2/index.html.haml
```diff
@@ -0,0 +1,120 @@
+- stash(
```
jy-binary: I don't think you need a home2. Home2 is just for a/b testing. This project isn't about a/b testing.

src/javascript/binary/static/verify_email.js
```diff
@@ -0,0 +1,30 @@
+//get gtm layer from ashkan before release
```
raunakkathuria: remove this

ashkanx: Let my name be there :)

borisyankov: you can add // todo:
to comments like this, and then grep them before committing ;)

negar-binary: @borisyankov good idea, thanks :D

PR #768, opened 2015-11-24
Nuru/reality check

nuruddeensalihu:
Hi,

Made some few changes to the code that should reduce the size and give us the same result. Should there be another issue again wit regard to this , please do let me know would be glad to check it :)

Regard;
Nuru;


src/templates/haml/user/reality_check.html.haml
```diff
@@ -43,7 +43,7 @@
     - }
   %p
     %label= l('Reality Check Interval:')
-    %input{:type => 'text' :interval => 1 :size => '6'}
+    %input{:type => 'text' :interval => 1 :size => '6' :id =>'realityDuration'}
```
borisyankov: <input type="number"> will solve a lot of your problems ;)

nuruddeensalihu: Hi boris ;

Thanks for the suggestion. Have though about it as well. But input type displays like below 
image

With an up and down arrow option. We though it does not looks good that way. :)

PR #778, opened 2015-11-25
Mohammad/trading_times_ws

binary-hamedanchi:


src/javascript/binary/pages/resources/trading_timesws.js
```diff
@@ -0,0 +1,156 @@
...
+        var markets = response.trading_times.markets;
+
+        var $ul = $('<ul/>');
+        var $contents = $('<div/>');
+
+        for(var m = 0; m < markets.length; m++) {
+            var tabID = 'tradingtimes-' + markets[m].name.toLowerCase();
+
+            // tabs
+            $ul.append($('<li/>', {class: 'ja-hide'}).append($('<a/>', {href: '#' + tabID, text: markets[m].name, id: 'outline'})));
```
jy-binary: why is there anything to do with ja-hide here?

ashkanx: For JA language we hide all markets except Forex major pair using ja-hide.
Another way is to remove all other markets with JS instead of hiding them with css.

PR #785, opened 2015-11-30
mohammad/paymentagent_withdraw_ws

binary-hamedanchi:


src/javascript/binary/static/cashier/paymentagent_withdrawws.js
```diff
@@ -0,0 +1,271 @@
...
+    var fieldIDs = {
+        ddlAgents : '#ddlAgents',
+        txtAmount : '#txtAmount',
+        txtDesc   : '#txtDescription'
+    };
+    var formData, isValid;
+    var withdrawCurrency = 'USD',
+        minAmount = 10,
+        maxAmount = 2000;
+
```
chylli: Maybe we will change these values (In fact, if no problem, I have changed the values in my branch')
So it will be great if we get these values from server directly.

binary-hamedanchi: Sure, I hope so, but unfortunately there is no WS method returning them at the moment.
Please let me know if there was an API call which can be used to retrieve such values.

PR #800, opened 2015-12-01
majid/fixbug799

majid-binary:
Fixed issue 'Unending page load if no open contracts


src/javascript/binary/pages/portfiliows.js
```diff
@@ -40,11 +40,12 @@ var PortfolioWS =  (function() {
         }
```
jy-binary: note that there's a typo in the file name - should be portfoliows.js not portfiliows.js

PR #819, opened 2015-12-09
text required by next-gen

qingweibinary:
This changes is hard to test and check alone, we need to have a thorough test before release to production


jy-binary:
No, we shouldn't add the text here. binary-next-gen is a totally independent project, nothing to do with binary-static. You have to organize it's setup in weblate totally independently.

borisyankov:
@jy-binary We do not have functionality to share the translations and 90% at least match what is already translated.
We also discussed with Raunak, unlikely for binary-static to be used as white-label fascility, means translations don't even need to be in binary-static.

raunakkathuria:
@borisyankov binary-static translations are not only for static, it covers all of our repo like bom, platform, web. IMO we should have separate project as it will more flexibility in future and we can save complete JSON in there or if you don’t want to have duplicate stuff, we can create PO, POT file for next gen and merge string that are already there in static to next-gen one

PR #850, opened 2015-12-23
negar/password_entropy

negar-binary:
password entropy


src/javascript/binary/base/passwordmeter.js
```diff
@@ -0,0 +1,179 @@
...
+       strVerdict = text.localize("Strong");
+    }
+    else
+    {
+       strVerdict = text.localize("Stronger");
+    }
+
+  var array = [intScore, strVerdict];
+  return array;
+}
```
borisyankov: That lib doesn't look very good, to be honest.
We had discussed, and I proposed that actually good one:
https://github.com/dropbox/zxcvbn

Based on:
https://xkcd.com/936/

:)

negar-binary: @boeisyankov that lib was rejected by JY and this was the lib he asked me to use

borisyankov: Why?

borisyankov: zxcvbn is a library created and used by Dropbox - a $10b company.
This one is 10 year old 'quick and dirty password quality meter' - by the words of the authors.
Its calculations give completely incorrect results of the actual entropy.

mwmiller:
You probably want something which actually [computes entropy](https://github.com/mvhenten/string-entropy) instead of an ad-hoc collection of heuristics.

You could wrap the library fairly simply to return score and verdict like the above.

negar-binary:
@borisyankov @mwmiller please confirm with JY as he is the one who chose this library

PR #879, opened 2016-01-05
Nuru/vrt topup

nuruddeensalihu:
Hi,

Attached contained Nuru_Vrt_topup changes for your perusal.


src/templates/haml/cashier/top_up_virtualws.html.haml
```diff
@@ -0,0 +1,9 @@
+.grd-grid-12
+  %p#VRT_topup_errorMessage
+  %h1#VRT_title= l('Give me more money !')
+  %div{:class => 'centerText', :id => 'VRT_topup_message'}
+    %p
+      %form{:action => url_for('/user/portfolio'),:method => 'POST'}
```
jy-binary: I think the new URL for portfolio has "ws" at the end?

ashkanx: It's /openpositionsws now.

PR #915, opened 2016-01-13
Hook balance subscription to auth, instead of onload

qingweibinary:


qingweibinary:
@ashkanx ok to merge ;)

ashkanx:
Thanks

PR #924, opened 2016-01-13
mohammad/reality_logout_ws

binary-hamedanchi:


src/javascript/binary/reality-check.js
```diff
@@ -165,8 +164,8 @@ RealityCheck = (function ($) {
             $('#reality-check').remove();
         });
 
-        $('#reality-check .blogout').on('click', function () {
-            window.location.href = logout_url;
+        $('#reality-check #btn_logout').unbind('click').click(function(){
+            BinarySocket.send({"logout": "1"});
```
raunakkathuria: will this redirect to login page after logging out? if this is not the case then client will still see logind and logout button in header

binary-hamedanchi: Yes, sure. Redirecting to homepage will be handled by `do_logout` function in `page.js` which will be called on the ws response. Our main logout button in header is already doing the same.

raunakkathuria: Ok sure, just wanted to confirm we handle in logout response.  Then it seems fine

binary-hamedanchi: Actually it's redirecting to homepage (not login page) because the old logout button was redirecting to homepage too.

raunakkathuria: as long its redirecting its fine

PR #927, opened 2016-01-13
Fix for validating cookies

4p00rv:
Hi,

There are many places where we take the cookies value and show it in the HTML without validating it. This can cause security issues. This fix validates the `login ids` and `logs out` a user if the value is invalid.


src/javascript/binary/pages.js
```diff
@@ -38,6 +38,35 @@ function get_login_page_url() {
...
+    if (getCookieItem('login')){
+        var accIds = $.cookie("loginid_list").split("+");
+        var loginid = $.cookie("loginid");
+        var logoutResponse = JSON.parse('{'+
+                                '"echo_req": {'+
+                                    '"logout": 1'+
+                                '},'+
+                                '"logout": 1,'+
+                                '"msg_type": "logout"'+
+                                '}');
```
binary-hamedanchi: Please do not pass a fake response. This will not run the related back-end code. 
You need to just send a `logout` request via BinarySocket.

4p00rv: Sure, will get that done.

4p00rv:
@binary-hamedanchi Made relevant changes. Please review now.

binary-hamedanchi:
Seems fine. Thank you.
You can remove that `console.log`, too.

4p00rv:
My bad. Done!

src/javascript/binary/base/page.js
```diff
@@ -523,6 +523,27 @@ Header.prototype = {
             BinarySocket.send({"logout": "1"});
         });
     },
+    
+    validate_cookies: function(){
+        function is_valid_id(id){
+            return /^(MX|MF|VRTC|MLT|CR|FOG|VRTJ|JP)[0-9]+$/i.test(id);
```
raunakkathuria: i don't these broker codes should be in frontend, we should have this check in backend in my opinion. Even if we go with this then remove FOG as will never be there

src/javascript/binary/base/page.js
```diff
@@ -523,6 +523,27 @@ Header.prototype = {
...
+        }
+        
+        if (getCookieItem('login')){
+            var accIds = $.cookie("loginid_list").split("+");
```
raunakkathuria: need to have check if `loginid_list` is present or not

4p00rv:
@raunakkathuria  Made the changes you asked for!

src/javascript/binary/base/page.js
```diff
@@ -523,6 +523,27 @@ Header.prototype = {
             BinarySocket.send({"logout": "1"});
         });
     },
+    
+    validate_cookies: function(){
+        function is_valid_id(id){
+            return /^(MX|MF|VRTC|MLT|CR|VRTJ|JP)[0-9]+$/i.test(id);
```
negar-binary: you can get the values from page.settings.get('valid_loginids'), it returns "MX|MF|VRTC|MLT|CR|FOG|VRTJ|JP'" so you don't have to hard code it

negar-binary: well, this.settings.get('valid_loginids') :)

src/javascript/binary/base/page.js
```diff
@@ -523,6 +523,27 @@ Header.prototype = {
...
+        }
+        
+        if (getCookieItem('login') && getCookieItem('loginid_list')){
+            var accIds = $.cookie("loginid_list").split("+");
+            var loginid = $.cookie("loginid");
+                                    
+            if(!is_valid_id(loginid)){
+                BinarySocket.send({"logout": "1"});
+            }
+
```
negar-binary: in form.js there is a function is_loginid_valid that is doing the same thing, I believe

negar-binary: for example, I can do client_form.is_loginid_valid(page.client.loginid) and it will return true or false

4p00rv: You mean that there's already an implementation of `is_valid_id` ?

negar-binary: yes, it's in form.js, it's called is_loginid_valid

PR #947, opened 2016-01-18
mohammad/switchto_open_market

binary-hamedanchi:
Try to select an open market, if the current market is not active.

[Related Trello Card](https://trello.com/c/2ol8Rqu6/)


src/templates/toolkit/bet/static.html.tt
```diff
@@ -3,6 +3,7 @@
 [% END %]
 
 <div id="trading_socket_container" class="tab-menu-wrap">
+   <p id="markets_closed_msg" class="notice-msg center hidden">[% l('All markets are closed now. Please try again later.') %]</p>
```
raunakkathuria: when this message will be shown to client? As random market is never closed

ashkanx: I guess it should be only for MF.

PR #951, opened 2016-01-19
binary-maksym/switch from closed market

binary-maksym:


raunakkathuria:
@binary-maksym there is already pull request for this https://github.com/binary-com/binary-static/pull/947

binary-maksym:
@raunakkathuria they are doing different things

PR #960, opened 2016-01-19
Fix multiple error messages

4p00rv:
Fix for this [trello card](https://trello.com/c/Uj6yisX7/109-different-error-messages)


src/templates/toolkit/bet/static.html.tt
```diff
@@ -50,7 +50,7 @@
                         </div>
                         <div class="big-col">
                            <div id="expiry_type_duration">
-                              <input id="duration_amount" type="number" min="1" max="366" class="small_width_input" /><label></label>
```
raunakkathuria: @4p00rv why we remove `min, max`, we use to validate the form. @negar-binary can you confirm this

raunakkathuria:
@4p00rv Can you add description of changes done and why it is required?

src/javascript/binary/pages/trade/event.js
```diff
@@ -237,8 +237,7 @@ var TradingEvents = (function () {
         var amountElement = document.getElementById('amount');
         if (amountElement) {
             amountElement.addEventListener('input', debounce( function(e) {
-                var amount_val = parseFloat(e.target.value);
-                if (e.target.value % 1 !== 0 && ((+amount_val).toFixed(10)).replace(/^-?\d*\.?|0+$/g, '').length>2) {
+                if (e.target.value % 1 !== 0 && isStandardFloat(parseFloat(e.target.value))) {
```
raunakkathuria: as you are checking this `e.target.value % 1 !== 0` everywhere why not move it to `isStandardFloat`

src/templates/toolkit/bet/static.html.tt
4p00rv: As discussed with Ashkan, we are only showing one error everywhere not multiple errors. The backend already replies with errors, if any, and is displayed in the frontend!

PR #996, opened 2016-01-27
remove all configs from config.json

raunakkathuria:


config.json
```diff
@@ -1,37 +1,3 @@
...
-        "default_telephone": {
-            "country": "United Kingdom"
-        },
-        "skype_helpline": {
-            "EN": "binaryhelpline",
-            "ID": "binaryhelplineid",
-            "RU": "binaryhelplineru"
-        }
-    }
+    "website_name": "Binary.com"
```
jy-binary: Why do you still need website_name in the back-end?
Remember that the back-end will have only one purpose: to expose the WSAPI. I don't see any API call that returns website_name.

raunakkathuria: this was kept so that config.json is not empty, i am considering other stated in other pull request

PR #997, opened 2016-01-27
negar/android_app_trading

negar-binary:
add a google play badge to trading page on randoms market


ashkanx:
Don't release this until application is ready

ashkanx:
@negar-binary we can release this anytime

PR #1000, opened 2016-01-27
reality_check-fix

4p00rv:
Reality check popup would keep on appearing again and again for large inputs (input >35800). This happened due to `window.setTimeout()` accepting `32bit integers` whose largest value would be `2147483647` in miliseconds. Changing the input limit to 4 digits is the suggested fix.
Related [Trello Card](https://trello.com/c/eWeNzONc/156-reality-check-interval-validation-is-not-working-properly).


4p00rv:
1000 PRs :astonished:

PR #1003, opened 2016-01-28
fix dob

4p00rv:
Fix default `1970-01-01` Date-of-birth for `null` values.


shuwnyuantee:
:+1:  Thanks.

4p00rv:
Np :smile:

PR #1009, opened 2016-01-29
negar/realws_residence

negar-binary:
set residence if not set for virtual accounts during account upgrade


src/javascript/binary/pages/client.js
```diff
@@ -51,7 +51,7 @@ var enable_residence_form_submit = function () {
     });
 };
 
-pjax_config_page('new_account/real', function() {
+/*pjax_config_page('new_account/real', function() {
```
jy-binary: why is this commented?

negar-binary: @jy-binary instead of changing the new_account/realws page name, I've commented the pjax for new_account/real so they don't clash, since we will be replacing it and not using it. I can instead change the page name if we are not planning to remove the code for new_account/real after this release

PR #1012, opened 2016-02-01
Added length constraint to stake/payout input

4p00rv:
Fix for this [trello card](https://trello.com/c/3ZrMwa6X/119-purchase-option-freezing-after-put-to-big-number-of-stake-payout)

### DIFF
--- a/src/javascript/binary/pages/trade/event.js
+++ b/src/javascript/binary/pages/trade/event.js
@@ -235,6 +235,7 @@ var TradingEvents = (function () {
         var amountElement = document.getElementById('amount');
         if (amountElement) {
             amountElement.addEventListener('input', debounce( function(e) {
+                e.target.value = e.target.value.replace(/[^0-9.]/g, '');
                 if (isStandardFloat(e.target.value)) {
                     e.target.value = parseFloat(e.target.value).toFixed(2);
                 }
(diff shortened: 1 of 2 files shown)


src/javascript/binary/pages/trade/event.js
```diff
@@ -235,6 +235,7 @@ var TradingEvents = (function () {
         var amountElement = document.getElementById('amount');
         if (amountElement) {
             amountElement.addEventListener('input', debounce( function(e) {
+                e.target.value = e.target.value.replace(/[^0-9.]/g, '');
```
4p00rv: For removing inputs except digits and decimal numbers.

src/templates/toolkit/bet/static.html.tt
```diff
@@ -154,7 +154,7 @@
                         <div class="row-inner big-col">
                            <div class="col-inner">
                               <select id="currency" class="small_width_input"></select>
-                              <input id="amount" type="number" step="any" value="10" class="medium_width_input" />
+                              <input id="amount" type="text" step="any" maxlength="10" value="10" class="medium_width_input" />
```
4p00rv: Setting maxlength to restrict input. Maxlength of valid payout is 8 for `99999.99` rounded it off to nearest tens place.

PR #1013, opened 2016-02-02
Disable non trading dates

4p00rv:
Disabled weekends from datepicker. Related [trello card](https://trello.com/c/LLdgQGAR/135-non-trading-dates-should-be-disabled-in-the-calendar-similar-to-legacy-trading-page)


4p00rv:
Because excessive complexity closing this for now.

PR #1025, opened 2016-02-05
Rewrite reality check for simplicity

qingweibinary:
Not Ready to merge ~~~


qingweibinary:
@ashkanx 
Things not done ~
a) Hook RealityCheck.init with login action

Solution
a) Change login form to use jquery.post instead of HTML form submission
b) use data return by jquery.post of login form, if user is successfully logged in, call RealityCheck.init()

after that it's good for testing ;)

PR #1026, opened 2016-02-05
Fix tick trade chart

4p00rv:
Fix for [trello card](https://trello.com/c/Qffy9ofY/161-missing-decimal-points-in-tick-trade-contract-chart).


src/javascript/binary/pages/bet/tick_trade.js
```diff
@@ -18,7 +18,7 @@ var TickDisplay = function() {
             $self.contract_start_ms = parseInt(data.contract_start * 1000);
             $self.contract_category = data.contract_category;
             $self.set_barrier = ($self.contract_category.match('digits')) ? false : true;
-            $self.display_decimals = data.display_decimals;
+            $self.display_decimals = data.display_decimals || 2;
```
4p00rv: `data.display_decimals` returned `undefined` which assigned the `new_y`, under `formatter`, an int value.

PR #1029, opened 2016-02-08
add gtm tag to head

negar-binary:


src/templates/toolkit/global/head.html.tt
```diff
@@ -5,6 +5,20 @@
 <html>
 [% END %]
     <head>
```
jy-binary: I don't think you need to add to the toolkit version. We only redirect to www2 when the client hits the homepage.

PR #1037, opened 2016-02-16
add scopes

fayland:


src/templates/toolkit/account/api_tokenws.html.tt
```diff
@@ -19,6 +19,18 @@
...
+                    <div class="grd-grid-4 form_label">
+                        <label for="txtName">[% l('Choose Scopes:') %]</label>
+                    </div>
+                    <div class="grd-grid-8">
+                        <input id="scopes" type="checkbox" name="scopes[]" value="read" checked="checked" /> Read
+                        <input type="checkbox" name="scopes[]" value="trade" /> Trade
+                        <input type="checkbox" name="scopes[]" value="payments" /> Payments
+                        <input type="checkbox" name="scopes[]" value="admin" /> Admin
+                    </div>
+                </div>
```
tfoertsch: I think we need an explanation what these scopes mean on this page.

tfoertsch: It's not obvious either that a token is valid for one loginid only. So, a token issued for VRTC90000002 is not valid for CR90000000.

fayland: Sorry, I do not understand that. a token will tell that which loginid it is.

fayland: for explanation. yes. it should be good. just do not know the sentences and how to put them

Thanks

ashkanx: @fayland send me a brief of difference between each scope. I will prepare the writing.

fayland: you can find the related APIs in https://developers.binary.com/api/

there is scope for each API

ashkanx: Thanks

tfoertsch: My remark about the loginid-token relationship was about the following. A user logs in once. Later he can switch loginids. So, not knowing about how our system works, he can get the impression that there is one account (namely the one he has logged in to) with just a number of loginids assigned to it. It's not clear when he creates a token that this token includes only one loginid instead of being related to the whole account which includes all his loginids.

fayland: that requires a full change I'm afraid. right now oauth and access token are both based on loginid. not user itself.

PR #1078, opened 2016-05-23
mohammad/update_websocket_url

binary-hamedanchi:
Update WebSocket URL

[Trello Card](https://trello.com/c/2EeQ7hTn/)

### DIFF
--- a/index.html
+++ b/index.html
@@ -41,16 +41,7 @@
             if(lang) {
                 redirect(lang);
             } else {
-                var host = window.location.host,
-                    socketUrl;
-                if((/www\.binary\.com/i).test(host)) {
-                    socketUrl = 'wss://ws.binaryws.com/websockets/v3';
-                } else if((/binaryqa/i).test(host)) {
-                    socketUrl = 'wss://' + host + '/websockets/v3';
-                } else {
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -41,16 +41,7 @@
...
-                    socketUrl;
-                if((/www\.binary\.com/i).test(host)) {
-                    socketUrl = 'wss://ws.binaryws.com/websockets/v3';
-                } else if((/binaryqa/i).test(host)) {
-                    socketUrl = 'wss://' + host + '/websockets/v3';
-                } else {
-                    socketUrl = 'wss://www2.binary.com/websockets/v3';
-                }
-                var ws = new WebSocket(socketUrl);
+                var ws = new WebSocket('wss://ws.binaryws.com/websockets/v3');
```
jy-binary: can this not pick up the URL mentioned in src/javascript/config.js ? There should only be one place where we mention the WSURL

binary-hamedanchi: I tried to make this index.html as small as it can be.

For reading config.js we have to insert binary.min.js which is almost ~2MB and affects this page's loading speed.

We may change it, once we remove concatenation of js files.

PR #1090, opened 2016-05-24
negar/pa_transfer_vrtc

negar-binary:
don't allow virtual clients to see this page


jy-binary:
can you add me as a collaborator to your binary-static fork so I can see the Update Branch button?

PR #1120, opened 2016-05-31
negar/remove_duplicated_js

negar-binary:
remove duplicated js functions
move files into a better hierarchy


jy-binary:
pls check conflicts

PR #1148, opened 2016-06-08
Fixed active_symbols bug, added dev on localhost

aminmarashi:


jy-binary:
can you add me as a collaborator so I can see the Update Branch button?

PR #1204, opened 2016-06-20
mohammad/revoke_confirmation

binary-hamedanchi:
Add confirmation for revoke an application

### DIFF
--- a/src/javascript/binary/websocket_pages/user/account/settings/authorised_apps/authorised_apps.js
+++ b/src/javascript/binary/websocket_pages/user/account/settings/authorised_apps/authorised_apps.js
@@ -35,8 +35,12 @@ var ApplicationsUI = (function(){
         var $viewButton = $viewButtonSpan.children(".button").first();
         $viewButton.text(text.localize("Revoke access"));
         $viewButton.on("click",function(){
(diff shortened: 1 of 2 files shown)


src/javascript/binary/websocket_pages/user/account/settings/authorised_apps/authorised_apps.js
```diff
@@ -35,8 +35,12 @@ var ApplicationsUI = (function(){
         var $viewButton = $viewButtonSpan.children(".button").first();
         $viewButton.text(text.localize("Revoke access"));
         $viewButton.on("click",function(){
-            ApplicationsData.revokeApplication(data.app_id);
-            $row.css({ opacity: 0.5 });
+            if(window.confirm(
+                text.localize('Are you sure that you want to permanently revoke access of application') +
```
jy-binary: change to "Are you sure that you want to permanently revoke the applications's access?"

binary-hamedanchi: @jy-binary 
This string later will concat with regarding application's name. The result would be:
```
Are you sure that you want to permanently revoke access to application: "Binary.com"?
```

Is it fine?

PR #1206, opened 2016-06-20
negar/exclude_minutes

negar-binary:
add time out facility


jy-binary:
please check conflicts

PR #1228, opened 2016-06-23
negar/tidy

negar-binary:
tidy code and write tests for statement and profit table pages


src/javascript/binary/websocket_pages/user/account/__tests__/profit_table.js
```diff
@@ -0,0 +1,41 @@
+var expect = require('chai').expect;
```
jy-binary: tests should go into a /t/ directory

negar-binary: as discussed with @aminmarashi and @jy-binary, we will follow the __tests__ convention of other front-end projects.

PR #1297, opened 2016-07-11
Minor clean up

eugene-eeo:
**Note:** Safari's User-Agent reports two versions, a technical version under `Safari/<version>`, and a user friendly version under `Version/<version>`. The rewritten parser prefers the user-friendly version.


src/javascript/binary/websocket_pages/user/account/settings/iphistory/iphistory.ui.js
```diff
@@ -1,6 +1,38 @@
...
+            {name: 'Firefox',   regex: 'Firefox/' + vn},
+            {name: 'Opera',     regex: 'OPR/' + vn},
+            {name: 'Opera',     regex: 'Opera/' + vn},
+            {name: 'Safari',    regex: 'Version/' + vn},
+            {name: 'Internet Explorer', regex: ';MSIE '+ vn +';'},
+        ];
+        var len = lookup.length;
+        var name = 'Unknown';
+        var version = 'Unknown';
+        while (len--) {
```
qingweibinary: Seems forEach would be more clearer

eugene-eeo: Yes, ``forEach`` is clearer but we need to exit early from the loop once we've "hit" a matching browser.

PR #1313, opened 2016-07-15
Refactor authorised apps

eugene-eeo:
- Clean up `Applications{UI,Data,}`
 - Since this is very similar to the `IPHistory` view (just populate the table with data) I have written a new wrapper around the `Table` class to make it more declarative. For example now we can just write:

    ```js
    new FlexTableUI({
        container: '#container-id',
        header: ["Header 1", "Header 2"],
        id:     'table-id',
        cols:   ['col1', 'col2'],
        data:   [/* from websocket */],
        formatter: function(datum) {
            return [datum.col1, datum.col2];
        },
        style: function($row, app) {
            /* can apply styles to $row */
        },
    });
    ```
    It also handles cleaning up the table, appending to the DOM, etc. The plan is to slowly move all code that currently uses `Table` to the new wrapper, so that views don't need to concern themselves with how `Table` works underneath.
 - Another advantage is that UI code is now (more) testable.


jy-binary:
Can you add me as a collaborator to your fork, so that I can see the "Update Branch" button?
Thanks, JY

eugene-eeo:
@jy-binary :+1: done

PR #1369, opened 2016-07-29
mohammad/unicode_names

binary-hamedanchi:


eugene-eeo:
We can use a more intuitive regex here, see http://stackoverflow.com/questions/6798745/regex-expression-for-name-validation. However note that first and last names (names in general) are *very* hard to validate anyways. If we want to completely validate them there are always people with names that contain special characters. Rather, I feel that we should prompt them to confirm their names, if it contains special characters.

src/javascript/binary/common_functions/valid_account_opening.js
```diff
@@ -57,7 +57,7 @@ var ValidAccountOpening = (function(){
       errorFname.innerHTML = Content.errorMessage('min', '2');
       Validate.displayErrorMessage(errorFname);
       window.accountErrorCounter++;
-    } else if (!/^[a-zA-Z\s-.']+$/.test(fname.value)){
+    } else if (/[`~!@#$%^&*)(_=+\[}{\]\\\/";:\?><,|\d]+/.test(fname.value)){
```
jy-binary: why is this regexp different to the one that Sasha used in the JSON schema, i.e. "^[\\p{L}\\s'.-]{2,30}$" ?

binary-hamedanchi: We can't use that regex in javascript. As an alternative way, this regex invalidates unwanted characters and allow others.

This is exactly same as what is used for Japanese clients here: https://github.com/binary-com/binary-static/blob/master/src/javascript/binary/websocket_pages/user/new_account/japan_acc_opening/japan_acc_opening.ui.js#L105

PR #1370, opened 2016-08-01
mohammad/blue_green_percent

binary-hamedanchi:


src/javascript/config.js
```diff
@@ -13,11 +13,30 @@ function getAppId() {
...
 
 function getSocketURL() {
-    var server_url = localStorage.getItem('config.server_url'),
-        loginid = $.cookie('loginid');
-    if(!server_url) server_url =
-        (/staging\.binary\.com/i.test(window.location.hostname) ? 'www2' :
-            (loginid && !/^VRT/.test(loginid) ? 'green' : 'blue')
-        ) + '.binaryws.com';
+    var server_url = localStorage.getItem('config.server_url');
+    if(!server_url) {
```
jy-binary: this means that if https://www.binary.com/en/endpoint.html is set, it over-rides?

binary-hamedanchi: Yes, both server_url and app_id would be overridden by endpoint, if it's been set.

PR #1389, opened 2016-08-05
mohammad/utm_source

binary-hamedanchi:


src/javascript/binary/common_functions/google_adwords.js
```diff
@@ -0,0 +1,61 @@
...
+ * Handles utm parameters/referrer to use on signup
+ * 
+ * Priorities:
+ * 1. Cookie having utm data (utm_source, utm_medium, utm_campaign) [Expires in 3 months]
+ * 2. Query string utm parameters
+ * 3. document.referrer
+ *
+ */
+
+var AdWords = (function(){
```
jy-binary: I don't think you need to call it "Adwords". UTM codes have become an industry standard, used by many systems, not just adwords

binary-hamedanchi: Noted, thanks. Changed to `TrafficSource`.

PR #1408, opened 2016-08-11
mohammad/new_trading

binary-hamedanchi:


eugene-eeo:
very rough look over the code, maybe my final code review 😁 `not_first_request`, well... it doesn't feel right. maybe we should use `is_first_request`? that way you can just set it equal to true initially, (saving one line! of code) and it reads better.

binary-hamedanchi:
@eugene-eeo  I agree. I meant to change it, but I've forgotten. Thanks for reminding me.
Wish you the best.

src/sass/pages/trading.scss
```diff
@@ -8,27 +8,22 @@
...
     color: $COLOR_WHITE;
-    border: solid 1px $COLOR_ORANGE;
-    border-radius: 0;
-    border-bottom-left-radius: 4px;
-    background: $COLOR_ORANGE;
-    font-size: 24px;
-    font-weight: bold;
-    line-height: 0px;
+    line-height: 35px;
+    font-size: 26px;
```
negar-binary: is there a font-size from constants we can use?

binary-hamedanchi: Done. Thanks.

PR #1419, opened 2016-08-14
Shashank/Fixing HTML injection

Shashank-In:


src/javascript/binary/websocket_pages/user/account/settings/settings_detailsws.js
```diff
@@ -164,8 +164,9 @@ var SettingsDetailsWS = (function() {
             }
 
             // town/city
-            isRequiredError(fieldIDs.city);
-
+            if(!isRequiredError(fieldIDs.city) && !(/^[a-zA-Z0-9\s\,\.\-\/\(\)#']+$/).test(city)) {
```
negar-binary: it should be `.test(fieldIDs.city)`

borisyankov: Nope, it should be isValidCity(fieldIDs.city) and isValidCity should be unit tested, and should be tested against all addresses already in DB to verify if it is correct or not. 🌴

Shashank-In:
````
should be unit tested, and should be tested against all addresses already in DB to verify if it is correct or not
````
Didn't get this 
Sorry poor coders here :(

borisyankov: I mean when white-listing characters or being over-specific with regex it is very very likely that you have many false positives.

For example it is likely that we accept localised cities and not just latin. For example I live in 'Пловдив', which is rejected. On the other hand, it might not be the case. I don't know and you don't know yet.

Unit tests = we don't have any in binary-static, but that is the best practice.

If we add a field validation but not test it well, fixing one issue might cause another much more severe issue.

For example this is what we use for email validation:
https://github.com/binary-com/binary-utils/blob/master/src/misc/isValidEmail.js

And this is our unit tests:
https://github.com/binary-com/binary-utils/blob/master/src/misc/__tests__/isValidEmail-test.js

BTW, they an be more extensive, I'd say this is the absolute minimum.

Shashank-In:
codes have been changed this wont work now

PR #1431, opened 2016-08-17
negar/disable_financial_binaries_belgium

negar-binary:
change yml file to follow back-end


ashkanx:
I didn't know we have such file in front-end :|

negar-binary:
@ashkanx we are using it in some places, @binary-hamedanchi and I were thinking that we should remove it in another card along with the other (mostly unused) settings stored in our header currently

PR #1432, opened 2016-08-17
negar/remove_unneeded_files

negar-binary:
remove unneeded files in head, use ws instead


build/fileindex.js
```diff
@@ -4,35 +4,33 @@ module.exports = {
...
+            dest: global.dist + '/javascript.json',
             src: [
-            'src/javascript/lib/jquery.js', 
-            'src/javascript/lib/highstock/highstock.js', 
+            'src/javascript/lib/jquery.js',
+            'src/javascript/lib/highstock/highstock.js',
             'src/javascript/lib/highstock/highstock-exporting.js',
-            'src/javascript/lib/moment/moment.js', 
+            'src/javascript/lib/moment/moment.js',
             'src/javascript/lib/**/*.js',
```
eugene-eeo: Does this include all `.js` files in the `lib/`? If so we can remove the above 4 lines.

negar-binary: let's leave it for now @eugene-eeo, I don't want this PR too get too scary :) we can do it in further refactoring, as there is a lot that can be done to optimize the builds and grunt tasks.

jy-binary:
got conflicts now

PR #1457, opened 2016-08-22
implement 404 page

eugene-eeo:


src/templates/static/404.html.tt
```diff
@@ -0,0 +1,16 @@
+<div class="404_page" id="404-container">
+    <h1>[% l('Oops... Page Not Available') %]</h1>
+    <div class="big-error-code">404</div>
+    <p>
+        [% l('The page you requested could not be found. Either it no longer exists or the address is wrong. Please check for any typos.') %]
+    </p>
+    <p>
+        [% l('Have better luck at trading!') %]
```
jy-binary: this sentence sounds a bit taunting (sounds like we are making fun of clients). Suggest to remove it or re-word.

eugene-eeo: cc @ashkanx

PR #1531, opened 2016-09-11
Shashank/matching_regex_with_schemas

Shashank-In:
Updated regex which works


binary-hamedanchi:
1. The RegExps are correct, but they are so huge.
Could you make it simple: allowing other characters except unwanted ones?
Like `/^[^~!@#$%^&*)(_=+\[}{\]\\\/";:\?><,|\d]+$/` which is also used for validating first name and last name here: https://github.com/binary-com/binary-static/blob/master/src/javascript/binary/common_functions/valid_account_opening.js#L60

2. It should also be used for some other fields such as account creations, MT account signup (other repo), etc.

Shashank-In:
@aminmarashi Actually if we see the codes . Our backend has p{L} which allows all the characters . But javascript doesn't support p{L} . So I had to add all the unicode charters which would be like a paragraph long regex and is still the same allowing everything as we do allow everything. The dangers are from the meta characters like backticks , angle brackets , brackets. So had to eliminate that . Over all here in this scenario either we black list or white list . Thing will be same as we need every character and not limited to english alphas

PR #1536, opened 2016-09-13
negar/fix_dots

negar-binary:


jy-binary:
PR deletes 11000 lines from the codebase?

PR #1557, opened 2016-09-20
negar/fix_jp_header

negar-binary:
fix jp header in portfolio and length of profit table


src/javascript/binary/websocket_pages/user/account/profit_table/profit_table.ui.js
```diff
@@ -79,11 +79,21 @@ var ProfitTableUI = (function(){
         var data = [jpClient ? toJapanTimeIfNeeded(transaction.purchase_time) : profit_table_data.buyDate, '<span' + showTooltip(profit_table_data.app_id, oauth_apps[profit_table_data.app_id]) + '>' + profit_table_data.ref + '</span>', jpClient ? format_money_jp(TUser.get().currency, profit_table_data.payout) : profit_table_data.payout , '', jpClient ? format_money_jp(TUser.get().currency, profit_table_data.buyPrice) : profit_table_data.buyPrice , (jpClient ? toJapanTimeIfNeeded(transaction.sell_time) : profit_table_data.sellDate), jpClient ? format_money_jp(TUser.get().currency, profit_table_data.sellPrice) : profit_table_data.sellPrice , jpClient ? format_money_jp(TUser.get().currency, profit_table_data.pl) : profit_table_data.pl , ''];
         var $row = Table.createFlexTableRow(data, cols, "data");
 
-        $row.children(".buy-date").addClass("pre");
         $row.children(".pl").addClass(plType);
-        $row.children(".sell-date").addClass("pre");
         $row.children(".contract").html(profit_table_data.desc + "<br>");
 
+        if (japanese_client()) {
```
jy-binary: these kind of code exceptions complicate the codebase. Can the page not be refactored to work in Japanese without any code hack?

negar-binary: The problem is that the line doesn't break for japanese because they don't have the GMT in those two columns, so this exception should only be applied to japan

src/javascript/binary/websocket_pages/user/account/profit_table/profit_table.ui.js
```diff
@@ -79,11 +79,21 @@ var ProfitTableUI = (function(){
...
+            $row.children(".buy-date").each(function() {
+                $(this).wrapInner('<div class="new-width"></div>');
+            });
+            $row.children(".sell-date").each(function() {
+                $(this).wrapInner('<div class="new-width"></div>');
+            });
+        } else {
+            $row.children(".buy-date").addClass("pre");
+            $row.children(".sell-date").addClass("pre");
```
negar-binary: @jy-binary and if the class `pre` gets added, then line doesn't break at all for japanese and the column get so wide that the column for explanation becomes too narrow

negar-binary:
@jy-binary exceptions have been removed, it's working with other languages as well

PR #1574, opened 2016-09-22
negar/transfer_if_mlt_mf

negar-binary:
only show transfer button if user has both mlt and mf enabled

### DIFF
--- a/src/javascript/binary/base/page.js
+++ b/src/javascript/binary/base/page.js
@@ -854,6 +854,20 @@ Contents.prototype = {
             $('.unbind_later').off();
         }
     },
+    has_gaming_financial_enabled: function() {
+        var has_financial = false,
+            has_gaming = false,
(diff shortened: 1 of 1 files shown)


src/javascript/binary/base/page.js
```diff
@@ -854,6 +854,20 @@ Contents.prototype = {
             $('.unbind_later').off();
         }
     },
+    has_mlt_mf_enabled: function() {
```
jy-binary: I don't like this kind of code. MLT and MF are broker codes. Business-logic should depend on the *landing company* not the broker code. Do we not have the landing company easily available in the front-end?

negar-binary: @jy-binary it was a poor choice of naming on my part. It's not actually checking broker codes. I have renamed the function and variables.

PR #1576, opened 2016-09-23
negar/handle_company_name

negar-binary:
in case no landing company name detected


src/javascript/binary/websocket_pages/user/tnc_approval.js
```diff
@@ -38,8 +38,9 @@ var TNCApproval = (function() {
         $('#tnc-loading').addClass(hiddenClass);
         $('#tnc_image').attr('src', page.url.url_for_static('images/pages/cashier/protection-icon.svg'));
         $('#tnc_approval').removeClass(hiddenClass);
+        var domain = StringUtil.toTitleCase(document.domain.split('.').slice(-2).join('.'));
         var tnc_message = template($('#tnc-message').html(), [
-            page.client.get_storage_value('landing_company_name'),
+            page.client.get_storage_value('landing_company_name') || domain,
```
jy-binary: but why would landing_company_name not be set?

negar-binary: @jy-binary it's a race condition. I've fixed it another way.

PR #1582, opened 2016-09-26
mohammad/endpoint_notification

binary-hamedanchi:


scripts/bin/GenerateStaticData.pm
```diff
@@ -802,6 +802,9 @@ sub _texts {
         push @texts, localize('Deposit [_1] [_2] virtual money into your account [_3]');
         push @texts, localize('Withdraw');
 
+        # strings for endpoint notification
+        push @texts, localize('This is a staging server - For testing purposes only - The server <a href="[_1]">endpoint</a> is: [_2]');
```
jy-binary: you can warn also on production. Basically we need to warn whenever the endpoint has been changed manually and isn't equal to the default value.

binary-hamedanchi: Sure, I will change the condition and message accordingly.

PR #1620, opened 2016-09-29
negar/whyus_jp_fix_text

negar-binary:
change duplicated line


ashkanx:
I merge this, but you need to update the weblate again :)

PR #1681, opened 2016-10-07
negar/static_cashier_jp

negar-binary:
add deposit and withdraw page for japan


scripts/config/pages.pl
```diff
@@ -54,9 +54,12 @@ sub all_pages {
         ['get-started-jp', 'japan/get_started',  'default', 'Get Started', 'NOT-ja'],
 
         ## ws
-        ['user/authenticatews',          'user/authenticatews',                       'default', 'Authenticate', 'ja'],
-        ['cashier/forwardws',            'cashier/deposit_withdraw_ws',               'default', 'Cashier'],
-        ['cashier/epg_forwardws',        'cashier/deposit_withdraw_ws',               'default', 'Cashier'],
+        ['user/authenticatews',          'user/authenticatews',                       'default', 'Authenticate',    'ja'],
+        ['cashier/forwardws',            'cashier/deposit_withdraw_ws',               'default', 'Cashier',         'ja'],
+        ['cashier/epg_forwardws',        'cashier/deposit_withdraw_ws',               'default', 'Cashier',         'ja'],
+        ['cashier/deposit-jp',           'japan/cashier/deposit',                     'default', 'Deposit',         'NOT-ja'],
```
negar-binary: @jy-binary means exclude compiling this page for any language that is not Japanese. Because this page is only used in Japan. So we shouldn't compile it for other languages.

PR #1740, opened 2016-10-14
negar/check_end_time

negar-binary:
show error if end time is before start time


negar-binary:
@raunakkathuria if 'now' is selected, we take current server time. it shouldn't cause any issues, please let me know if I'm missing something.

PR #1886, opened 2016-11-23
Alex/some refactoring

AlexShpak:


src/javascript/binary/websocket_pages/trade/analysis.js
```diff
@@ -239,5 +237,5 @@ var TradingAnalysis = (function() {
 })();
 
 module.exports = {
-    TradingAnalysis: TradingAnalysis,
```
negar-binary: @AlexShpak generally we keep this comma. in case anyone adds anything under this, they need to add comma to the line above and it shows as change in github, or causes error if they miss out on adding this comma. I think it should be better to keep it, but it's up to you.

src/javascript/binary/websocket_pages/trade/tick_trade.js
```diff
@@ -314,36 +314,20 @@ WSTickDisplay.dispatch = function(data) {
...
-  if (data.tick) {
-    spots2 = Tick.spots();
-    epoches = Object.keys(spots2).sort(function(a, b) {
-        return a - b;
-    });
-  } else if (data.history) {
-    epoches = data.history.times;
+  
+  if (data.history) {
+      ticks = Tick.history2ticks(data.history);
```
binary-hamedanchi: @AlexShpak Is it needed to change the `var epoches, ...` to `var ticks, ...` as well?

AlexShpak: fixed

PR #1909, opened 2016-12-05
Using ESLint

binary-hamedanchi:


.eslintrc
```diff
@@ -0,0 +1,94 @@
...
+        'processTick_Beta':                false,
+        'processProposal_Beta':            false,
+        'processTradingTimesRequest_Beta': false,
+        'processTradingTimes_Beta':        false,
+    },
+    'rules': {
+        'array-callback-return': 0,
+        'func-names':  ['error', 'never'],
+        'indent':      ['error', 4],
+        'key-spacing': ['error', { align: 'colon' }],
```
AlexShpak: don't think it's a good idea to align colons. In my opinion, such code looks nasty

binary-hamedanchi: This alignment makes the code more readable in different editors/IDEs. Also makes it consistent to the assignment sign alignment.

ashkanx:
please fix the conflicts

PR #1910, opened 2016-12-05
kelly/caps

kellybinary:
remove all capitalization


ashkanx:
Please fix the conflicts

negarn:
@kellybinary @binary-hamedanchi I think we can remove all instances of `text-transform`, no?

src/javascript/binary/websocket_pages/trade/price.js
```diff
@@ -202,7 +202,7 @@ const Price = (function() {
                     elementTextContent(h4, Content.localize().textSpreadTypeShort);
                 }
             } else {
-                elementTextContent(h4, display_text);
+                elementTextContent(h4, display_text.replace(/\b\w/g, l => l.toUpperCase())); // capitalise each word
```
negarn: I think this will break translations?

kellybinary: @negarn any ideas on how we can capitalised this? `display_text` is lowercased, but in the UI, we need it to be capitalised. maybe we can use css text-transform ?

PR #1928, opened 2016-12-09
remove user-testing

ashkanx:


src/templates/partners/user_testing.html.tt
```diff
@@ -12,24 +12,7 @@
         [% l("Register to become a tester using the links below:") %]
```
negarn: @ashkanx I think we should change `links` to `link` since now there is only one link to choose

src/templates/partners/user_testing.html.tt
```diff
@@ -12,24 +12,7 @@
...
-                <br/>
-                <span>
-                  [% l("Sign up as a tester and look for Binary.com-related projects.") %]
-                </span>
-              </div>
-            </div>
-          </div>
-        </div>
+
         <div class="gr-6 gr-12-m gr-6-p">
```
negarn: @ashkanx and the `gr-6` and `gr-6-p`should be changed or centered, now that there's only one element

src/templates/partners/user_testing.html.tt
ashkanx: Yes. Reuben will give me a new text for this page.

src/templates/partners/user_testing.html.tt
```diff
@@ -1,52 +1,44 @@
...
+                  [% l("Explore our HackerOne bounty programme") %]
                 </a>
               </div>
             </div>
           </div>
         </div>
       </div>
       <p>
-        [% l("Earn money for any verifiable errors that you find.") %]
+        [% l("Earn up to USD 1,000 for finding high-risk security vulnerabilities.") %]
```
jy-binary: it looks weird to have that line at the bottom on its own. I think you need to move it up into the Benefits section.

PR #1942, opened 2016-12-14
negar/onclick_refactor

negarn:
remove global onclick handlers


src/javascript/binary/websocket_pages/trade/beta/duration.js
```diff
@@ -1,5 +1,6 @@
 var Barriers_Beta    = require('./barriers').Barriers_Beta;
```
jy-binary: what does the word "Beta" refer to?

negarn: @jy-binary it's code from the trading_beta page.

jy-binary:
please check conflicts

PR #1961, opened 2016-12-16
negar/webpack

negarn:
continue webpack refactoring


negarn:
@aminmarashi you chose the worst PR to comment that on :D I'm working on refactoring page.js which is the most common file in binary-static; functions of which are used in at least 70+ other files; so each refactoring of a function to bring it out of page will cause changes in 70+ files, making each commit large automatically :)

aminmarashi:
True, I didn't read the whole commit. 👍

PR #1962, opened 2016-12-18
Shashank/fixing html injection + content spoofing

Shashank-In:
🐛 🐛 😬


src/javascript/binary/websocket_pages/user/account/settings/iphistory/iphistory.ui.js
```diff
@@ -44,6 +44,12 @@ var IPHistoryUI = (function() {
         var browserString = browser ?
             browser.name + ' v' + browser.version :
             'Unknown';
+        var patt = (/(opera|chrome|safari|firefox|IE|Edge|SeaMonkey|Chromium) v[0-9.]+/gi);
```
binarytom: Not sure where we're setting the version information, but might want to anchor the regex here:

```
var patt = /^(opera|chrome|safari|firefox|IE|Edge|SeaMonkey|Chromium) v[0-9.]+$/i;
```

to guard against things like "This browser is not really chrome v123 but try our site at www.evilsite.com"

Shashank-In: Ahh my bad . I will fix this

PR #1980, opened 2016-12-22
negar/highcharts

negarn:
update highcharts and import via npm


ashkanx:
**24,047** lines :1st_place_medal:

PR #1999, opened 2016-12-28
negar/fix_message

negarn:
don't show unless no data


negarn:
@ashkanx if possible, it would be good, but don't think many clients will go through the scenario to reproduce this.

PR #2094, opened 2017-02-03
kelly/real_account

kellybinary:
Add hint under first line address field.


negarn:
@ashkanx should we include this in financial account opening page as well? could be VRTC upgrading straight to MF, won't see this realws page

ashkanx:
Yes. let's do that @kellybinary

PR #2150, opened 2017-02-20
kelly/rate_limit

kellybinary:
Don't update balance if reached rate limit.


src/javascript/binary/websocket_pages/socket.js
```diff
@@ -300,7 +300,7 @@ const BinarySocketClass = function() {
                         }
                         sendBufferedSends();
                     }
-                } else if (type === 'balance') {
+                } else if ((type === 'balance') && (error_code !== 'RateLimit')) {
```
negarn: @kellybinary why don't we check for error in general, not just ratelimit?

PR #2163, opened 2017-02-23
kelly/pushwoosh

kellybinary:


jy-binary:
remember to delete all the OneSignal-related code

ashkanx:
@jy-binary it was removed and released in another PR

PR #2200, opened 2017-03-02
kelly/refactoring

kellybinary:
Refactor modules


src/javascript/binary/websocket_pages/cashier/payment_agent_list.js
```diff
@@ -1,31 +1,30 @@
-const showLoadingImage = require('../../base/utility').showLoadingImage;
-const Cookies          = require('../../../lib/js-cookie');
 const Content          = require('../../common_functions/content').Content;
+const Cookies          = require('../../../lib/js-cookie');
```
negarn: I think Cookies can move to the last line since it's three folders up, @binary-hamedanchi is that right?

src/javascript/binary/websocket_pages/cashier/payment_agent_list.js
```diff
@@ -1,120 +1,115 @@
...
                 collapsible: true,
                 active     : false,
             });
         });
 
-        hiddenClass = 'hidden';
-        ddlCountriesID = '#target_country';
-        $paListContainer = $('#pa_list');
-        agentTemplate = $paListContainer.find('#accordion').html();
+        ddl_countries_id   = '#target_country';
```
negarn: You can declare this value above, similar to hidden_class, since it's a string and not a DOM or jQuery object

src/javascript/binary/websocket_pages/user/account/profit_table/profit_table.init.js
```diff
@@ -106,33 +100,32 @@ const ProfitTableWS = (function() {
...
+        if (opts) $.extend(true, req, opts);
+
+        BinarySocket.send(req).then((response) => {
+            ProfitTable.profitTableHandler(response);
+            showLocalTimeOnHover('td.buy-date,td.sell-date');
+            $('.barspinner').addClass('hidden');
+        });
+    };
+
+    const initSocket = () => {
```
negarn: can you change the name of this function? it's not initializing socket anymore, something like get apps or similar would be more appropriate

src/javascript/binary/websocket_pages/user/account/statement/statement.init.js
```diff
@@ -83,7 +81,7 @@ const StatementWS = (function() {
             StatementUI.updateStatementTable(getNextChunkStatement());
 
             // Show a message when the table is empty
-            if ((transactionsReceived === 0) && (currentBatch.length === 0)) {
+            if ((transactions_received === 0) && (current_batch.length === 0)) {
```
negarn: we don't need the extra brackets around the two checks

src/javascript/binary/websocket_pages/user/account/profit_table/profit_table.init.js
kellybinary: would it be better to just move it under onLoad() instead? since onLoad is calling that

negarn: up to you, either will be fine

kellybinary: alrighty

src/javascript/binary/websocket_pages/user/account/profit_table/profit_table.init.js
```diff
@@ -106,18 +102,12 @@ const ProfitTable = (() => {
         if (opts) $.extend(true, req, opts);
 
         BinarySocket.send(req).then((response) => {
-            ProfitTable.profitTableHandler(response);
+            ProfitTableInit.profitTableHandler(response);
```
negarn: is this referring to a local function? if so it doesn't need to call the module name

src/javascript/binary/websocket_pages/cashier/payment_agent_list.js
```diff
@@ -1,120 +1,114 @@
...
-        cList.map(function(country) {
-            if (country === requestedCountry) {
+        countries.map((country) => {
+            if (country === requested_country) {
                 found = true;
             }
-            insertListOption($ddlCountries, country[1], country[0]);
+            insertListOption($ddl_countries, country[1], country[0]);
+            $('#target_country').removeClass(hidden_class);
+            $('.barspinner').addClass(hidden_class);
```
binary-hamedanchi: Could we move these 2 lines to outside the loop, since we only need to add/remove class once?

kellybinary: referring to line 68 and 69?

binary-hamedanchi: Yes

PR #2229, opened 2017-03-10
Update MT Terms and Conditions to Version 4

kellybinary:


src/templates/legal/metatrader/terms_and_conditions.html.tt
```diff
@@ -29,347 +29,349 @@
...
         <li>[% l("If any provision of this Agreement shall be held invalid or unenforceable by a court or regulatory body of competent jurisdiction, the remainder of this Agreement shall remain in
```
negarn: any reason why we are keeping these commented lines?

kellybinary: @negarn no reason, i'll remove it

PR #2230, opened 2017-03-13
Fix validations

kellybinary:


src/javascript/binary/common_functions/account_opening.js
```diff
@@ -146,14 +146,14 @@ const commonValidations = () => {
         { selector: '#first_name',         validations: ['req', ['length', { min: 2, max: 30 }], 'letter_symbol'] },
```
binary-hamedanchi: Please move the `'letter_symbol'` next to `'req'`.

src/javascript/binary/common_functions/account_opening.js
```diff
@@ -146,14 +146,14 @@ const commonValidations = () => {
...
         { selector: '#last_name',          validations: ['req', ['length', { min: 2, max: 30 }], 'letter_symbol'] },
```
binary-hamedanchi: Same here.

PR #2245, opened 2017-03-15
kelly/pushwoosh

kellybinary:
Exclude all except staging and production

### DIFF
--- a/src/javascript/lib/push_notification.js
+++ b/src/javascript/lib/push_notification.js
@@ -9,6 +9,9 @@ const BinaryPushwoosh = (() => {
     let initialised = false;
 
     const init = () => {
+        const hostname = window.location.hostname;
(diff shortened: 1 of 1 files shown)


src/javascript/lib/push_notification.js
```diff
@@ -9,6 +9,9 @@ const BinaryPushwoosh = (() => {
     let initialised = false;
 
     const init = () => {
+        const hostname = window.location.hostname;
+        if (!(hostname.match(/^(www.binary.com|staging.binary.com)$/))) return;
```
binary-hamedanchi: What do you think about rewriting the above two lines to:
```js
if (!/^(www|staging)\.binary\.com$/.test(window.location.hostname)) return;
```

kellybinary: 👍

PR #2264, opened 2017-03-20
kelly/fix

kellybinary:
Fix numbered list


src/sass/static/metatrader/terms.scss
```diff
@@ -0,0 +1,50 @@
+/* Multilevel numbered list */
+#mt-tnc {
```
binary-hamedanchi: Is there any occurrences of `id='mt-tnc'` in templates?

kellybinary: thanks for catching that, forgot to include the changes in the template

PR #2393, opened 2017-04-27
Added 'Please select' option during new account creation

fanxuricky:


src/javascript/binary/common_functions/account_opening.js
```diff
@@ -162,7 +163,7 @@ const AccountOpening = (() => {
             id;
         $(form_id).find('select, input[type=checkbox]').each(function() {
             id = $(this).attr('id');
-            if (id !== 'tnc') {
+            if (id !== 'tnc' && id !== 'address_state') {
```
binary-hamedanchi: What about:
```js
!/^(tnc|address_state)$/.test(id)
```

fanxuricky: Is it to replace:
```
id !== 'tnc' && id !== 'address_state'
```

binary-hamedanchi: Yes.

fanxuricky: Thanks for your suggestion.

PR #2409, opened 2017-05-05
kelly/display_asterisk

kellybinary:
Display * on required field


src/templates/user/settings/personal_details.html.tt
```diff
@@ -11,7 +11,7 @@
 
             <div class="gr-padding-10 gr-row rowName clear invisible RealAcc JpAcc">
                 <div class="gr-4 form_label">
-                    <label><em class="required_asterisk">**</em> [% l('Name') %]</label>
+                    <label>[% l('Name') %]</label>
```
binary-hamedanchi: Removing the class from css as well?

kellybinary: @binary-hamedanchi I removed it :) thanks!

PR #2502, opened 2017-06-05
negar/japan_epg

negarn:
remove old deposit and withdraw for cashier and replace all cashier links with epg links


scripts/sitemap.pl
```diff
@@ -45,6 +45,7 @@
     ['security-testing'       , 'monthly', '0.80'],
 
     ['cashier'                     , 'monthly', '0.80'],
+    ['cashier/epg_forwardws'       , 'monthly', '0.80', 'NOT-ja'],
```
binary-hamedanchi: We can probably omit this from sitemap, as it's an authenticated page.
Thanks.

PR #2579, opened 2017-07-07
negar/auto_refresh_crypto_cashier_deposit

negarn:
auto refresh crypto cashier deposit page every 5 minutes


ashkanx:
@negarn why should we refresh the cashier?

negarn:
@ashkanx the idea is that we don't want them to reuse the cryptocurrency address, so if they make a deposit, go away for ten minutes and come back we don't want to show them the same address in case they deposit money into the same address by accident.

PR #2655, opened 2017-07-28
Compliance changes

4p00rv:


src/templates/global/financial_form.html.tt
```diff
@@ -54,6 +54,24 @@
...
+        <option value="Investments &amp; Dividends">[% l("Investments &amp; Dividends") %]</option>
+        <option value="Pension">[% l("Pension") %]</option>
+        <option value="Other">[% l("Other") %]</option>
+    [% END %]
+
+    [% WRAPPER form_row id='employment_status' label=l("Employment Status") type='select' %]
+        <option value="">[% l("Please select") %]</option>
+        <option value="Employed">[% l("Employed") %]</option>
+        <option value="Pensioner">[% l("Pensioner") %]</option>
+        <option value="Self Employed">[% l("Self Employed") %]</option>
```
kellybinary: i think this should be Self-Employed (with a dash)

4p00rv: Yep, just checked the PR. Thanks for pointing out. 😄

PR #2691, opened 2017-08-07
Telegram bot

4p00rv:


scripts/config/pages.pl
```diff
@@ -112,6 +112,7 @@ sub all_pages {
         ['user/settings/assessmentws',       'user/settings/financial_assessment',   'default',    'Financial Assessment',    'ja'],
         ['user/settings/detailsws',          'user/settings/personal_details',       'default',    'Personal Details'],
         ['user/set-currency',                'user/set_currency',                    'default',    'Set Currency',            'ja'],
+        ['user/telegram-bot',                 'user/telegram_bot',                    'default',    'Telegram Bot'],
```
negarn: could you match the indentation with above fields, please?

src/javascript/binary/websocket_pages/user/telegram_bot.js
```diff
@@ -0,0 +1,26 @@
+const TelegramBot = (() => {
+    'use strict';
+
+    const onLoad = () => {
+        const bot_name = 'binary_test_bot';
+        const submit_button = $('#go_to_telegram');
```
negarn: can we make a global variable within this file `const submit_button = '#go_to_telegram';` and use it on line 7 and 17 as `$(submit_button)` instead of declaring the same value twice?

4p00rv: @negar-binary Sure.

src/templates/user/telegram_bot.html.tt
```diff
@@ -0,0 +1,12 @@
+[% PROCESS components/forms.html.tt %]
+
+<h1>[% l('Telegram Bot') %]</h1>
+
+<p>[% l("In order to access <strong>Telegram Bot</strong>, you'll first need to generate an API Token then enter it in the input box below.") %]</p>
```
negarn: is this the final (approved) wording? to me if we add and `and` to make it `you'll first need to generate an API Token and then enter it` or any such connective, it makes it flow better

src/javascript/binary/websocket_pages/user/telegram_bot.js
```diff
@@ -0,0 +1,26 @@
...
+        submit_button.click((e) => {
```
negarn: is it better to have an onclick event on the button or onsubmit event on the form? then enter will also trigger submit event, which seems more accessible

src/templates/user/telegram_bot.html.tt
4p00rv: No, it is not. I will make the change.

PR #2768, opened 2017-08-28
Authentication

4p00rv:


src/javascript/binary/base/header.js
```diff
@@ -196,6 +196,7 @@ const Header = (() => {
 
             const messages = {
                 authenticate   : () => buildMessage('[_1]Authenticate your account[_2] now to take full advantage of all payment methods available.',     'user/authenticate'),
+                under_review   : () => buildMessage('We are reviewing your account. For more details [_1]contact us[_2].',                                'contact'),
```
binary-hamedanchi: Please add the new string to `scripts/bin/GenerateStaticData.pm`.

src/javascript/binary/websocket_pages/user/account/settings.js
```diff
@@ -20,6 +20,10 @@ const Settings = (() => {
                 $('#change_password').removeClass(class_hidden);
             }
 
+            if (!response.get_account_status.prompt_client_to_authenticate) {
+                $('#authenticate').addClass(class_hidden);
```
binary-hamedanchi: What about `removeClass` and reverse condition? as it's already invisible in the template.

src/templates/user/authenticate.html.tt
```diff
@@ -14,6 +14,11 @@
     <div id="not_authenticated_financial" class="invisible">
         [% INCLUDE global/authenticate_message_financial.html.tt %]
     </div>
+    <div id="success-message" class="center-text invisible">
+        [% INCLUDE global/separator_line.html.tt invisible=1 class="gr-padding-10" %]
+        <h2>Thank you for submitting the required documents</h2>
```
binary-hamedanchi: `l('...')` is missing for translation.

src/templates/user/authenticate.html.tt
```diff
@@ -14,6 +14,11 @@
...
+        <p>We are reviewing your documents and will get back to you within one working day.</p>
```
binary-hamedanchi: Same here.

src/javascript/binary/base/header.js
4p00rv: :+1:

src/javascript/binary/base/header.js
```diff
@@ -205,6 +206,7 @@ const Header = (() => {
 
             const validations = {
                 authenticate   : () => +get_account_status.prompt_client_to_authenticate,
+                under_review   : () => get_account_status.status.indexOf('under_review') !== -1,
```
negarn: the functions here are ordered alphabetically, could you move it after tnc, before unwelcome?

4p00rv: Ok. 👍

src/javascript/binary/websocket_pages/user/account/authenticate.js
```diff
@@ -1,5 +1,7 @@
 const Client       = require('../../../base/client');
 const BinarySocket = require('../../socket');
```
negarn: can you align line 1 & 2 with 3 & 4? (by equal sign)

src/javascript/binary/websocket_pages/user/account/authenticate.js
```diff
@@ -23,6 +26,192 @@ const Authenticate = (() => {
...
+        const submitFiles = ($files) => {
+            // Disable submit button
+            disableButton();
+            const files = [];
+            $files.each((i, e) => {
+                if (e.files && e.files.length) {
+                    const $e = $(e);
+                    const id = $e.attr('id');
+                    const type = `${($e.attr('data-type') || '').replace(/\s/g, '_').toLowerCase()}_${id.split('_')[0]}`;
+                    const $inputs = $e.parent().parent().parent().find('input[type="text"]');
```
negarn: is there any way to use class or id identifier instead of calling so many `parent()`s? in case structure changes later

src/javascript/binary/websocket_pages/user/account/authenticate.js
```diff
@@ -23,6 +26,192 @@ const Authenticate = (() => {
...
+        const showError = (e) => {
+            const $error = $('.error-msg');
+            const message = e.message || e.message_to_client;
+            enableButton();
+            $error.removeClass('invisible').text(message);
+            setTimeout(() => { $error.empty().addClass('invisible'); }, 3000);
+        };
+
+        const showSuccess = () => {
+            $('#authentication-message > div').not('.invisible').addClass('invisible');
```
negarn: anywhere in the code, `addClass('invisible')` can be replaced with `setVisibility(0)`

src/javascript/binary/websocket_pages/user/account/authenticate.js
```diff
@@ -23,6 +26,192 @@ const Authenticate = (() => {
...
+            $('#success-message').removeClass('invisible');
```
negarn: anywhere in the code, `removeClass('invisible')` can be replaced with `setVisibility(1)`

src/sass/pages/static_pages.scss
```diff
@@ -334,3 +334,37 @@ body #not_authenticated_financial ul.checked > li {
...
+                width: 1em;
+                height: 1em;
+                margin-left: 10px;
+                margin-top: -2px;
+
+                .add {
+                    content: url('../images/pages/authenticate/add.svg');
+                }
+                .remove {
+                    content: url('../images/pages/authenticate/delete/Red.svg');
```
negarn: can we have the file name something like `remove.svg` to keep it consistent?

src/javascript/binary/websocket_pages/user/account/authenticate.js
```diff
@@ -133,8 +133,7 @@ const Authenticate = (() => {
             $files.each((i, e) => {
                 if (e.files && e.files.length) {
                     const $e = $(e);
-                    const id = $e.attr('id');
-                    const type = `${($e.attr('data-type') || '').replace(/\s/g, '_').toLowerCase()}_${id.split('_')[0]}`;
+                    const type = `${($e.attr('data-type') || '').replace(/\s/g, '_').toLowerCase()}`;
```
negarn: none of the `data-type` values have any space, is this just for forward-proofing?

src/templates/global/authenticate_message.html.tt
```diff
@@ -81,7 +81,10 @@
         l('Only JPG, JPEG, GIF, PNG and PDF formats are accepted'), 
         l('Maximum upload size for each file is 3MB')]
     type="poi"
-    accepted_documents=[l('Passport'), l('Identity card'), l('Driving license')]
+    accepted_documents=[{name => l('Passport'), value => 'passport'},
```
binary-hamedanchi: Nothing serious, but appreciate it if you move the first index to next line and vertically align keys for better readability.
Thanks.

build/aliases.yaml
```diff
@@ -3,6 +3,9 @@ css:
   - 'postcss'
   - 'cssmin'
 
+svg:
```
negarn: how is this to be used? what's the difference between using this and simply typing `svgo [...]`?

src/javascript/binary/base/header.js
```diff
@@ -254,25 +254,27 @@ const Header = (() => {
...
-                risk           : () => buildMessage('Please complete the [_1]financial assessment form[_2] to lift your withdrawal and trading limits.',  'user/settings/assessmentws'),
-                tax            : () => buildMessage('Please [_1]complete your account profile[_2] to lift your withdrawal and trading limits.',           'user/settings/detailsws'),
-                tnc            : () => buildMessage('Please [_1]accept the updated Terms and Conditions[_2] to lift your withdrawal and trading limits.', 'user/tnc_approvalws'),
-                unwelcome      : () => buildMessage('Your account is restricted. Kindly [_1]contact customer support[_2] for assistance.',                'contact'),
+                authenticate         : () => buildMessage('[_1]Authenticate your account[_2] now to take full advantage of all payment methods available.',     'user/authenticate'),
+                document_under_review: () => buildMessage('We are reviewing your account. For more details [_1]contact us[_2].',                                'contact'),
```
negarn: could we rename `document_under_review` to something shorter? maybe `document_review`, it still gets the same message across and it's not so long, so it doesn't push everything out.

src/javascript/binary/base/header.js
```diff
@@ -254,25 +254,27 @@ const Header = (() => {
...
-                tnc            : () => Client.shouldAcceptTnc(),
-                unwelcome      : () => /unwelcome|(cashier|withdrawal)_locked/.test(status),
+                authenticate         : () => +get_account_status.prompt_client_to_authenticate,
+                currency             : () => !Client.get('currency'),
+                financial_limit      : () => /ukrts_max_turnover_limit_not_set/.test(status),
+                residence            : () => !Client.get('residence'),
+                risk                 : () => riskAssessment(),
+                tax                  : () => Client.shouldCompleteTax(),
+                tnc                  : () => Client.shouldAcceptTnc(),
+                document_under_review: () => get_account_status.status.indexOf('document_under_review') !== -1,
```
negarn: `get_account_status.status` is already stored under the variable `status`, so you can use that, and in all other functions here we are using a regex to test, so it should be written as `/document_under_review/.test(status)`, or we should change everything else to the `indexOf` check instead to make it consistent. I vote for the regex test.

src/javascript/binary/websocket_pages/user/account/authenticate.js
```diff
@@ -1,18 +1,27 @@
-const Client       = require('../../../base/client');
-const BinarySocket = require('../../socket');
+const Client              = require('../../../base/client');
+const BinarySocket        = require('../../socket');
+const DocumentUploader    = require('binary-document-uploader');
+const showLoadingImage    = require('../../../base/utility').showLoadingImage;
+const localize            = require('../../../base/localize').localize;
+const displayNotification = require('../../../base/header').displayNotification;
+const Url                 = require('../../../base/url');
```
negarn: can we order these based on how far they are from current file? e.g. line 1 and line 3 should swap, line 4 should go to the end (alphabetic ordering of pathname, `utility` goes after `url`, and line 6 should go to line 4

src/javascript/binary/websocket_pages/user/account/authenticate.js
```diff
@@ -1,18 +1,27 @@
...
 
 const Authenticate = (() => {
     const onLoad = () => {
         BinarySocket.send({ get_account_status: 1 }).then((response) => {
             if (response.error) {
                 $('#error_message').setVisibility(1).text(response.error.message);
             } else {
-                const get_account_status  = response.get_account_status;
```
negarn: why the space removal? they were aligned by `=`, I think that's fine to keep

src/javascript/binary/websocket_pages/user/account/authenticate.js
```diff
@@ -1,18 +1,27 @@
...
+                const get_account_status = response.get_account_status;
                 const should_authenticate = +get_account_status.prompt_client_to_authenticate;
                 if (should_authenticate) {
                     const status = get_account_status.status;
                     if (!/authenticated/.test(status)) {
-                        $(`#not_authenticated${Client.isAccountOfType('financial') ? '_financial' : ''}`).setVisibility(1);
+                        $('#not_authenticated').setVisibility(1);
+                        if (Client.isAccountOfType('financial')) {
+                            $('#not_authenticated_financial').setVisibility(1);
+                        }
```
negarn: why change the one line code to this? it's not making a difference in functionality

src/javascript/binary/websocket_pages/user/account/authenticate.js
```diff
@@ -1,18 +1,27 @@
...
+                        init();
```
negarn: can we call init first? i.e on line 20, so it sets everything up before making the section visible

src/javascript/binary/websocket_pages/user/account/authenticate.js
```diff
@@ -23,6 +32,236 @@ const Authenticate = (() => {
...
+    const init = () => {
+        // Setup accordion
+        $('.files').accordion({
+            heightStyle: 'content',
+            collapsible: true,
+            active     : false,
+        });
+        // Setup Date picker
+        const tomorrow = new Date();
+        tomorrow.setDate(tomorrow.getDate() + 1);
```
negarn: can we use something like 

```
const tomorrow = window.time ? new Date(window.time.valueOf()) : new Date();
tomorrow.setDate(tomorrow.getDate() + 1);
```

similar to what we are using in duration file for trade page? as we already have time stored in window

src/javascript/binary/websocket_pages/user/account/authenticate.js
```diff
@@ -23,6 +32,236 @@ const Authenticate = (() => {
...
+        $('.date-picker').datepicker({
+            dateFormat : 'yy-mm-dd',
+            changeMonth: true,
+            changeYear : true,
+            minDate    : tomorrow,
+        });
+
+        // Submit button
+        let $button;
```
negarn: can we define this somewhere above `showSubmit` function as it's not used anywhere from line 53 to 101?

PR #2800, opened 2017-09-12
zaki/updated_addressline_regex

zakibinary:
Updated addressline regex


src/javascript/binary/common_functions/form_validation.js
```diff
@@ -9,6 +9,7 @@ const compareBigUnsignedInt = require('../common_functions/string_util').compare
 const Validation = (() => {
     const forms        = {};
     const error_class  = 'error-msg';
+    const error_address_class = 'error-address-msg';
```
negarn: can you align line 10 to 13 by the `=` sign?

zakibinary: Will do.

src/javascript/binary/websocket_pages/user/account/settings/personal_details.js
```diff
@@ -311,6 +311,18 @@ const PersonalDetails = (() => {
...
+                heightStyle: 'content',
+                collapsible: true,
+                active     : false,
+            });
+            $('#form_addressline2_message').accordion({
+                heightStyle: 'content',
+                collapsible: true,
+                active     : false,
+            });
+        });
```
negarn: instead of repeating the same lines in all these files we can add it in a common file and call the function everyone, probably account_opening.js under common_functions folder would be a good place

src/javascript/binary/common_functions/form_validation.js
```diff
@@ -73,6 +74,7 @@ const Validation = (() => {
                             $parent.append($('<div/>', { class: `${error_class} ${hidden_class}` }));
                         }
                         field.$error = $parent.find(`.${error_class}`);
+                        field.$error_address = $parent.find(`.${error_address_class}`);
```
negarn: why are we setting this for everyone? can we just set it for selectors address_line_1 and address_line_2?

src/javascript/binary/common_functions/form_validation.js
```diff
@@ -163,7 +165,7 @@ const Validation = (() => {
         email        : { func: validEmail,        message: 'Invalid email address' },
         password     : { func: validPassword,     message: 'Password should have lower and uppercase letters with numbers.' },
         general      : { func: validGeneral,      message: 'Only letters, numbers, space, hyphen, period, and apostrophe are allowed.' },
-        address      : { func: validAddress,      message: 'Only letters, numbers, space, hyphen, period, and apostrophe are allowed.' },
+        address      : { func: validAddress,      message: 'Invalid character used.' },
```
negarn: we need to add `Invalid character used.` to translation file

src/javascript/binary/common_functions/form_validation.js
```diff
@@ -234,11 +236,21 @@ const Validation = (() => {
...
+
+        if (field.$error_address && field.$error.length) {
+            field.$error_address.setVisibility(0);
+        }
     };
 
     const showError = (field, message) => {
-        clearError(field);
-        field.$error.text(localize(message)).setVisibility(1);
+        if (field.selector === '#address_line_1' || field.selector === '#address_line_2') {
```
negarn: this can be refactored to `if (/#address_line_[1|2]/.test(field.selector)) {`

src/javascript/binary/common_functions/form_validation.js
```diff
@@ -234,11 +236,21 @@ const Validation = (() => {
...
+            clearError(field);
+            field.$error.text(localize(message)).setVisibility(1);
+            field.$error_address.setVisibility(1);
+        } else {
+            clearError(field);
+            field.$error.text(localize(message)).setVisibility(1);
```
negarn: line 251 & 252 are repeated in both if and else (line 247 and 248). we can move them out and call it regardless of any condition.

src/sass/common.scss
```diff
@@ -122,6 +122,36 @@ label {
...
+    font-size: 13px !important;
+    font-style: italic !important;
+    color: $COLOR_BLACK !important;
+    background-color: transparent !important;
+    background-image: none !important;
+    border: 0 !important;
+    padding: 5px 0px 10px !important;
+}
+
+.form-validation-message h4:after {
```
negarn: this can be written under form-validation-message h4 as sass styling `&:after` on line 132 instead of a separate section since they have the same parent

src/sass/common.scss
```diff
@@ -122,6 +122,36 @@ label {
...
+    content: '\25BE';
+    font-style: normal;
+}
+
+.form-validation-message > .ui-accordion-header-collapsed {
```
negarn: same here, this can be a child after line 132

src/sass/common.scss
```diff
@@ -122,6 +122,36 @@ label {
...
+    margin-bottom: -15px !important;
+}
+
+.form-validation-desc {
+    padding: 0px 10px !important;
+    border: 0 !important;
+    background-color: $COLOR_LIGHT_GRAY;
+}
+
+.form-validation-desc span {
```
negarn: this can be a child after line 147

src/templates/components/forms.html.tt
```diff
@@ -40,6 +40,32 @@
...
+                        [% INCLUDE address_validation_header %]
+                    </h4>
+                    <div class = "gr-9 gr-10-m gr-push-1-m form-validation-desc">
+                        <div class="gr-row">
+                            [% INCLUDE address_validation_desc %]
+                          </div>
+                    </div>
+                </div>
+            [% END %]
+            [% IF id == 'address_line_2' %]
```
negarn: if the block address_line_1 and address_line_2 are exactly the same apart from their IDs, they can be added as a single block that accepts ID as an argument and called in these two places.

src/templates/components/forms_common_rows.html.tt
```diff
@@ -44,6 +44,26 @@
         hint=hint || (no_hint ? '' : l('Kindly provide your complete address.<br>This will be used to authenticate your identity when you open a real account.')) %]
 [% END %]
 
+[% BLOCK address_validation_header %]
+    <span>[% l("Please refer to the") %]<strong>&nbsp[% l("characters that can be used") %]</strong></span>
```
negarn: this line will not be possible for translators to translate properly, because they will see "Please refer to the" and "characters that can be used" as two separate sentences. it's better to use something like `[% l("Please refer to the [_1]characters that can be used", "<strong>") %]` if you want to avoid adding the tag in the sentence.

src/javascript/binary/common_functions/form_validation.js
zakibinary: Hi negarn, I'm not completely familiar with how the localization works for this project, do I have to add the lines below:
`msgid "Invalid character used."`
`msgstr ""`
to the en.po and all the .po files in the locales folder?

binary-hamedanchi: @zakibinary There is no need to update `.po` files manually.
The strings which used in javascript, need to be also included in `scripts/bin/GenerateStaticData.pm` to get involved in the translation process. So, adding new strings there is enough.

zakibinary: @binary-hamedanchi , noted with thanks!

src/javascript/binary/common_functions/form_validation.js
```diff
@@ -242,12 +247,19 @@ const Validation = (() => {
         if (field.$error && field.$error.length) {
             field.$error.setVisibility(0);
         }
+
+        if (field.$error_address && field.$error.length) {
```
negarn: @binary-hamedanchi should the second condition be `field.$error_address.length`?

src/templates/components/forms_common_rows.html.tt
```diff
@@ -44,6 +44,12 @@
         hint=hint || (no_hint ? '' : l('Kindly provide your complete address.<br>This will be used to authenticate your identity when you open a real account.')) %]
 [% END %]
 
+[% BLOCK address_validation_text %]
+    <span class="gr-12">
+        <p>[% l("You can use letters and numbers (A-Z, a-z, 0-9) and the following special characters: -.'#;:()") %]</p>
```
negarn: since we're not using accordions anymore, can we just add this as the error message for address field instead of the old message? or there is some other styling difference that I'm not aware of?

binary-hamedanchi: 1. I'm not sure mentioning `A-Z, a-z` here is a good idea, as it is allowed to use characters from other languages.
2. To put special characters in placeholder, prevents translation update in case of changing the characters list in future.
`<p>[% l("... and the following special characters: [_1]", "-.'#;:()") %]</p>`

Thanks.

src/javascript/binary/common_functions/form_validation.js
zakibinary: Noted, will make the change.

src/templates/components/forms_common_rows.html.tt
zakibinary: For 1) I'll revert back to Ashkan regarding this, as the text to be displayed was based on feedback we received.
I'll make with the changes for 2).

PR #2801, opened 2017-09-12
zaki/lastdigits_sync

zakibinary:
Synchronised the value selected in digits_underlying select-box with the value selected in underlying select-box.


src/javascript/binary/websocket_pages/trade/get_ticks.js
```diff
@@ -125,8 +125,18 @@ const GetTicks = (() => {
         }
     };
 
+    const populateDigits = () => {
+        const tick = $('#tick_count').val() || 100;
+        request('', {
+            ticks_history: underlying,
```
negarn: are we sure that the value of `underlying` is always the updated value that we are looking for? I think we should either keep the old way of `$('#digit_underlying option:selected').val() || $('#underlying').find('option:selected').val();` to get value for underlying or to look at the websocket request and make sure it is always requesting the expected underlying

zakibinary: Thanks for the feedback and noted, i'll try to alter the populateDigits function so that underlying will use the value based on where it's called from. So if from digits_underlying dropdown it'll use the value from `'$(#digit_underlying option:selected).val()`. Is this alright?

negarn: yes, I think we can pass the underlying value to this function, then it will be clear

PR #2821, opened 2017-09-15
kelly/update-messages

kellybinary:
Update messages


src/javascript/binary/websocket_pages/user/lost_password.js
```diff
@@ -6,7 +6,7 @@ const LostPassword = (() => {
 
     const responseHandler = (response) => {
         if (response.verify_email) {
-            $(form_id).html($('<div/>', { class: 'notice-msg', text: localize('Please check your email to complete the process.') }));
+            $(form_id).html($('<div/>', { class: 'notice-msg', text: localize('Please check your email for the password reset link.') }));
```
negarn: please update the text in GenerateStaticData.pm as well or translations won't be reflected

scripts/bin/GenerateStaticData.pm
```diff
@@ -427,7 +427,9 @@ sub _texts {
         # Strings for lost_password
         push @texts, localize('[_1] Please click the link below to restart the password recovery process.');
         push @texts, localize('Your password has been successfully reset. Please log into your account using your new password.');
-        push @texts, localize('Please check your email to complete the process.');
+        push @texts, localize('Please check your email for the verification link to complete the process.');
+        push @texts, localize('If you don’t receive the email within the next few minutes, please check your junk/spam folder.');
+        push @texts, localize('To reset your password, enter the email address you used to create your account into the field below and click \'Reset password\'.');
```
negarn: why do we need to add these three strings to this file? as per this PR they are only being used in templates and no .js files, or am I wrong?

src/javascript/binary/websocket_pages/user/lost_password.js
kellybinary: ok, noted

scripts/bin/GenerateStaticData.pm
kellybinary: ah yes! it's used in templates only

PR #2843, opened 2017-09-19
negar/change_to_js

negarn:
change jquery to js where performance of js is better


src/javascript/binary/base/language.js
```diff
@@ -48,7 +49,7 @@ const Language = (() => {
             const crowdin_lang = Cookies.get('jipt_language_code_binary-static'); // selected language for in-context translation
             if (crowdin_lang) {
                 current_lang = crowdin_lang.toUpperCase().replace('-', '_').toUpperCase();
-                $('body').addClass(current_lang); // set the body class removed by crowdin code
+                document.getElementsByTagName('body')[0].addClass(current_lang); // set the body class removed by crowdin code
```
kellybinary: shouldn't this be `document.getElementsByTagName('body')[0].classList.add(current_lang)` ?

negarn: yes, thank you

negarn: I suspect there are more cases I've missed, so would appreciate it if you could review with wide eyes :)

src/javascript/binary/base/utility.js
```diff
@@ -77,23 +78,52 @@ const getPropertyValue = (obj, k) => {
 const handleHash = () => {
     const hash = window.location.hash;
     if (hash) {
-        $(`a[href="${hash}"]`).click();
+        document.querySelector(`a[href="${hash}"]`).click();
     }
 };
 
 const clearable = (element) => {
     element.addClass('clear');
```
kellybinary: here's another one :)

kellybinary: here's another one :)

negarn: afaik, this is from before and the element being passed to this function is a jquery element, so should be fine for now

kellybinary: ok cool, how about line 86: `element.addClass('clear')` ?

negarn: yes, that's what I thought we are discussing

negarn: this is confusing because when this function was created element should have been named $element but it wasn't. it's part of the code from before and unchanged by this PR. it will be changed later when we get to files that are calling this function and we stop sending jquery elements to it.

kellybinary: oh ya, i thought you were talking about line 80 there 😅

kellybinary: otherwise, pr looks good 👍

PR #2851, opened 2017-09-21
kelly/hackathon-agenda

kellybinary:


src/templates/landing_pages/hackathon.html.tt
```diff
@@ -126,11 +126,42 @@
...
             <div class="white-bg">
                 <div class="container">
                     <h2>Agenda</h2>
-                    <p>To be confirmed</p>
+                    <h3 class="text-bold">Saturday, 30th September 2017</h3>
+                    <div class="gr-row">
+                        <div class="gr-10 gr-push-1 gr-12-m gr-push-0-m">
+                            <div class="agenda">
+                                <div class="agenda__vertical-line"></div>
+                                [% BLOCK agenda_list %]
```
binary-hamedanchi: To define a `BLOCK` and including it once seems unnecessary here. Just keeping the `FOREACH` loop would be fine.

PR #2902, opened 2017-10-06
negar/audit_style

negarn:
fix centered text to keep background at full height


ashkanx:
@negar-binary conflict

PR #2956, opened 2017-10-19
negar/audit_from_proposal

negarn:
populate audit table from proposal


src/javascript/binary/websocket_pages/user/view_popup/view_popup.js
```diff
@@ -303,11 +302,10 @@ const ViewPopup = (() => {
         tr.appendChild(createElement('th', { class: 'gr-8 gr-6-t gr-6-p gr-6-m', text: localize('Audit Page') }));
         tr.appendChild(createElement('th', { class: 'gr-2 gr-3-t gr-3-p gr-3-m' }));
         table.appendChild(tr);
-        populateAuditTable(show);
```
negarn: this was only working previously because we were waiting for tick history response. now it won't work as the div it's trying to append to doesn't exist on the page until `div.appendChild(table);` and `div.insertAfter(document.getElementById('sell_details_chart_wrapper'));` are done.

PR #2978, opened 2017-10-24
kelly/ico-news

kellybinary:
Add ICO news coverage section


src/javascript/lib/landing_pages/ico.js
```diff
@@ -37,7 +37,6 @@ window.onload = function() {
 
         if (!hash) {
             switchView('home');
-            scrollTo(0);
```
negarn: why did we remove this?

src/sass/pages/ico_landing_page.scss
```diff
@@ -535,10 +535,160 @@ input[type='text']:focus, input[type='submit']:focus {
     }
 }
 
-#ico_subscribe_section, #lykke-section {
+#ico_subscribe_section, #lykke-section, #media-coverage {
     padding: 50px 15px;
 
-    .margin-top-40 {
-        margin-top: 40px;
+    &#ico_subscribe_section {
```
negarn: this is the parent id (repeated twice, `#ico_subscribe_section, [...] &#ico_subscribe_section` are we repeating the same id? if yes we should change it, if no we should fix this hirearchy

src/javascript/lib/landing_pages/ico.js
kellybinary: to avoid scrolling to the top of the page everytime a user refreshes the page

src/sass/pages/ico_landing_page.scss
```diff
@@ -535,10 +535,160 @@ input[type='text']:focus, input[type='submit']:focus {
...
+}
+
+#media-coverage {
+    .slider-container {
+        position: relative;
+    }
+    .media {
+        background: $white;
+        border: 1px solid $COLOR_DARK_WHITE;
+        border-radius: 3px;
```
negarn: we can use the variable `$BORDER_RADIUS` for the value

src/javascript/lib/landing_pages/ico.js
negarn: was it requested? because we're doing this on our site, should we remove it there too?

kellybinary: no, it wasn't requested and i don't think we're doing this on binary-static since the js and css for landing pages are sort of separated from binary-static

negarn: so isn't it better to keep it and maintain consistency?

PR #2983, opened 2017-10-25
zaki/add_welcome_page

zakibinary:
Added welcome page after user completes virtual account signup.


src/javascript/binary/base/client.js
```diff
@@ -324,7 +324,7 @@ const Client = (() => {
     const getUpgradeInfo = (landing_company, jp_account_status = State.getResponse('get_settings.jp_account_status.status')) => {
         let type         = 'real';
         let can_upgrade  = false;
-        let upgrade_link = 'realws';
+        let upgrade_link = 'welcome';
```
binary-hamedanchi: This function is being called from many places to get the upgrade link, etc. (i.e. the upgrade link in header needs this).
I'd suggest to change it [here](https://github.com/binary-com/binary-static/blob/master/src/javascript/binary/websocket_pages/user/new_account/virtual_acc_opening.js#L115) instead (else part).

src/templates/new_account/welcome_page.html.tt
```diff
@@ -0,0 +1,48 @@
...
+                            <img src="[% request.url_for('images/pages/set_currency/aud_icon.svg') %]">
+                            <img src="[% request.url_for('images/pages/set_currency/eur_icon.svg') %]">
+                            <img src="[% request.url_for('images/pages/set_currency/gbp_icon.svg') %]">
+                            <img src="[% request.url_for('images/pages/set_currency/btc.svg') %]">
+                            <img src="[% request.url_for('images/pages/set_currency/ltc.svg') %]">
+                            <img src="[% request.url_for('images/pages/set_currency/bch.svg') %]"></p>
+                        <p class="font-s">[% l('Upgrade to a real account and start trading using a wide range of crypto and fiat currencies.') %]</p>
+                    </div>
+                    <div class="box-item-end">
+                        <p><a class="button" href="[% request.url_for('realws.html') %]"><span>[% l('Upgrade now') %]</span></a></p>
```
binary-hamedanchi: The URL differs according to the account type which client can upgrade to.
Need to set the URL via js, like other places.
Thanks.

src/javascript/binary/base/client.js
zakibinary: Alright, previously it was changed in `virtual_account_opening.js`. I'll just revert back. Thanks

src/templates/new_account/welcome_page.html.tt
zakibinary: Ok, noted. So is it alright if I create a new module for `WelcomePage` in `static_pages.js` to assign the `href` to the buttons in js?

binary-hamedanchi: Sure, as long as it's not getting so long.

negarn:
conflict

src/javascript/binary/websocket_pages/user/new_account/virtual_acc_opening.js
```diff
@@ -106,13 +106,13 @@ const VirtualAccOpening = (() => {
             const residence   = response.echo_req.residence;
             Client.set('residence', residence, new_account.client_id);
             LocalStore.remove('gclid');
-            BinarySocket.send({ landing_company: residence }).then((response_lc) => {
+            BinarySocket.send({ landing_company: residence }).then(() => {
```
negarn: if we don't need the response from landing company anymore can we just remove this websocket send and wait?

src/javascript/binary/websocket_pages/user/new_account/welcome_page.js
```diff
@@ -0,0 +1,38 @@
+const Client              = require('../../../base/client');
+const localize            = require('../../../base/localize').localize;
+const createElement       = require('../../../base/utility').createElement;
+const Url                 = require('../../../base/url');
+const State               = require('../../../base/storage').State;
+const BinarySocket        = require('../../../websocket_pages/socket');
```
negarn: alignment is off, the `=` are too far from the longest variable name

src/javascript/binary/websocket_pages/user/new_account/welcome_page.js
```diff
@@ -0,0 +1,38 @@
...
+            const upgrade_btn       = document.getElementById('upgrade_btn');
+            const upgrade_info      = Client.getUpgradeInfo(landing_company, jp_account_status);
+            const show_welcome_msg  = upgrade_info.can_upgrade;
+
+            const setButtonLink = (url, msg) => {
+                upgrade_btn.html(createElement('span', { text: localize(msg) })).setAttribute('href', Url.urlFor(url));
+            };
+
+            const welcome_msg = document.getElementById('welcome_container');
+            welcome_msg.setVisibility(1);
```
negarn: can you check if `welcome_msg` then only set visibility to 1?

src/javascript/binary/websocket_pages/user/new_account/welcome_page.js
```diff
@@ -0,0 +1,38 @@
...
+
+        BinarySocket.wait('authorize', 'landing_company', 'get_settings').then(() => {
+            const landing_company   = State.getResponse('landing_company');
+            const jp_account_status = State.getResponse('get_settings.jp_account_status.status');
+            const upgrade_btn       = document.getElementById('upgrade_btn');
+            const upgrade_info      = Client.getUpgradeInfo(landing_company, jp_account_status);
+            const show_welcome_msg  = upgrade_info.can_upgrade;
+
+            const setButtonLink = (url, msg) => {
+                upgrade_btn.html(createElement('span', { text: localize(msg) })).setAttribute('href', Url.urlFor(url));
```
negarn: can you check if `upgrade_btn` then set the html?

src/javascript/binary/websocket_pages/user/new_account/welcome_page.js
```diff
@@ -16,13 +16,17 @@ const WelcomePage = (() => {
...
-            welcome_msg.setVisibility(1);
+            if(welcome_msg) {
+                welcome_msg.setVisibility(1);
+            }
 
-            if (show_welcome_msg) {
+            if (show_welcome_msg && upgrade_btn) {
                 setButtonLink(upgrade_info.upgrade_link, 'Upgrade now');
             } else {
                 upgrade_btn.classList.add('button-disabled');
```
negarn: here you are saying if `upgrade_btn` doesn't exit, add `button-disabled` class to it, which won't work because the `btn_upgrade` doesn't exist :)

zakibinary: Ahh, right, should've put more thought into it. XO

PR #2993, opened 2017-10-27
Ico subscribe

4p00rv:


src/javascript/binary/base/binary_pjax.js
```diff
@@ -133,7 +133,7 @@ const BinaryPjax = (() => {
     const handlePopstate = (e) => {
         const url = e.state ? e.state.url : window.location.href;
         if (url) {
-            processUrl(url, true);
+            handleClick(url, true);
```
binary-hamedanchi: Does it break the browser's back button? (`handleClick()` expects the first parameter to be the event, otherwise does nothing)

4p00rv: @binary-hamedanchi It actually breaks the back button functionality, let me see if I can make it work.

PR #2998, opened 2017-10-27
zaki/update_payment_methods_page

zakibinary:
Added BCH & ETH to payment methods page.


negarn:
@lokemunsei this is a public repo, we can't call a private repo's information.

PR #3014, opened 2017-11-01
kelly/ico-language-dropdown

kellybinary:


src/javascript/lib/landing_pages/common.js
```diff
@@ -1,15 +1,17 @@
 // Check view width, add navbar height as offset if on desktop
 function checkWidth() {
-    let mq = window.matchMedia("(max-width: 1199px)");
+    var mq = window.matchMedia("(max-width: 1199px)");
```
binary-hamedanchi: These js files will be transpiled [here](https://github.com/binary-com/binary-static/pull/3022).

kellybinary: noted, thanks!

src/javascript/lib/landing_pages/common.js
```diff
@@ -7,9 +7,11 @@ function checkWidth() {
 function toggleMobileMenu() {
     const toggleButton = document.getElementById('toggle-menu');
     const navbar       = document.getElementById('navigation');
+    const navbar_item  = document.getElementsByClassName('navbar-collapse')[0];
     toggleButton.addEventListener('click', function (e) {
-        navbar.classList[navbar.classList.contains('expand') ? 'remove' : 'add']('expand');
         e.stopPropagation();
+        navbar.classList[navbar.classList.contains('expand') ? 'remove' : 'add']('expand');
```
aminroosta: @kellybinary  Since we are supprting `IE9+`, we can also use `navbar.classList.toggle('expand');` here.

src/javascript/lib/landing_pages/common.js
```diff
@@ -48,15 +50,15 @@ Math.easeInOutQuad = function (current_time, start_value, change_in_value, durat
 };
 
 function getParamValue(url, key) {
-    var regex   = new RegExp('[?&]' + key + '(=([^&#]*)|&|#|$)');
-    var results = regex.exec(url);
+    let regex   = new RegExp('[?&]' + key + '(=([^&#]*)|&|#|$)');
```
aminroosta: @kellybinary Just to follow eslint setting in other binary-static pages ... could you please use `const` for constant variables instead of `let`.

kellybinary: sure @aminroosta 👍

src/javascript/lib/landing_pages/common.js
kellybinary: @aminroosta ok, i've changed to use `toggle`

PR #3016, opened 2017-11-01
Count down on ICO subscribe page

4p00rv:


src/javascript/binary/websocket_pages/user/ico_subscribe.js
```diff
@@ -43,6 +43,7 @@ const ICOSubscribe = (() => {
         BinarySocket.wait('website_status').then((response) => {
             if (response.website_status.ico_status === 'closed') {
                 $(form_id).replaceWith($('<p/>', { class: 'notice-msg center-text', text: localize('The ICO is currently unavailable.') }));
+                ICOcountDown();
```
binary-hamedanchi: Does this display the countdown even after the ICO ended? (`ico_status` is `closed` also after it ends)

4p00rv: @binary-hamedanchi No, it doesn't. I'm checking it [here](https://github.com/binary-com/binary-static/pull/3016/files#diff-0086c60d4479798647265119fb303f63R118) if there's some time left or not.

src/javascript/binary/websocket_pages/user/ico_subscribe.js
```diff
@@ -109,6 +110,31 @@ const ICOSubscribe = (() => {
         $total.html(formatMoney(currency, total));
     };
 
+    const ICOcountDown = () => {
+        const timer = window.setInterval(() => {
+            const start_time = 1510704000; 
+            const current_time = Math.round((new Date()).getTime()/1000);
```
binary-hamedanchi: Need to check against GMT time as client time is not reliable. You can use `window.time` instead.

src/javascript/binary/websocket_pages/user/ico_subscribe.js
binary-hamedanchi: I see. Then it needs to make it invisible again, in case the page was still open.

src/javascript/binary/websocket_pages/user/ico_subscribe.js
```diff
@@ -109,6 +110,31 @@ const ICOSubscribe = (() => {
         $total.html(formatMoney(currency, total));
     };
 
+    const ICOcountDown = () => {
+        const timer = window.setInterval(() => {
+            const start_time = 1510704000; 
+            const current_time = Math.round((new Date()).getTime()/1000); 
+            const time_left = start_time - current_time; 
+            if(time_left >= 0) {
+                $('.timer').removeClass('invisible'); // Make the timer visible.
```
binary-hamedanchi: Could you get all elements one time only (outside the interval function), so every second it doesn't need to go through DOM again.

src/sass/pages/static_pages.scss
```diff
@@ -267,6 +267,17 @@ body #not_authenticated_financial ul.checked > li {
...
         }
     }
+    .timer {
+        .time {
+            .colon {
+                color: $COLOR_BLUE;
+            }
+        }
+        .text {
+            font-size: 13px;
```
negarn: can we use `$FONT_SIZE_S` or `$FONT_SIZE_XS` instead of hardcoding this?

src/sass/pages/static_pages.scss
```diff
@@ -267,6 +267,17 @@ body #not_authenticated_financial ul.checked > li {
             }
         }
     }
+    .timer {
+        .time {
+            .colon {
+                color: $COLOR_BLUE;
```
negarn: can you try giving them a `primary-color` class instead of adding extra styling?

src/javascript/binary/websocket_pages/user/ico_subscribe.js
```diff
@@ -109,6 +110,31 @@ const ICOSubscribe = (() => {
...
+                const s = (`0${  time_left % 60}`).slice(-2); 
+                const m = (`0${  Math.floor(time_left/ 60) % 60}`).slice(-2); 
+                const h = (`0${  Math.floor(time_left / 3600) % 24}`).slice(-2); 
+                const d = (`0${  Math.floor(time_left / (3600 * 24))}`).slice(-2); 
+                $('.timer .time .days').text(d);
```
negarn: we can just assign `const $timer =  $('.timer'); const $time = $timer.find('.time'); const $days = $time.find('.days'); const $hours = $time.find('.hours');` and so on outside of interval function and call the variables here.

src/javascript/binary/websocket_pages/user/ico_subscribe.js
```diff
@@ -109,6 +110,36 @@ const ICOSubscribe = (() => {
...
+        const days = timer.find('.time .days');
+        const hours = timer.find('.time .hours');
+        const minutes = timer.find('.time .minutes');
+        const seconds = timer.find('.time .seconds');
+        const timerID = window.setInterval(() => {
+            const start_time = 1510704000; 
+            const current_time = window.time.unix(); 
+            const time_left = start_time - current_time; 
+            if(time_left >= 0) {
+                timer.removeClass('invisible'); // Make the timer visible.
```
negarn: can we make it visible only after all the text has been added? also you can use `.setVisibility(1)` instead of `.removeClass('invisible')`

4p00rv: Sure @negar-binary

src/javascript/binary/websocket_pages/user/ico_subscribe.js
4p00rv: I think my earlier comment got removed, adding it here.

I'm handling the case where user is still on the page and timer runs down by doing a page reload when the time_left is 0 seconds.

PR #3017, opened 2017-11-01
zaki/add_ico_banner

zakibinary:
Added ICO banner to the site.


src/javascript/binary/base/page.js
```diff
@@ -101,6 +113,11 @@ const Page = (() => {
         Menu.onUnload();
     };
 
+    const removeIcoBanner = () => {
+        document.getElementById('ico_banner').classList.add('invisible');
+        sessionStorage.setItem('icoBanner', true);
```
negarn: maybe we can rename `icoBanner` to something like `hide_ico_banner`? to make it more clear what it's being set for, right now it seems the opposite of what it is

src/templates/global/ico_banner.html.tt
```diff
@@ -0,0 +1,16 @@
+<div id="ico_banner" class="ico-banner invisible [% class %]">
+    <div class="gr-row gr-centered">
+        <div class="gr-12">
+            <div class="gr-row gr-no-gutter ico-box">
+                <div class="gr-3 gr-padding-10 ico-column left-bg"></div>
+                <div class="gr-9 gr-padding-10 ico-column">
+                    <div class="gr-row gr-no-gutter">
+                        <div class="gr-9 gr-12-p gr-12-m"><p>[% l('Binary.com ICO starts at 15th November 2017') %]</p></div>
```
negarn: instead of hardcoding `Binary.com` can you change it to a placeholder? so it would be `l('[_1] ICO starts at 15th November 2017', website_name)`

src/javascript/binary/base/page.js
zakibinary: Noted, I'll rename the session key.

src/templates/global/ico_banner.html.tt
zakibinary: Will do.

src/templates/global/layout.html.tt
```diff
@@ -21,14 +21,27 @@
 [% END %]
 [% IF layout %]
         <div id="content"[% IF current_route +%] class="[% current_route %]-content"[% END %]>
+            [% IF layout == 'full_width' %]
+                <div class="container">
```
binary-hamedanchi: Having more than one `.container` in `#content` breaks generic error message [here](https://github.com/binary-com/binary-static/blob/master/src/javascript/binary/base/binary_loader.js#L111)
i.e. browse a real only page using a virtual account (limits, cashier password, ...)
You can also handle this by updating the selector in `displayMessage()`.

zakibinary: I see, I'll remove the `.container` then and assign the css values directly to the `ico-banner` class.

binary-hamedanchi: Thanks.

src/templates/global/layout.html.tt
```diff
@@ -21,14 +21,23 @@
 [% END %]
 [% IF layout %]
         <div id="content"[% IF current_route +%] class="[% current_route %]-content"[% END %]>
+            [% IF layout == 'full_width' %]
+                [% PROCESS global/ico_banner.html.tt container="ico-banner-container" class='top-margin' %]
+            [% END %]
+            [% IF layout == 'default' %]
```
binary-hamedanchi: How about `[% ELSIF ...`?

zakibinary: Done. Thanks for pointing it out.

PR #3019, opened 2017-11-01
Document upload improvements

4p00rv:


src/javascript/binary/base/header.js
```diff
@@ -265,27 +265,29 @@ const Header = (() => {
...
+                authenticate         : () => buildMessage('[_1]Authenticate your account[_2] now to take full advantage of all payment methods available.',                                      'user/authenticate'),
+                currency             : () => buildMessage('Please set the [_1]currency[_2] of your account.',                                                                                    'user/set-currency'),
+                document_review      : () => buildMessage('We are reviewing your documents. For more details [_1]contact us[_2].',                                                               'contact'),
+                financial_limit      : () => buildMessage('Please set your [_1]30-day turnover limit[_2] to remove deposit limits.',                                                             'user/security/self_exclusionws'),
+                document_needs_action: () => buildMessage('[_1]Your Proof of Identity or Proof of Address[_2] did not meet our requirements. Please check your email for further instructions.', 'user/authenticate'),
```
negarn: string needs to be added to translations? also should be on line 270 to follow alphabetical order

.eslintrc
```diff
@@ -25,6 +25,7 @@
         'one-var':               ['error', { initialized: 'never', uninitialized: 'always' }],
         'semi':                  ['error', 'always'],
         'camelcase':             0,
+        "no-bitwise":            1,
```
4p00rv: @binary-hamedanchi @negar-binary I hope this is not a problem?

binary-hamedanchi: @4p00rv Since bitwise operators are rarely used, could you use single line disable like `// eslint-disable-line no-bitwise` instead, as this way, I'm concerned about mistyped `&&` or `||` to `&` or `|` in the future.
Thanks.

PR #3041, opened 2017-11-07
Try to fix crowdin in context translation in ico.html page

aminroosta:


aminroosta:
@4p00rv  please review.

PR #3060, opened 2017-11-08
kelly/ico-howto-doc

kellybinary:
Get howto_ico document url based on current language


src/javascript/lib/landing_pages/ico.js
```diff
@@ -431,3 +436,10 @@ function setupCrowdin() {
         }
     }
 }
+
+function getDocumentUrl(lang = 'en') {
+    if (lang === 'en') {
+        return 'https://ico_documents.binary.com/howto_ico.pdf';
+    }
+    return `https://ico_documents.binary.com/howto_ico_${lang}.pdf`;
```
jy-binary: this is wrong - only ru has its separate PDF. Other languages should default to English.

src/javascript/lib/landing_pages/ico.js
```diff
@@ -431,3 +436,10 @@ function setupCrowdin() {
         }
     }
 }
+
+function getDocumentUrl(lang = 'en') {
+    if (lang === 'ru') {
+        return 'https://ico_documents.binary.com/howto_ico_ru.pdf';
+    }
+    return `https://ico_documents.binary.com/howto_ico.pdf`;
```
4p00rv: Can you please change this to regex, like /^(ru)$/i.test(lang) and use the lang var in url? This would make future additions easier to implement.

src/javascript/lib/landing_pages/ico.js
```diff
@@ -431,3 +436,10 @@ function setupCrowdin() {
         }
     }
 }
+
+function getDocumentUrl(lang = 'en') {
+    if (lang === 'ru') {
```
4p00rv: Can you please change this to regex, like `/^(ru)$/i.test(lang)` and use the `lang`var in url? This would make future additions easier to implement.

src/javascript/lib/landing_pages/ico.js
4p00rv: This doesn't need to templated. Please use single quotes here.

PR #3064, opened 2017-11-08
Allow user to declare themselves as professional

4p00rv:


src/javascript/binary/websocket_pages/user/account/settings/professional_client.js
```diff
@@ -66,23 +69,39 @@ const professionalClient = (() => {
...
-                .on('submit', () => {
+                .on('submit', (e) => {
+                    e.preventDefault();
                     if ($chk_professional.is(':checked')) {
-                        // TODO: add the call to send when back-end adds it
-                        BinarySocket.send({}).then((response) => {
+                        BinarySocket.send(populateReq()).then((response) => {
                             if (response.error) {
-                                $('#form_message').text(response.error.message);
+                                $error.text(response.error.message).removeClass('invisible');
```
kellybinary: should we localize and add the error message here to GenerateStaticData too?

PR #3065, opened 2017-11-09
Ico subscribe

aminroosta:


src/javascript/binary/websocket_pages/user/ico_portfolio.js
```diff
@@ -94,11 +95,10 @@ const ICOPortfolio = (() => {
             $('#portfolio-table').setVisibility(0);
         } else {
             $('a[action="cancel"]:not(.button-disabled)').on('click', function () {
-                BinarySocket.send({
-                    sell : $(this).attr('contract_id'),
-                    price: 0,
-                });
+                const contract_id = $(this).attr('contract_id');
```
kellybinary: need to add `e.preventDefault()`?

src/javascript/binary/websocket_pages/user/ico_portfolio.js
```diff
@@ -107,6 +107,35 @@ const ICOPortfolio = (() => {
...
+        const lightbox = $('#cancel_bid_confirmation');
+        document.body.appendChild(lightbox[0]);
+        lightbox.setVisibility(1);
+        lightbox.find('.error-msg').addClass('invisible');
+        lightbox.find('#chk_confirm').prop('checked', false);
+
+        const confirm_form_id = '#frm_confirm';
+        if (!lightbox.find(confirm_form_id).hasClass('validation_initialized')) {
+            Validation.init(confirm_form_id, [
+                { selector: '#chk_confirm', validations: [['req', { hide_asterisk: true }]] },
```
kellybinary: do we need the nested array for validations?

aminroosta: @kellybinary Just checked the `form_validation.js`, Since we are passing an option(`{hide_asterisk: true}`) it should be a nester array!

PR #3073, opened 2017-11-09
update expiray date min value

aminroosta:


src/templates/global/authenticate_message.html.tt
```diff
@@ -17,6 +17,9 @@
                                 <li>[% instruction %]</li>
                             [% END %]
                         </ul>
+                        <p>
+                            <a href="https://marketing.binary.com/authentication/2017_Authentication_Process.pdf" target="_blank">[% l('View sample') %]</a>
```
jy-binary: seems strange that the link is titled 'View sample' - can you check with Joanna/Derek?

jy-binary: Change to "Learn More"

PR #3078, opened 2017-11-09
kelly/fix-limits

kellybinary:


jy-binary:
please check why tests fail

PR #3093, opened 2017-11-10
Add Coinist news

kellybinary:


src/templates/landing_pages/ico.html.tt
```diff
@@ -535,6 +535,7 @@
 
             <!-- news order: latest to oldest -->
             [% news=[
+                { logo="coinist"                timestamp="Nov 10, 2017"    headlines="Binary.com spearheads a new wave of ICOs with its “IPO on the blockchain"  url="https://www.coinist.io/binary-com-and-its-ipo-on-the-blockchain/" },
```
jy-binary: there's a missing double-quote at the end of “IPO on the blockchain

jy-binary: You can also use single quotes around 'IPO on the blockchain'

PR #3094, opened 2017-11-10
zaki/ico_acc_market_fix

zakibinary:
Return 'None' for available markets in accounts page if account is `ico_only`.


src/javascript/binary/websocket_pages/user/accounts.js
```diff
@@ -93,6 +92,9 @@ const Accounts = (() => {
     };
 
     const getAvailableMarkets = (loginid) => {
+        if(Client.get('is_ico_only', loginid)) {
```
kellybinary: add white space? `if (Client.get(....`

src/javascript/binary/websocket_pages/user/accounts.js
```diff
@@ -93,6 +92,9 @@ const Accounts = (() => {
...
+            return [localize('None')];
```
kellybinary: need to return an array?

PR #3099, opened 2017-11-13
negar/set_is_ico_only

negarn:
set is_ico_only flag when logging in


src/javascript/binary/base/logged_in.js
```diff
@@ -86,6 +87,8 @@ const LoggedInHandler = (() => {
 
     const isDisabled = (loginid_list, loginid) => (loginid_list ? +(new RegExp(`${loginid}:[VR]:D`)).test(loginid_list) : 0);
 
+    const isIcoOnly = (loginid_list, loginid) => (loginid_list ? +(new RegExp(`${loginid}:[VR]:[DE]:I`)).test(loginid_list) : 0);
```
4p00rv: @negar Regex should be `${loginid}:[VR]:[DE]:[IN]` `N` is for normal accounts.

PR #3100, opened 2017-11-13
kelly/ico-disclaimer-dropdown

kellybinary:


src/javascript/lib/landing_pages/ico_disclaimer.js
```diff
@@ -74,9 +74,21 @@ window.onload = function() {
...
     });
 
+    var validation_set = false; // To prevent validating before submit
+
     frm_select_residence.addEventListener('submit', function (e) {
         e.preventDefault();
         var val = document.getElementById('residence_list').value;
+        if (!validateResidence(this)) {
+            if (!validation_set) {
+                this.addEventListener('change', function (evt) {
```
4p00rv: @kellybinary Why do we need to set the eventListener here? Won't this work for submit event?

PR #3110, opened 2017-11-13
Add ICO bids link for all accounts except japan

4p00rv:


scripts/lib/BS.pm
```diff
@@ -253,6 +253,14 @@ sub menu {
 
     # push @{$menu}, $self->_main_menu_trading();
 
+    push @menu,
+        {
+        id         => 'topMenuIcoBids',
+        class      => 'ja-hide client_logged_in',
+        url        => url_for('/user/ico-subscribe'),
+        text       => localize('ICO Bids'),
```
negarn: @4p00rv need to add the translation string in `GenerateStaticData.pm` file. also can we move the comment above this menu item instead of below it? :)

src/javascript/binary/base/client.js
```diff
@@ -246,7 +247,8 @@ const Client = (() => {
                 }
 
                 applyToAllElements('.client_logged_in', (el) => {
-                    if (!/ico-only-hide/.test(el.classList) || !is_ico_only) {
+                    if (!/ico-only-hide/.test(el.classList) || !is_ico_only ||
```
negarn: the last condition should be &&, similar to line 262

jy-binary:
@4p00rv strange, I merged https://trello.com/c/j4iIJLpr/2493-negar-hideprofittableshopico and expected this PR would merge those changes cleanly - pls check conflicts

4p00rv:
@jy-binary  I don't see any conflicts 🤔

PR #3111, opened 2017-11-13
zaki/ico_add_participate_now

zakibinary:
Added learn more button in ico-subscribe page. Altered layout for ICO page once countdown ends and added mobile bottom banner for when  ico is live.


src/javascript/lib/landing_pages/ico.js
```diff
@@ -311,6 +311,10 @@ function initCountdown(start_epoch) {
                 clearInterval(countdownd_interval);
             }
         }
+
+        if (is_started) {
+            document.getElementById('ico-bottom-banner').classList.remove(hidden_class);
```
negarn: can we check that element exists before manipulating it?

PR #3119, opened 2017-11-14
kelly/ico-fixed-header

kellybinary:


src/javascript/lib/landing_pages/common.js
```diff
@@ -15,6 +15,11 @@ function toggleMobileMenu() {
     });
 }
 
+function collapseMenu() {
+    document.getElementById('navigation').classList.remove('expand');
+    document.getElementsByClassName('navbar-collapse')[0].classList.remove('expand');
```
negarn: can we check both elements exist first? :)

PR #3120, opened 2017-11-14
Use ico_status Api

aminroosta:


src/javascript/binary/websocket_pages/user/ico_info.js
```diff
@@ -189,8 +186,8 @@ const ICOInfo = (() => {
 
         getHighstock((Highstock) => {
             Highcharts = Highstock;
-            BinarySocket.send({website_status: 1}, {forced: true}).then((response) => {
-                init(response.website_status);
+            BinarySocket.send({ico_status: 1}, {forced: true}).then((response) => {
+                init(response.ico_status);
```
negarn: do we want to check if response.error?

src/javascript/binary/websocket_pages/user/ico_info.js
```diff
@@ -189,8 +186,8 @@ const ICOInfo = (() => {
 
         getHighstock((Highstock) => {
             Highcharts = Highstock;
-            BinarySocket.send({website_status: 1}, {forced: true}).then((response) => {
-                init(response.website_status);
+            BinarySocket.send({ico_status: 1}, {forced: true}).then((response) => {
```
negarn: can remove forced object

PR #3122, opened 2017-11-14
zaki/ico_fix_target_links

zakibinary:
Append attribute `rel="noopener noreferrer"` to all `target="_blank"` links.
Use `window.opener = null` before appending `window.location`.


src/javascript/binary/base/page.js
```diff
@@ -121,7 +123,9 @@ const Page = (() => {
 
     const clickIcoBannerButton = (e) => {
         e.stopPropagation();
-        window.open(Url.urlFor('ico'), '_blank');
+        const openLink = window.open();
```
negarn: variable name should be open_link

src/javascript/lib/landing_pages/ico.js
```diff
@@ -452,6 +452,12 @@ function setupCrowdin() {
     }
 }
 
+function openSubscribeLink(link) {
+    var openLink = window.open();
```
negarn: same here

PR #3147, opened 2017-11-15
zaki/ico_hide_signup_for_existing_users

zakibinary:
Hide email sign up form if active user session is found on binary-static site.


src/javascript/lib/landing_pages/ico.js
```diff
@@ -465,6 +466,19 @@ function setupCrowdin() {
     }
 }
 
+function checkUserSession() {
+    var signup_form        = document.getElementById('sign-up-section');
+    var account_exists_msg = document.getElementById('account_exists_message');
+    if (!localStorage.getItem('active_loginid')) {
```
negarn: in client.js we have this function:
```
 const isLoggedIn = () => (
        !isEmptyObject(getAllAccountsObject()) &&
        get('loginid') &&
        get('token')
    );
```
can you add a similar check so we are really sure that client is logged in and keep it consistent?

PR #3157, opened 2017-11-16
Add warning for eu clients on deposit

4p00rv:


src/templates/cashier/index.html.tt
```diff
@@ -54,6 +54,10 @@
         <span class="invisible normal_currency client_logged_out">[% l('Bank-wire, credit card, e-cash wallet') %]</span>
         <span class="invisible crypto_currency">[% l('Cryptocurrency') %]</span>
     </h3>
+    <div class="notice-msg invisible eu-only gr-padding-20">
+        [% l('If you’d like to participate in the Binary.com ICO, please fund your ICO account instead of your trading account.') %]
```
ashkanx: please use `website_name` instead of Binary.com

src/templates/cashier/index.html.tt
```diff
@@ -54,6 +54,10 @@
...
+    </div>
+    <div class="gr-padding-10 invisible eu-only"></div>
```
ashkanx: what is it for? just adding more space?

4p00rv: @ashkanx yes.

PR #3170, opened 2017-11-17
ICO message changes for non-ico CR clients

4p00rv:


4p00rv:
Please don't mind the whitespace changes. To view only the intended changes uses this url:
https://github.com/binary-com/binary-static/pull/3170/files?w=1

src/templates/user/ico_subscribe.html.tt
```diff
@@ -176,6 +176,10 @@
     <p>[% l("Please select your ICO account to proceed.") %]</p>
 </div>
 
+<div class="invisible" id="ico_account_message_real">
+    <p>[% l("Please select your real account to proceed.") %]</p>
```
negarn: can we change the wording to real money account

4p00rv: Sure

PR #3176, opened 2017-11-17
Add 1px to all bars

aminroosta:


src/javascript/binary/websocket_pages/user/ico_info.js
```diff
@@ -162,6 +162,11 @@ const ICOInfo = (() => {
                 values         : allValues,
                 finalPriceLabel: `${localize('Final Price')} ($${final_price})`,
                 callback       : () => {
+                    const $bars = $root.find('.barChart svg .highcharts-column-series > rect');
+                    $bars.each((inx, bar) => {
+                        const y = +$(bar).attr('y');
+                        $(bar).attr('y', `${y-1}`);
```
negarn: would it be worth changing to `+y-1`?

PR #3185, opened 2017-11-20
zaki/translate_btn_fix

zakibinary:
Updated styling for buttons on `authenticate.html`.


src/templates/global/authenticate_message.html.tt
```diff
@@ -47,19 +47,19 @@
                                         <div class="gr-row form-row center-text-m">
                                             <div class="gr-6 gr-12-m">
                                                 <input id="front_file[% loop.count %]" class="file-picker" type="file" accept=".jpg, .jpeg, .gif, .png, .pdf" data-type="[% document.value %]">
-                                                <label for="front_file[% loop.count %]" class="button">[% l('Front Side') %] <span class="add" /></button>
```
negarn: is there a `.add` styling that should be removed? or it's being used in other places?

zakibinary: there's `.add` and `.remove`, it's only being used in `authenticate.html` via `authenticate_message.html.tt`. 

I'll reconfirm again with design team again whether we should remove this as it indicates the state of the buttons for adding and deleting attached documents.

PR #3203, opened 2017-11-22
zaki/ico_analysis_update

zakibinary:
Added Nishant San's report to reports section.


src/javascript/lib/landing_pages/ico.js
```diff
@@ -115,6 +115,13 @@ window.onload = function() {
         open_link.location = getLykkeReport(language.toLowerCase());
     });
 
+    document.getElementById('nishant-btn').addEventListener('click', function(e) {
+        e.preventDefault();
+        const open_link    = window.open();
+        open_link.opener   = null;
+        open_link.location = getNishantReport();
```
negarn: if we are repeating this functionality everywhere can we add it as a function that expects a location?

zakibinary: Done.

PR #3204, opened 2017-11-22
kelly/remove-ios

kellybinary:


src/templates/static/platforms.html.tt
```diff
@@ -14,11 +14,8 @@
             <a class="button" href="[% url %]"[% IF target %] target="[% target %]"[% END %][% IF url.search('^http') %] rel="noopener noreferrer"[% END %]><span>[% button_text %]</span></a>
         </p>
         [% END %]
-        [% IF app_store_url and google_play_url %]
+        [% IF google_play_url %]
             <div class="gr-row">
-                <a class="gr-5 gr-6-m" href="[% app_store_url %]" target="_blank" rel="noopener noreferrer">
-                    <div class="app-store-badge"></div>
```
negarn: can we remove the styling for this class too? also if you search for `app-store-badge` in the code base you will see some other code in `download.html.tt` as well

negarn: and we have all the svg files `appstore_`lang`.svg`

kellybinary: we're still keeping iOS link for MT5

PR #3210, opened 2017-11-23
ICO mobile view fixes

4p00rv:


src/templates/user/ico_subscribe.html.tt
```diff
@@ -73,8 +73,16 @@
     [% INCLUDE global/separator_line.html.tt show_mobile=1 class='gr-padding-20' invisible=1 %]
     <div id="ico_bids">
         <div class="gr-row">
-            <h1 class="gr-7">[% l('[_1] ICO Auction Bids', website_name) %]</h1>
-            <div class="gr-5 align-end invisible" id="view_ico_info">
+            <div class="gr-12-m gr-hide gr-show-m gr-padding-10 view_ico_info">
```
negarn: I think instead of duplicating the code for mobile you can change the position priority (first versus second) using flexbox styling

4p00rv:
@negarn Thanks for suggesting. Now the changes are pretty clean 😅

PR #3216, opened 2017-11-24
zaki/tooltip_tax_residence

zakibinary:
Added tooltip for tax id and tax residence.


negarn:
do we want to add this to personal details page too?

PR #3238, opened 2017-11-30
negar/fix_match

negarn:
use existing parent element for comparison instead of putting inner html in a div


binary-hamedanchi:
Does the polyfill need to be imported in `utility.js`?
```js
require('../_common/lib/polyfills/element.matches');
```

PR #3247, opened 2017-12-07
Lookback

4p00rv:


src/images/pages/trade/High_ICON.svg
```diff
@@ -0,0 +1 @@
+<svg width="48" height="48" viewBox="0 0 48 48" xmlns="http://www.w3.org/2000/svg"><title>High ICON</title><g fill="none" fill-rule="evenodd"><path fill="#2AAAE2" d="M0 48h48V0H0z"/><path fill="#FFF" d="M0 33h48v-2H0z"/><path stroke="#FFF" stroke-width="4" d="M5 10l11.477 18L34 10"/><path fill="#FFF" d="M36 8h-8l-3.5 3.5 5.126.094 2.824 2.854.05 5.052L36 16z"/><path d="M41 29V7m-3 19l3 3 3-3m-6-16l3-3 3 3" stroke="#FFF" stroke-width="2"/></g></svg>
```
negarn: can we change all the image file names? without any caps and standardize `-` or `_`, at least between these 3 image files

4p00rv: sure.

src/javascript/app/pages/trade/charts/highchart.js
```diff
@@ -360,13 +367,36 @@ const Highchart = (() => {
             const low_barrier  = contract.low_barrier;
             if (barrier) {
                 addPlotLine({ id: 'barrier',      value: barrier * 1,      label: localize('Barrier ([_1])', [addComma(barrier)]),           dashStyle: 'Dot' }, 'y');
+                prev_barriers[0] = barrier; // Store values for future references.
             } else if (high_barrier && low_barrier) {
                 addPlotLine({ id: 'high_barrier', value: high_barrier * 1, label: localize('High Barrier ([_1])', [addComma(high_barrier)]), dashStyle: 'Dot' }, 'y');
                 addPlotLine({ id: 'low_barrier',  value: low_barrier * 1,  label: localize('Low Barrier ([_1])', [addComma(low_barrier)]),   dashStyle: 'Dot' }, 'y');
+                prev_barriers[0] = high_barrier; // Store values for future references.
+                prev_barriers[1] = low_barrier; // Store values for future references.
```
negarn: can we change 0 to low and 1 to high? it's easier to remember since 1 is higher than 0 :P

src/javascript/app/pages/trade/contract.js
```diff
@@ -157,6 +157,10 @@ const Contract = (() => {
                             trade_contract_forms.evenodd   = localize('Even/Odd');
                             trade_contract_forms.overunder = localize('Over/Under');
                         }
+                    } else if (contract_category === 'lookback') {
+                        trade_contract_forms.lookbackhigh    = localize('High');
+                        trade_contract_forms.lookbacklow     = localize('Low');
+                        trade_contract_forms.lookbackhighlow = localize('High-Low');
```
negarn: do we need to add to `generate-static-data.js`?

src/javascript/app/pages/trade/charts/highchart.js
4p00rv: That was an impulsive decision 😅 Will change it.

src/javascript/app/pages/trade/lookback.js
```diff
@@ -0,0 +1,35 @@
+const Contract         = require('./contract');
+const Defaults         = require('./defaults');
```
negarn: alignment for `=` :)

src/javascript/app/pages/trade/lookback.js
```diff
@@ -0,0 +1,35 @@
...
+    };
+
+    return {
+        display: displayLookback,
+    };
+})();
+
+module.exports = {
+    Lookback,
+};
```
negarn: since it's just one it can be `module.exports = Lookback;` instead of an object

src/javascript/app/pages/trade/price.js
```diff
@@ -118,6 +119,10 @@ const Price = (() => {
             proposal.barrier = parseInt(prediction.value);
         }
 
+        if (lots && isVisible(lots) && lots.value) {
+            proposal.amount = lots.value;
```
negarn: can we combine this on line 56 were proposal.amount is being decided?

src/javascript/app/pages/trade/price.js
```diff
@@ -151,6 +156,21 @@ const Price = (() => {
             return;
         }
 
+        // hide all containers except current one
+        if (position !== 'middle') {
```
negarn: can we switch this around so it becomes `if (position === 'middle') {...} else {...`

src/javascript/app/pages/trade/price.js
```diff
@@ -151,6 +156,21 @@ const Price = (() => {
...
+            if ($(ele).is(':visible')) {
+                $(ele).fadeOut(200);
+            }
+        } else {
+            const ele_top = document.getElementById('price_container_top');
+            const ele_bottom = document.getElementById('price_container_bottom');
+            if ($(ele_top).is(':visible') || $(ele_bottom).is(':visible')) {
+                $(ele_top).fadeOut(200);
+                $(ele_bottom).fadeOut(200);
+            }
```
negarn: if all of these elements are going to be jquery why get them from javascript at all? can just be `const $ele_top = $('#price_container_top');` etc

src/javascript/app/pages/trade/price.js
```diff
@@ -213,11 +233,16 @@ const Price = (() => {
             }
             comment.show();
             error.hide();
-            commonTrading.displayCommentPrice(comment, (currency.value || currency.getAttribute('value')), proposal.ask_price, proposal.payout);
+            if (!/^(LBFLOATCALL|LBFLOATPUT|LBHIGHLOW)$/.test(type)) {
```
negarn: can we also switch here so it's `if (/^(LBFLOATCALL|LBFLOATPUT|LBHIGHLOW)$/.test(type)) {...} else {...`

src/javascript/app/pages/trade/process.js
```diff
@@ -5,6 +5,7 @@ const Contract         = require('./contract');
 const Defaults         = require('./defaults');
 const Durations        = require('./duration');
 const GetTicks         = require('./get_ticks');
+const lookback         = require('./lookback').Lookback;
```
negarn: since this is a module, it should be capital so `const Lookback         = require('./lookback');`, without the `.Lookback` at the end because it should be a variable not an object if it's a single export

src/javascript/app/pages/trade/price.js
4p00rv: I don't think those two blocks can be combined. I will just move it below the block on :56.

src/javascript/app/pages/trade/price.js
```diff
@@ -157,18 +157,13 @@ const Price = (() => {
...
-                $(ele).fadeOut(200);
-            }
-        } else {
-            const ele_top = document.getElementById('price_container_top');
-            const ele_bottom = document.getElementById('price_container_bottom');
-            if ($(ele_top).is(':visible') || $(ele_bottom).is(':visible')) {
-                $(ele_top).fadeOut(200);
-                $(ele_bottom).fadeOut(200);
+        if (position === 'middle') {
+            if ($('price_container_top').is(':visible') || $('price_container_bottom').is(':visible')) {
```
negarn: missing selector '#'

src/images/pages/trade/High_ICON.svg
negarn: I think you changed the reference but not the file name for high_icon and low_icon :)

4p00rv: That's git cache :/

negarn: https://stackoverflow.com/a/17688308/4425240

negarn:
conflict

src/templates/app/trade/explanation.jsx
```diff
@@ -169,6 +194,36 @@ const Explanation = () => (
                 <p>{it.L('The <strong>end time</strong> is the selected number of minutes/hours after the <strong>start time</strong> (if less than one day in duration), or at the end of the trading day (if one day or more in duration).')}</p>
                 <p className='hint'><strong>{it.L('Note')}: </strong>{it.L('Touch/No Touch contracts will be refunded at the purchase price if there are less than 2 ticks between the start and end times.')}</p>
             </div>
+            <div id='explain_lookbacklow' className='invisible'>
+                <h3>{it.L('High, Low and Close')}</h3>
+                <p>{it.L('The [_1]high[_2] is the highest point ever reached by the market during the contract period.', '<strong>','</strong>')}</p>
+                <p>{it.L('The [_1]low[_2] is the lowest point ever reached by the market during the contract period.', '<strong>','</strong>')}</p>
```
ashkanx: @4p00rv Using placeholder is not consistent with how we have handled `<strong>` so far. Translators got confused as well. is there any reason or advantage we use a placeholder here?

PR #3253, opened 2017-12-08
Replace Perl with JS.

aminroosta:


scripts/compile.js
```diff
@@ -0,0 +1,331 @@
...
+    const config = {
+        languages : common.languages,
+        sections  : ['app', 'static'],
+        root_path : common.root_path,
+        add_translations: false,
+    }
+    if (program.branch === 'translations') {
+        config.languages = ['ACH'];
+    }
+    config.dist_path = Path.join(config.root_path, 'dist');
```
binary-hamedanchi: `branch` might be missing here.

scripts/compile.js
```diff
@@ -0,0 +1,331 @@
...
+        throw new TypeError(`Invalid url ${url}`);
+    }
+}
+
+const createBuilder = async () => {
+    const config = getConfig();
+
+    let static_hash = Math.random().toString(36).substring(2, 10);
+    try {
+        static_hash = await common.readFile(Path.join(config.dist_path, 'version'));
```
binary-hamedanchi: Does the perl code uses the current hash only if `--path` is available?

scripts/compile.js
```diff
@@ -0,0 +1,331 @@
...
+}
+
+function create_directories() {
+    const config = getConfig();
+    const mkdir = path => fs.existsSync(path) || fs.mkdirSync(path);
+
+    console.log(`Target: ${config.dist_path}`.yellow);
+    mkdir(Path.join(config.dist_path));
+    config.languages.forEach(lang => {
+        mkdir(Path.join(config.dist_path, lang));
```
binary-hamedanchi: `lang` needs to be converted to lower case or better to convert the whole `path` in `mkdir()` above.

scripts/compile.js
aminroosta: When `branch` is missing then `config.languages` will remain as `common.languages` which has the list of all languages.

scripts/compile.js
aminroosta: Perl uses current hash only if `--path` is given.

Fixed.

scripts/compile.js
aminroosta: Fixed.

scripts/compile.js
binary-hamedanchi: If `branch` is given, does it make the directories as `dist/[branch]/[lang]/...`?

scripts/compile.js
```diff
@@ -0,0 +1,334 @@
...
+    });
+    await Promise.all(tasks);
+    console.timeEnd(page.save_as);
+}
+
+createDirectories();
+(async () => {
+    const config = getConfig();
+    try {
+        await compile(common.pages.find(p => p.save_as === (program.path || 'home')));
```
binary-hamedanchi: Would be more convenient to allow using regular expressions like the current one:
```js
await compile(common.pages.filter(p => new RegExp(program.path, 'i').test(p.save_as)));
```

src/templates/_common/components/elements.jsx
```diff
@@ -0,0 +1,187 @@
...
+    className,
+    id,
+    href,
+    param = '',
+    target,
+    text,
+    header,
+    p,
+}) => {
+    const content = p ? <p> {text} </p> : text;
```
negarn: we should remove the white space between `<p>` and `{text}` (?)

src/templates/_common/components/elements.jsx
```diff
@@ -0,0 +1,187 @@
...
+};
+
+export const Select = ({
+    id,
+    className,
+    options,
+}) => (
+    <select id={id} className={className} >
+        {options.map(option => (
+            <option value={option.value || undefined} selected={!!option.selected} >
```
negarn: do we need `|| undefined`?

src/templates/_common/components/elements.jsx
aminroosta: @negar-binary Yea, it's a styling improvement ...however the functionality will remain intact.

src/templates/_common/components/elements.jsx
aminroosta: Yes absolutely :-)
React will omit to render properties with null or undefined values.
We are taking advantages.

negarn: but if it doesn't exist won't it be undefined by default?

src/templates/_common/components/elements.jsx
negarn: are you sure? as we tested with @binary-hamedanchi it keeps the space in the generated template

aminroosta: @negarn Does the browser care about the whitespace?

negarn: I think sometimes, yes, but if we want to rely on it handling most cases then you can leave as is

src/templates/_common/components/elements.jsx
aminroosta: @negar-binary True, you are right. I guess for this it needed to omit the attribute when `option.value === ""`.

PR #3283, opened 2017-12-15
zaki/cashier_pw_align

zakibinary:
Aligns error message according to alignment set for `submit_button` element.

### DIFF
--- a/src/templates/_common/components/forms.html.tt
+++ b/src/templates/_common/components/forms.html.tt
@@ -59,8 +59,7 @@
         [% END -%]
     [% END %]
         [% UNLESS no_error %]
-        <p id="[% msg_id || 'msg_form' %]" class="[% is_centered ? 'error-msg no-margin center-text invisible' : 'error-msg no-margin
-invisible' %]"></p>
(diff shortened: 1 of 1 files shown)


src/templates/_common/components/forms.html.tt
```diff
@@ -59,7 +59,8 @@
         [% END -%]
     [% END %]
         [% UNLESS no_error %]
-        <p id="[% msg_id || 'msg_form' %]" class="error-msg no-margin invisible"></p>
+        <p id="[% msg_id || 'msg_form' %]" class="[% is_centered ? 'error-msg no-margin center-text invisible' : 'error-msg no-margin
+invisible' %]"></p>
```
negarn: 1. why is there a line break?
2. better written as the following to avoid repeating classes:

`<p id="[% msg_id || 'msg_form' %]" class="error-msg no-margin invisible[% is_centered ? ' center-text' : '' %] "></p>`

zakibinary: @negarn , noted with thanks.

PR #3375, opened 2018-01-11
Fixes

kellybinary:


src/javascript/landing_pages/common.js
```diff
@@ -170,3 +174,112 @@ function setupCrowdin() {
...
+            value = key ? current_client[key] : current_client;
+        }
+        if (!Array.isArray(value) && (+value === 1 || +value === 0 || value === 'true' || value === 'false')) {
+            value = JSON.parse(value || false);
+        }
+        return value;
+    }
+
+    function getAllAccountsObject() {
+        return JSON.parse(localStorage.getItem('client.accounts'));
```
zakibinary: Maybe we should check if the item in localStorage exists first before parsing?

kellybinary: it will parse and return `null` if the item doesn't exist

zakibinary: Sorry, what I meant was check if it's an empty object. :p

PR #3378, opened 2018-01-14
zaki/career_page_update

zakibinary:
Updates to career page layout.


src/sass/static/pages.scss
```diff
@@ -470,3 +470,213 @@ body #not_authenticated_financial ul.checked > li {
...
+                    color: $COLOR_WHITE;
+                }
+                &:focus {
+                    border: 0;
+                    outline: 0;
+                }
+            }
+        }
+        .footer-email {
+            background-color: $COLOR_ORANGE;
```
negarn: I think you can use the class `secondary-bg-color` for orange background

src/sass/static/pages.scss
```diff
@@ -470,3 +470,213 @@ body #not_authenticated_financial ul.checked > li {
...
+                    max-width: 80%;
+                    font-weight: 300;
+                }
+            }
+        }
+        .margin-top-100 {
+            margin-top: 100px;
+        }
+        .no-margin-top {
+            margin: 0 0 1em;
```
negarn: from the class name it sounds like the style should be `margin-top: 0;`, can we change the style or update the class name?

src/sass/static/pages.scss
```diff
@@ -470,3 +470,213 @@ body #not_authenticated_financial ul.checked > li {
...
+                        max-width: 80%;
+                    }
+                }
+            }
+            @media (max-width: 768px) {
+                background-position: 75% -30px;
+            }
+        }
+        .center-banner {
+            background: url('../images/pages/careers/careers-2_bg@3x.jpg') no-repeat;
```
negarn: I suggest we change all the image names to standardize `-` or `_` and remove all the `@3x` etc from it

src/sass/static/pages.scss
```diff
@@ -470,3 +470,213 @@ body #not_authenticated_financial ul.checked > li {
...
+.careers-content {
+    .career {
+        margin-top: 0;
+
+        .top-banner {
+            background: url('../images/pages/careers/careers-1_bg@3x.jpg') no-repeat;
+            background-size: cover;
+            min-height: 300px;
+            position: relative;
+            background-position: center center;
```
negarn: is it the same styling if we emit one `center`? or we need to keep both?

zakibinary: Should be the same if we use only one `center`, i'll update it along with the other fixes.

src/templates/static/about/careers.jsx
```diff
@@ -1,132 +1,268 @@
...
-                <div className='gr-row gr-padding-30 center-text'>
-                    <div className='gr-12 gr-padding-30'>
-                        <h1>{it.L('Our locations')}</h1>
+        <div className='container'>
+            <div className='gr-padding-30 center-text'>
+                <h1>{it.L('Interview process')}</h1>
+            </div>
+            <div className='has-tabs'>
+                <ul className='role-section'>
+                    <li className='role'><a className='role-btn' href='#technical'><span>Technical roles</span></a></li>
```
binary-hamedanchi: Missing localisation?

src/templates/static/about/careers.jsx
```diff
@@ -1,132 +1,268 @@
...
+                    <li className='role'><a className='role-btn' href='#non-technical'><span>Non-technical roles</span></a></li>
```
binary-hamedanchi: Same here.

PR #3400, opened 2018-01-18
kelly/remove_next-gen

kellybinary:


src/templates/static/platforms.jsx
```diff
@@ -112,15 +112,14 @@ const Platform = () => (
                                     />
                                     <div className='gr-2'></div>
                                     <Platforms
-                                        image_class='gr-12 gr-7-p gr-10-m'
-                                        image='nextgen'
```
negarn: we can also remove `nextgen.png`?

kellybinary: sure ;)

src/templates/static/platforms.jsx
```diff
@@ -122,8 +123,20 @@ const Platform = () => (
...
+                                        header={it.L('Binary Bot')}
+                                        description={it.L('Auto-trader programming tool')}
+                                        text={it.L('Automate your trading strategies with our simple, "drag-and-drop" bot creation tool.')}
+                                        url='https://bot.binary.com'
+                                        target='_blank'
+                                        button_text={it.L('Try Binary Bot')}
+                                    />
                                 </div>
                                 <div className='gr-row gr-padding-30'>
+                                    {/*
```
binary-hamedanchi: I think this can be deleted as it moved to another row.

PR #3407, opened 2018-01-19
kelly/fix-cashier

kellybinary:


src/javascript/app/japan/cashier.js
```diff
@@ -11,16 +11,33 @@ const CashierJP = (() => {
...
-                $('#japan_cashier_container button').on('click', (e) => {
-                    const result = errorHandler();
-                    if (!result) e.preventDefault();
+        BinarySocket.send({ cashier_password: 1 }).then((response) => {
+            if (!response.error && response.cashier_password === 1) {
+                $container.find('#cashier_locked_message').setVisibility(1);
+            } else {
+                BinarySocket.send({ get_account_status: 1 }).then((response_status) => {
+                    if (!response_status.error && /cashier_locked/.test(response_status.get_account_status.status)) {
+                        $container.find('#cashier_locked_message').html('Your cashier is locked.').setVisibility(1); // Locked from BO
```
negarn: translation?

src/javascript/app/japan/cashier.js
```diff
@@ -11,16 +11,33 @@ const CashierJP = (() => {
...
+                    } else {
+                        const limit = State.getResponse('get_limits.remainder');
+                        if (typeof limit !== 'undefined' && limit < 1) {
+                            $container.find('#cashier_locked_message').html('You have reached the limit.').setVisibility(1);
```
negarn: same here. and do we need to be more specific about what limit they have reached? they will take care of it in translation but it's good for us too

src/javascript/app/pages/cashier/deposit_withdraw.js
```diff
@@ -223,10 +224,25 @@ const DepositWithdraw = (() => {
...
-            if ('error' in response) {
+            if (!response.error && response.cashier_password === 1) {
+                showMessage('cashier_locked_message');
+            } else if ('error' in response) {
                 showError('custom_error', response.error.message);
             } else {
-                init(response.cashier_password);
+                BinarySocket.send({ get_account_status: 1 }).then((response_status) => {
+                    if (!response_status.error && /cashier_locked/.test(response_status.get_account_status.status)) {
+                        showError('custom_error', 'Your cashier is locked.'); // Locked from BO
```
negarn: also need to handle translation here and line 238

src/javascript/app/pages/cashier/deposit_withdraw.js
```diff
@@ -223,10 +224,25 @@ const DepositWithdraw = (() => {
     const onLoad = () => {
         getCashierType();
         BinarySocket.send({ cashier_password: 1 }).then((response) => {
-            if ('error' in response) {
+            if (!response.error && response.cashier_password === 1) {
+                showMessage('cashier_locked_message');
+            } else if ('error' in response) {
```
binary-hamedanchi: By swapping this and the above `if`, it would be more readable. (Also `!response.error` can be removed):
```js
if ('error' in response) {
    showError('custom_error', response.error.message);
} else if (response.cashier_password === 1) {
    showMessage('cashier_locked_message');
} else {
    ...
```

src/javascript/app/japan/cashier.js
kellybinary: sure, for translation use `localize` function and add `{JAPAN ONLY}` tag right?

src/javascript/app/pages/cashier/deposit_withdraw.js
kellybinary: yes it does, thanks :)

src/javascript/app/japan/cashier.js
```diff
@@ -17,11 +17,11 @@ const CashierJP = (() => {
             } else {
                 BinarySocket.send({ get_account_status: 1 }).then((response_status) => {
                     if (!response_status.error && /cashier_locked/.test(response_status.get_account_status.status)) {
-                        $container.find('#cashier_locked_message').html('Your cashier is locked.').setVisibility(1); // Locked from BO
+                        $container.find('#cashier_locked_message').text(localize('{JAPAN ONLY}Your cashier is locked.')).setVisibility(1); // Locked from BO
```
negarn: do we need to add these to the generate static data file?

kellybinary: `Your cashier is locked.` is already in generate static data tho, should we remove `{JAPAN ONLY}` tag here instead

negarn: sure

kellybinary: ok :)

PR #3423, opened 2018-01-26
kelly/enable-oneall

kellybinary:


src/javascript/app/common/form_manager.js
```diff
@@ -115,6 +115,9 @@ const FormManager = (() => {
                 }
                 options.fnc_response_handler(response);
             }
+            if (response.echo_req.residence) {
+                window.location.reload(); // reload page if we are setting residence
```
negarn: I suggest we don't do this in such a common place. instead we can make use of line 116 `fnc_response_handler`, passing this as a function to reload page when it's done, if there is residence in response for the form that needs it.

jy-binary:
got conflicts

PR #3431, opened 2018-01-29
payment agent

4p00rv:


raunakkathuria:
As per code changes I can see that we are getting min max from paymentagent list. Just want to confirm if thats the case

4p00rv:
@raunakkathuria yes that's the case

PR #3461, opened 2018-02-08
remove eurobasket wolves

ruskakimov:


src/templates/static/charity.jsx
```diff
@@ -44,17 +44,6 @@ const Charity = () => (
             </div>
 
             <div className='section-content'>
-                <CharityHelper
-                    country={it.L('Malta')}
-                    charities={[
-                        {
-                            image   : 'eurobasket_wolves',
```
negarn: probably you can also remove the image if it's not used anywhere else

PR #3467, opened 2018-02-09
kelly/appropriateness_test

kellybinary:


src/templates/app/_includes/financial_form.jsx
```diff
@@ -34,26 +34,24 @@ const Values = () => (
...
 
-        <SelectRow con={Experience} id='indices_trading_experience' label={it.L('Indices trading experience')} />
-        <SelectRow con={Frequency}  id='indices_trading_frequency'  label={it.L('Indices trading frequency')} />
-
-        <SelectRow con={Experience} id='commodities_trading_experience' label={it.L('Commodities trading experience')} />
-        <SelectRow con={Frequency}  id='commodities_trading_frequency'  label={it.L('Commodities trading frequency')} />
-
-        <SelectRow con={Experience} id='stocks_trading_experience' label={it.L('Stocks trading experience')} />
-        <SelectRow con={Frequency}  id='stocks_trading_frequency'  label={it.L('Stocks trading frequency')} />
+        <SelectRow con={Experience} id='other_derivatives_trading_experience' label={it.L('Binary options trading experience')} />
```
sd256: I named these as `binary_options_trading_experience` and `binary_options_trading_frequency` in `(get|set)_financial_assessment`

src/templates/app/_includes/financial_form.jsx
```diff
@@ -34,26 +34,24 @@ const Values = () => (
...
+        <SelectRow con={Frequency}  id='other_derivatives_trading_frequency'  label={it.L('Binary options trading frequency')} />
 
-        <SelectRow con={Experience} id='other_derivatives_trading_experience' label={it.L('Binary options or other financial derivatives trading experience')} />
-        <SelectRow con={Frequency}  id='other_derivatives_trading_frequency'  label={it.L('Binary options or other financial derivatives trading frequency')} />
+        <SelectRow con={Experience} id='cfds_trading_experience' label={it.L('CFDs trading experience')} />
```
sd256: `cfd_trading_experience`

src/templates/app/_includes/financial_form.jsx
```diff
@@ -34,26 +34,24 @@ const Values = () => (
...
+        <SelectRow con={Frequency}  id='cfds_trading_frequency'  label={it.L('CFDs trading frequency')} />
```
sd256: `cfd_trading_frequency`

src/javascript/app/pages/user/account/settings/financial_assessment.js
```diff
@@ -45,6 +45,11 @@ const FinancialAssessment = (() => {
             });
         }
 
+        const is_professional = State.getResponse('get_account_status.professional') || State.getResponse('get_account_status.professional_requested');
```
negarn: what if we store `State.getResponse('get_account_status')` in one variable e.g. `const get_account_status` and then do `get_account_status.professional || get_account_status.professional_requested` to avoid calling state twice?

src/javascript/app/pages/user/account/settings/financial_assessment.js
```diff
@@ -45,6 +45,11 @@ const FinancialAssessment = (() => {
...
+        if (is_professional) {
+            $('#trading_experience_form').setVisibility(0);
```
negarn: this can come outside of the if condition and change to `$('#trading_experience_form').setVisibility(! is_professional);`

src/javascript/app/pages/user/account/settings/professional_client.js
```diff
@@ -53,9 +53,13 @@ const professionalClient = (() => {
                 $(popup_selector).find('#btn_accept, #btn_decline').off('click').on('click dblclick', function() {
                     if ($(this).attr('data-value') === 'decline') {
                         $chk_professional.prop('checked', false);
+                        $('#trading_experience_form').setVisibility(1);
                     }
                     $('#professional_popup').remove();
                 });
+                $('#trading_experience_form').setVisibility(0);
+            } else {
+                $('#trading_experience_form').setVisibility(1);
```
negarn: if we're going to call this element many times, better to store it outside, e.g. on line 41 etc, something like `const $trading_experience_form = $('#trading_experience_form');`

src/javascript/app/pages/user/account/settings/financial_assessment.js
kellybinary: ok, sure

src/javascript/app/pages/user/account/settings/financial_assessment.js
```diff
@@ -45,10 +45,8 @@ const FinancialAssessment = (() => {
             });
         }
 
-        const is_professional = State.getResponse('get_account_status.professional') || State.getResponse('get_account_status.professional_requested');
-        if (is_professional) {
-            $('#trading_experience_form').setVisibility(0);
-        }
+        const is_professional = /professional|professional_requested/.test((State.getResponse('get_account_status') || {}).status);
```
negarn: it would have to check exactly for professional term, right now it will catch anything that starts with professional. it needs to be `/^(professional|professional_requested)$/`

negarn: you can do `State.getResponse('get_account_status.status')` instead

src/javascript/app/pages/user/account/settings/professional_client.js
```diff
@@ -53,13 +54,13 @@ const professionalClient = (() => {
                 $(popup_selector).find('#btn_accept, #btn_decline').off('click').on('click dblclick', function() {
                     if ($(this).attr('data-value') === 'decline') {
                         $chk_professional.prop('checked', false);
-                        $('#trading_experience_form').setVisibility(1);
+                        $frm_trading_exp.setVisibility(1);
                     }
-                    $('#professional_popup').remove();
+                    $(popup_selector).remove();
```
negarn: it would be good to also change `const popup_selector    = '#professional_popup';` to `const $popup_selector    = $('#professional_popup');` and use that instead of fetching with jquery every time, especially since there is no other use for the popup_selector id

src/javascript/app/pages/user/account/settings/financial_assessment.js
kellybinary: should we check `professional|professional_requested` from `status` array?

negarn: I think so, yes

src/javascript/app/pages/user/account/settings/professional_client.js
kellybinary: i tried putting it outside, it didn't work :/. it has to be assigned after the element existed, which is after we append the popup to the body

src/javascript/app/pages/user/account/settings/financial_assessment.js
```diff
@@ -44,6 +45,12 @@ const FinancialAssessment = (() => {
             });
         }
 
+        if (LocalStore.get('financial_assessment_redirect')) {
+            const is_mt5_financial = /^(real_vanuatu_standard)$/.test(LocalStore.get('mt5_account'));
```
negarn: couldn't you get this from `Client.get('mt5_account')`?

kellybinary: yes i could, but we need to remove this check if the user navigates away and go back to the page via Settings menu. we need to show Trading Experience form only when the user go the Financial Assessment page from MT5 new account page

negarn: I see! thanks

negarn: but isn't that covered by the `financial_assessment_redirect`? let's talk in person 😅

src/javascript/app/pages/user/account/settings/financial_assessment.js
```diff
@@ -30,20 +30,30 @@ const FinancialAssessment = (() => {
...
     };
 
+    const displayHighRiskClassification = () => {
+        const status       = State.getResponse('get_account_status.status');
+        const is_high_risk = /high/.test(State.getResponse('get_account_status.risk_classification'));
+
+        $('#high_risk_classification').setVisibility(
+            is_high_risk && Client.isAccountOfType('financial') ?
+                /(financial_assessment|trading_experience)_not_complete/.test(status) :
+                /financial_assessment_not_complete/.test(status)
```
negarn: this looks like exactly the check in header, but missing the jp client check. should we move it to a common function and call it in both places?

src/javascript/app/base/client.js
```diff
@@ -117,6 +117,17 @@ const Client = (() => {
 
     const isJPClient = () => State.get('is_jp_client');
 
+    const getRiskAssessment = () => {
+        const status       = State.getResponse('get_account_status.status');
+        const is_high_risk = /high/.test(State.getResponse('get_account_status.risk_classification'));
+
+        return (
+            is_high_risk && ClientBase.isAccountOfType('financial') ?
```
negarn: I think we can remove `ClientBase.` since it's in the same module

PR #3474, opened 2018-02-13
negar/new_header

negarn:
add new header for logged in client


src/sass/_common/common.scss
```diff
@@ -389,8 +389,6 @@ button .barspinner.white {
 }
 
 .static_full {
-    margin-top: 40px;
-
```
negarn: later we can remove `static_full` from pages that don't need it anymore. I didn't want this PR to get too big.

src/javascript/app/base/header.js
```diff
@@ -1,3 +1,5 @@
+import { hide_menu as hideMenu, show_menu as showMenu } from 'binary-style';
```
binary-hamedanchi: Is it possible to use `require` for consistency?

PR #3480, opened 2018-02-14
kelly/fix_spacing

kellybinary:


src/templates/app/_includes/pep_declaration.jsx
```diff
@@ -10,7 +10,7 @@ const PepDeclaration = () => (
         <div className='gr-padding-10 gr-12'>
             <input id='not_pep' type='checkbox' />
             <label htmlFor='not_pep'>
-                {it.L('I acknowledge that I am not a politically exposed person (PEP).')}
+                {it.L('I acknowledge that I am not a politically exposed person (PEP).')}{' '}
```
negarn: why not add `&nbsp` instead of a literal empty space?

kellybinary: i thought `{' '}` would be more consistent with jsx syntax, but sure we can use `&nbsp;` if that's preferred

negarn: up to you, I guess :) there are more inconsistent things in our code than spacing convention.

kellybinary: hmm, better to stick to one thing, i'll change it to `&nbsp;` :P

PR #3485, opened 2018-02-15
remove careers for americans

ruskakimov:


ashkanx:
@ruskakimov  Only remove the link in career page. keep the page itself

PR #3502, opened 2018-02-21
Lookback issues

4p00rv:


src/javascript/app/pages/trade/purchase.js
```diff
@@ -79,8 +79,8 @@ const Purchase = (() => {
             let formula, multiplier;
             const {contract_type} = passthrough;
             if (isLookback(contract_type)) {
-                multiplier = formatMoney(currency, passthrough.amount);
-                formula    = getLookBackFormula(contract_type, multiplier, false, 3, 2);
+                multiplier = formatMoney(currency, passthrough.amount, multiplier, false, 3, 2);
```
binary-hamedanchi: Seems `multiplier` shouldn't be in the list of arguments?

4p00rv: @binary-hamedanchi yeah.

PR #3505, opened 2018-02-21
new `data-show` attribute that controls element visibility based on landing company

ruskakimov:
`data-show` is an optional attribute, that will control element visibility based on the current landing company.

## How it works
All elements with `data-show` attribute and without `.visible` class are `display: none` by default. JS takes every element with `data-show` attr on the page and adds `.visible` class if current landing company passes specified conditions.

## Examples of its use:
- `data-show='-costarica'` - hide only if costarica is the landing company
- `data-show='virtual'` - show only for virtual
- `data-show='japan, maltainvest'` - show only for specified companies
- `data-show='-japan, -maltainvest'` - hide only for specified companies
- `data-show='default, costarica'` - show only for logged out and costarica users
- `data-show='-default'` - hides for logged out users
- `data-show='mtcompany'` - show only if account has mt_financial or mt_gaming company
- `data-show='-mtcompany'` - hide only if account has mt_financial or mt_gaming company

### Prohibited values:
- `data-show='-virtual, japan'` - cannot mix includes and excludes (results in ambiguous behaviour)
- `data-show='costarica,,japan'` - no empty names
- `data-show='costarica,+japan'` - no invalid characters (valid chars: `a-z - ,` )
- `data-show='Costarica'` - no capital letters

> whitespace around names  is trimmed during parsing


src/javascript/app/base/page.js
```diff
@@ -65,6 +65,63 @@ const Page = (() => {
         });
     };
 
+    const showHiddenElementsBasedOnCompany = (landing_company_name) => {
+        const VISIBLE_CLASSNAME = 'visible';
```
zakibinary: I think we should keep variable names in lower case for consistency's sake.

ruskakimov:
What about changing `.visible` class to something more specific, like `.data-show-visible`? To prevent future class collisions.

ashkanx:
@ruskakimov please resolve the conflict

PR #3514, opened 2018-02-22
kelly/add_lookback_icon

kellybinary:
- Add lookback icon to Home page


src/templates/static/home.jsx
```diff
@@ -299,9 +300,16 @@ const HomeBeta = () => {
                                     image='images/pages/home/trade/metals.svg'
                                 />
                             </TabContent>
+                            <TabContent id='lookbacks'>
+                                <MarketsTabContent
+                                    header={it.L('Lookbacks')}
+                                    text={it.L('Lookbacks options are currently only available for <a target="_blank" href=[_1]>Volatility Indices</a>.', `${it.url_for('get-started/binary-options')}#range-of-markets`)}
```
negarn: it would be good to hide the entire link in placeholder, e.g. `it.L('Lookbacks options are currently only available for [_1]Volatility Indices[_2].', `<a target="_blank" href="${it.url_for('get-started/binary-options')}#range-of-markets`">`, '</a>')`, this way translators won't get confused with html tags

PR #3524, opened 2018-02-26
Allow self excluded client to login

4p00rv:


src/javascript/app/pages/user/accounts.js
```diff
@@ -70,13 +70,7 @@ const Accounts = (() => {
     const populateExistingAccounts = () => {
         const all_login_ids = Client.getAllLoginids();
         all_login_ids
-            .filter(loginid => !Client.get('is_disabled', loginid) && !Client.get('excluded_until', loginid))
-            .sort((a, b) => a > b)
-            .forEach((loginid) => {
-                appendExistingAccounts(loginid);
-            });
-        all_login_ids
-            .filter(loginid => Client.get('is_disabled', loginid) || Client.get('excluded_until', loginid))
```
negarn: we don't want to show a list of disabled and excluded accounts in account page anymore?

4p00rv: @negarn I see. I couldn't figure out what this was for. I will add these checks back.

negarn: we can add a comment explaining what they are for :)

src/javascript/app/base/header.js
```diff
@@ -315,6 +317,7 @@ const Header = (() => {
                 'tnc',
                 'financial_limit',
                 'risk',
+                'excluded_until',
```
binary-hamedanchi: I think it should be the first item in the list. Because if a client is self-excluded, doesn't have permission to do any of the above.

src/javascript/app/base/client.js
```diff
@@ -133,6 +133,17 @@ const Client = (() => {
...
         set('landing_company_shortcode', authorize.landing_company_name);
+        updateAccountList(authorize.account_list);
+    };
+
+    const updateAccountList = (account_list) => {
+        account_list.forEach((account) => {
+            if (!account.excluded_until) {
+                set('excluded_until', undefined, account.loginid);
+            } else {
+                set('excluded_until', account.excluded_until, account.loginid);
```
negarn: aren't these set on login? why do we need to set them again on authorize response?

4p00rv: If the `exclude_until` is removed from back office, the user will need to log out and log in again for the Account status header to be removed. With this, it is updated after each reloads.

negarn: I see! then, is `set('excluded_until', account.excluded_until, account.loginid);` enough instead of an if/else? if account.excluded_until is null/undefined it will just be set as null/undefined.

negarn: and if we are doing it for excluded_until, is it worth updating all the params? the same case will happen if BO updates anything else

4p00rv: @negarn That's a good idea. I have added it.

src/javascript/app/base/client.js
```diff
@@ -133,6 +133,20 @@ const Client = (() => {
         set('is_virtual', +authorize.is_virtual);
         set('session_start', parseInt(moment().valueOf() / 1000));
         set('landing_company_shortcode', authorize.landing_company_name);
+        updateAccountList(authorize.account_list);
+    };
+
+    const updateAccountList = (account_list) => {
+        account_list.forEach((account) => {
+            set('excluded_until', account.excluded_until, account.loginid);
```
negarn: is this overwritten on line 146? do we need to explicitly mention it here?

4p00rv: @negarn We need to mention it explicitly because if the client is not `excluded_until` no property is returned from backend for it.

negarn: I see! should we change it to `account.excluded_until || ''` to be consistent with how we store undefined values on line 145 and in logged_in?

PR #3531, opened 2018-02-27
disable restricted countries in tax residence select

ruskakimov:


src/javascript/app/common/__tests__/get_app_details.js
```diff
@@ -22,7 +23,7 @@ describe('GetAppDetails', () => {
             expect(GetAppDetails.addAppIdName(2, 'Binary.com')).to.equal('Transaction performed by Binary.com (App ID: 2)');
         });
         it('doesn\'t show for same app id as current', () => {
-            expect(GetAppDetails.addAppIdName(1, 'Binary.com')).to.equal('');
+            expect(GetAppDetails.addAppIdName(getAppId(), 'Binary.com')).to.equal('');
```
negar-binary: why? getAppId won't always return 1, unless the host is binary.com

ruskakimov: yep, but if you add your app_id then this test fails

PR #3533, opened 2018-02-27
Add mt5 platform image

kellybinary:


negarn:
should we move the image to platforms folder so that we don't remove it again accidentally in the future?

negarn:
there is already a metatrader.png image there that is not being used, perhaps we can replace it with this one.

PR #3534, opened 2018-02-27
Add PWA features

ashkanx:


negarn:
icons shouldn't be in binary-style? :P

ashkanx:
agree. but this is more for new UI. anyway, we will need to refactor image folder and decide what to do with binary-style

PR #3539, opened 2018-02-28
mohammad/network_monitor

binary-hamedanchi:


src/javascript/app/base/network_monitor.js
```diff
@@ -0,0 +1,118 @@
...
+        if (isOnline() && BinarySocket.hasReadyState(2, 3)) { // CLOSING or CLOSED
+            BinarySocket.init(ws_config);
+            clearPendings();
+        } else {
+            BinarySocket.send({ ping: 1 }); // trigger a request to get stable status sooner
+        }
+    };
+
+    const setStatus = (status) => {
+        if (!isOnline() || status === pending_keys.ws_init) {
```
binarytom: The `status` field is slightly misleading here, I think: `ws_init` doesn't really mean "we're in `ws_init state", it's specifically "we had a timeout in `ws_init` state"? At the moment, the logic for this block could be read as "if we've just started the connection, display the 'Connection error' text" (whereas that doesn't actually happen until after the timeout has expired).

I'd suggest marking this as blinking while we're waiting for the initial websocket connection, by the way - seems more accurate than claiming we're connected?

binary-hamedanchi: You're right, it's misleading. Will change it.
Thanks.

PR #3543, opened 2018-03-01
kelly/remove_place_of_birth

kellybinary:


negarn:
wait, does japan need residence_list still after removing place of birth?

kellybinary:
no, i don't think so 🤔

PR #3559, opened 2018-03-05
duplicate uploads

4p00rv:


src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -264,12 +278,30 @@ const Authenticate = (() => {
...
+                if (!file.warning) {
+                    successAny = true;
+                } else {
+                    dup_files.push(file.passthrough.filename);
+                }
+            });
+            if (successAny) {
+                showSuccess();
+            } else {
+                showError({message: localize('Following file(s) were already uploaded: [_1]', [JSON.stringify(dup_files)])});
```
kellybinary: add the string to generate-static-data.js ? :)

PR #3607, opened 2018-03-20
kelly/update_virtual_topup_description

kellybinary:


src/javascript/app/pages/user/account/top_up_virtual.js
```diff
@@ -16,12 +17,12 @@ const TopUpVirtual = (() => {
 
         BinarySocket.send({ topup_virtual: '1' }).then((response) => {
             if (response.error) {
-                showMessage(localize(response.error.message), false);
+                showMessage(localize('You can only request additional funds if your virtual account balance falls below USD 1,000.00.'), false);
```
negarn: should we check for a specific error code? what if error message is something general, like a server failure?

kellybinary: yeah, i think we should 👍

src/javascript/app/pages/user/account/top_up_virtual.js
```diff
@@ -17,7 +17,11 @@ const TopUpVirtual = (() => {
 
         BinarySocket.send({ topup_virtual: '1' }).then((response) => {
             if (response.error) {
-                showMessage(localize('You can only request additional funds if your virtual account balance falls below USD 1,000.00.'), false);
+                if (/TopupVirtualError/.test(response.error.code)) {
+                    showMessage(localize('You can only request additional funds if your virtual account balance falls below USD 1,000.00.'), false);
+                } else {
+                    showMessage(response.error.message, false);
+                }
```
negarn: we could combine this into one:
```
showMessage(/TopupVirtualError/.test(response.error.code) ? localize('You can only request additional funds if your virtual account balance falls below USD 1,000.00.') : response.error.message, false);
```

PR #3625, opened 2018-03-23
kelly/no_limit

kellybinary:


src/javascript/app/pages/user/account/settings/self_exclusion.js
```diff
@@ -148,6 +160,14 @@ const SelfExclusion = (() => {
                     ['custom', { func: value => !value.length || getMoment(exclude_until_id).isAfter(moment().add(6, 'months')), message: 'Exclude time cannot be less than 6 months.' }],
                     ['custom', { func: value => !value.length || getMoment(exclude_until_id).isBefore(moment().add(5, 'years')), message: 'Exclude time cannot be for more than 5 years.' }],
                 ],
+            },
+            {
+                selector        : max_30day_turnover_id,
+                exclude_if_empty: 1,
+                value           : () => $(max_30day_turnover_id).val(),
```
negarn: do we need this? or it will just read the value of the field by id? in above validations we need a `value` property because they call custom functions, but here we're just simply getting the value of the field, something that form manager is able to do by itself

kellybinary: ohh, no we don't it then. i'll change accordingly

PR #3628, opened 2018-03-26
zaki/trade_redesign_ui

zakibinary:
Updated Trading Page UI
---
1. Updated tabs design for analysis section.
2. New markets and contracts elements design.
3. Updated text input, radio buttons and select box style.


scripts/generate-static-data.js
```diff
@@ -150,6 +150,8 @@ const texts = [
     'High-Close',
     'Close-Low',
     'High-Low',
+    '"AUD/JPY" or "Apple"',
```
negarn: where is this string being used?

scripts/generate-static-data.js
```diff
@@ -150,6 +150,8 @@ const texts = [
...
+    'Select Asset',
```
negarn: also this

src/javascript/app/common/account_opening.js
```diff
@@ -124,6 +125,7 @@ const AccountOpening = (() => {
                 }
             }
             $address_state.parent().parent().setVisibility(1);
+            Dropdown('#address_state');
```
negarn: from the function name it's not very intuitive what it's supposed to do, can we change to something like `createDropdown`, or maybe `transformDropdown` even

src/javascript/app/common/form_validation.js
```diff
@@ -92,6 +93,14 @@ const Validation = (() => {
                 });
             }
         }
+
+        // need to init Dropdown after we have responses from ws
+        const el_all_select = document.querySelectorAll('select');
+        el_all_select.forEach((el) => {
+            if (el.id !== 'tax_residence') {
```
negarn: instead of specific id, can we check for the attribute 'multiple'? e.g. something like `if (el.getAttribute('multiple') === 'multiple')` since I assume we want to leave any multi-selects alone, and that is a better way of catching all of them instead of individually excluding them

src/javascript/app/japan/knowledge_test/knowledge_test.ui.js
```diff
@@ -4,25 +4,29 @@ const urlForStatic = require('../../../_common/url').urlForStatic;
 
 const KnowledgeTestUI = (() => {
     const createTrueFalseBox = (question, show_answer) => {
-        const qid      = question.id;
-        const true_id  = `${qid}true`;
-        const false_id = `${qid}false`;
+        const qid               = question.id;
+        const true_id           = `${qid}true`;
+        const false_id          = `${qid}false`;
+        const center_text_class = 'center-text';
```
negarn: since this is just text assignment it can be done outside, e.g. on line 6 so we don't have to keep assigning it every time the function is called

src/javascript/app/japan/knowledge_test/knowledge_test.ui.js
```diff
@@ -4,25 +4,29 @@ const urlForStatic = require('../../../_common/url').urlForStatic;
...
         const $true_button = $('<input />', {
             type : 'radio',
             name : qid,
             id   : true_id,
             value: '1',
         });
-        const $true_td     = $('<td></td>').append($true_button);
+
+        const $true_el = $true_button.add(`<label for=${true_id}>&#8203</label>`);
+        const $true_td = $(`<td class=${center_text_class}></td>`).append($true_el);
```
negarn: can we create these jquery elements the same way as on line 12? so something like `$('<td />', { class: center_text_class })`

src/javascript/app/japan/knowledge_test/knowledge_test.ui.js
```diff
@@ -4,25 +4,29 @@ const urlForStatic = require('../../../_common/url').urlForStatic;
...
 
         const $false_button = $('<input />', {
             type : 'radio',
             name : qid,
             id   : false_id,
             value: '0',
         });
-        const $false_td     = $('<td></td>').append($false_button);
+        const $false_el = $false_button.add(`<label for=${false_id}>&#8203</label>`);
+        const $false_td = $(`<td class=${center_text_class}></td>`).append($false_el);
```
negarn: same here

src/javascript/app/pages/trade/analysis.js
```diff
@@ -1,13 +1,13 @@
...
+const showChart          = require('./charts/webtrader_chart').showChart;
+const Defaults           = require('./defaults');
+const getActiveTab       = require('./get_active_tab').getActiveTab;
+const GetTicks           = require('./get_ticks');
+const MBDefaults         = require('../mb_trade/mb_defaults');
+const JapanPortfolio     = require('../../japan/portfolio');
+const getElementById     = require('../../../_common/common_functions').getElementById;
+const getLanguage        = require('../../../_common/language').get;
+const State              = require('../../../_common/storage').State;
+const Url                = require('../../../_common/url');
```
negarn: why all the extra white space?

src/javascript/app/pages/trade/analysis.js
```diff
@@ -73,10 +81,24 @@ const TradingAnalysis = (() => {
...
+                    const digit_underlying_dropdown = el_digit_underlying.next('div.select-dropdown');
+
+                    // check if custom dropdown exists and sync with underlying dropdown
+                    if (digit_underlying_dropdown) {
+                        const digit_underlying_list = digit_underlying_dropdown.next('ul.select-options').children('li');
+                        const underlying_text = $('#underlying').attr('data-text');
+                        digit_underlying_dropdown.text(underlying_text);
+
+                        digit_underlying_list.not(this).each((idx, el) => {
+                            $(el).removeClass('selected');
```
negarn: we can use javascript here, `el.classList.remove('selected');`

src/javascript/app/pages/trade/analysis.js
```diff
@@ -73,10 +81,24 @@ const TradingAnalysis = (() => {
                 showChart();
             } else if (current_tab === 'tab_last_digit') {
                 const el_digit_underlying = $('#digit_underlying');
-                const underlying = $('#underlying option:selected').val();
+                const underlying = $('#underlying').val();
                 const tick       = $('#tick_count').val() || 100;
+
                 if (underlying !== el_digit_underlying.val() && el_digit_underlying.val() !== null ) {
                     el_digit_underlying.find(`option[value="${underlying}"]`).prop('selected', true).trigger('change');
+                    const digit_underlying_dropdown = el_digit_underlying.next('div.select-dropdown');
```
negarn: is this jquery? if yes, please add `$` to it, like `const $digit_underlying_dropdown`

src/javascript/app/pages/trade/analysis.js
```diff
@@ -89,14 +111,63 @@ const TradingAnalysis = (() => {
                 showExplanation();
             }
         }
+        if (current_tab) {
+            const el_to_show = getElementById(current_tab);
+            slideSelector(tab_selector_id, el_to_show);
+            const mobile_tab_header = document.getElementById('tab_mobile_header');
```
negarn: any reason why we're using `document. getElementById` instead of `getElementById` as above (line 115)? also missing the `el_` in the beginning to make it consistent

src/javascript/app/pages/trade/analysis.js
```diff
@@ -89,14 +111,63 @@ const TradingAnalysis = (() => {
...
+            if (mobile_tab_header) mobile_tab_header.innerHTML = el_to_show.firstChild.innerHTML;
+        }
+        // workaround for underline during window resize
+        window.addEventListener('resize', () => {
```
negarn: we should probably remove this event listener somewhere, maybe onUnload of trade page

src/javascript/app/pages/trade/analysis.js
```diff
@@ -89,14 +111,63 @@ const TradingAnalysis = (() => {
...
+        }
+    };
+
+    const goLeft = (e) => {
+        changeTab({ selector: e.target.getAttribute('data-parent'), direction: 'left' });
+    };
+
+    const goRight = (e) => {
+        changeTab({ selector: e.target.getAttribute('data-parent'), direction: 'right' });
+    };
```
negarn: I'm not sure if the three above functions belong here. they seem better suited for a common file and then imported here.

src/javascript/app/pages/trade/analysis.js
```diff
@@ -89,14 +111,63 @@ const TradingAnalysis = (() => {
...
+                index_to_show = active_index + 1;
+                index_to_show = index_to_show === array_length ? 0 : index_to_show;
+            }
+        }
+        options.el_to_show = selector_array[index_to_show].id;
+        if (!options.el_to_show || !options.selector) {
+            return;
+        }
+        sessionStorage.setItem('currentAnalysisTab', options.el_to_show);
+        if (!getElementById(options.el_to_show).classList.contains('active')) {
```
negarn: it would be good to move this function too, and add a param `onChangeTab` etc which accepts a function so we could pass this if block to it and call it at the end of the block of code

scripts/generate-static-data.js
zakibinary: @4p00rv , it's used as the placeholder search string for asset dropdown right?

src/javascript/app/pages/trade/charts/digit_info.js
```diff
@@ -108,6 +109,8 @@ const DigitInfo = (() => {
         $('#digit_underlying').html($(elem)).val(underlying);
         $('#digit_domain').text(domain.charAt(0).toUpperCase() + domain.slice(1));
         $('#digit_info_underlying').text($('#digit_underlying option:selected').text());
+        Dropdown('#digit_underlying');
+        Dropdown('#tick_count');
```
negarn: couldn't we add two selectors at the same time? have you tried `Dropdown('#digit_underlying, #tick_count');`? theoretically jquery should be able to handle this, not sure about any exceptions in the code that would break with this

src/javascript/app/pages/trade/common.js
```diff
@@ -21,182 +18,25 @@ const commonTrading = (() => {
...
-            if (current.disabled) { // there is no open market
-                Notifications.show({ text: localize('All markets are closed now. Please try again later.'), uid: 'MARKETS_CLOSED' });
-                getElementById('trading_init_progress').style.display = 'none';
-            }
+        if (!contracts_element) {
+            contracts_element = contractsElement.init(elements, contracts_tree);
+        } else { // Update the component.
+            contracts_element.updater.enqueueSetState(contracts_element, {
+                contracts: elements,
+                contracts_tree,
```
negarn: variable that doesn't need a value assigned to it should move to the first line

src/javascript/app/pages/trade/contracts.jsx
```diff
@@ -0,0 +1,175 @@
...
+import {getElementById} from '../../../_common/common_functions';
+import {localize} from '../../../_common/localize';
+
+class Contracts extends React.Component {
+    constructor (props) {
+        super(props);
+        const {contracts, contracts_tree} = props;
+        const formname = Defaults.get('formname');
+        this.references = {};
+        this.$contract = getElementById('contract');
```
negarn: why not `el_contract`?

src/javascript/app/pages/trade/contracts.jsx
```diff
@@ -0,0 +1,175 @@
...
+
+    openDropDown = () => {
+        if (this.state.contracts_tree.length <= 1) return;
+        this.setState({open: true});
+        this.positionDropDown();
+    };
+    closeDropDown = () => this.setState({open: false});
+
+    positionDropDown = () => {
+        const $dropdown = this.references.wrapper;
```
negarn: why not el_dropdown?

src/javascript/app/pages/trade/contracts.jsx
```diff
@@ -0,0 +1,175 @@
...
+        this.setState({formname});
+    }
+    /* eslint-enable no-undef */
+    /* eslint-disable class-methods-use-this */
+    dismiss() {
+        console.warn('removing');
+    }
+    /* eslint-enable class-methods-use-this */
+
+    render () {
```
negarn: `dismiss()` has no space before brackets but `render ()` has space, we should make this consistent with the rest of the codebase

src/javascript/app/pages/trade/contracts.jsx
```diff
@@ -0,0 +1,175 @@
...
+        const { contracts, contracts_tree, open,
+            formname } = this.state;
```
negarn: either all should be on one line each or all on the same line

src/javascript/app/pages/trade/markets.jsx
```diff
@@ -0,0 +1,441 @@
...
+    const requestAnimationFrame = window.requestAnimationFrame ||
+        function (...args) {
+            return setTimeout(args[0], 10);
+        };
+    if (duration <= 0) {
+        element.scrollTop = to;
+        return;
+    }
+    const difference = to - element.scrollTop;
+    const perTick = difference / duration * 10;
```
negarn: variable names are separated by `_`, not camelCase

src/javascript/app/pages/trade/markets.jsx
```diff
@@ -0,0 +1,441 @@
...
+            },
+            underlying: {
+                symbol: underlying_symbol,
+                name  : this.underlyings[underlying_symbol],
+            },
+        });
+
+        // Trigger change event.
+        // TODO: move this block to componentDidUpdate
+        this.$underlying.value = underlying_symbol;
```
negarn: why not el_underlying?

src/javascript/app/pages/trade/markets.jsx
```diff
@@ -0,0 +1,441 @@
...
+        const market = e.target.dataset.market;
+        this.scrollToElement(`${market}_market`, 120, 0);
+    }
+
+    saveRef = (node_name, node) => this.references[node_name] = node;
+
+    scrollToElement = (id, duration = 120, offset = 0) => {
+        // handleScroll is triggered automatically which sets the active market.
+        const {list} = this.references;
+        const toOffset = getElementById(id).offsetTop - list.offsetTop - offset;
```
negarn: variables should be separated by `_`

src/javascript/app/pages/trade/markets.jsx
```diff
@@ -0,0 +1,441 @@
...
+
+        const class_sticky = 'sticky';
+        const class_under = 'put_under';
+        const DEFAULT_TOP = this.references.list.offsetTop;
+
+        if (curr) {
+            curr.children[0].removeAttribute('style');
+            curr.removeAttribute('style');
+            curr.children[0].classList.remove(class_under);
+            const diff = (+curr.dataset.offsetHeight + +curr.dataset.offsetTop) - position ;
```
negarn: extra space before semi-colon

src/javascript/app/pages/trade/markets.jsx
```diff
@@ -0,0 +1,441 @@
...
+        }
+        if (prev) {
+            prev.removeAttribute('style');
+            prev.children[0].removeAttribute('style');
+            prev.children[0].classList.remove(class_under);
+            prev.children[0].classList.remove(class_sticky);
+        }
+        if (next) {
+            next.children[0].classList.remove(class_sticky);
+            next.children[0].classList.remove(class_under);
```
negarn: you can remove multiple classes in the same line, e.g. `next.children[0].classList.remove(class_sticky, class_under)`. same for above

PR #3631, opened 2018-03-26
Add tradingview and smartchart links to header

ruskakimov:


negarn:
do we want to hide these for japan? if so, they all need the ja-hide class, similar to asset index

ruskakimov:
@negarn done

src/templates/app/_layout/header.jsx
```diff
@@ -123,6 +123,8 @@ const Header = () => (
                                             subitems : [
                                                 { text: it.L('Asset Index'),   href: it.url_for('resources/asset_indexws'), className: 'ja-hide' },
                                                 { text: it.L('Trading Times'), href: it.url_for('resources/market_timesws') },
+                                                { text: it.L('TradingView'),   href: 'https://tradingview.binary.com', target: '_blank', className: 'ja-hide' },
```
negarn: could you align them so className is on the same indentation as asset index's className property? then the targets should align together after that :)

PR #3639, opened 2018-03-28
Always show website status message when available

zakame:
This allows the site to observe `website_status.message` from the websocket API and show it to users when available, even when the site status is "up".


src/javascript/app/base/socket_general.js
```diff
@@ -31,14 +31,18 @@ const BinarySocketGeneral = (() => {
     const onMessage = (response) => {
         Header.hideNotification('CONNECTION_ERROR');
         let is_available = false;
+        let has_message = false;
```
negarn: I don't see why we need to declare it false here, and then assign it on line 39. why not just change line 42 to `} else if (response.website_status.message) {`

src/javascript/app/base/socket_general.js
```diff
@@ -31,14 +31,18 @@ const BinarySocketGeneral = (() => {
...
         switch (response.msg_type) {
             case 'website_status':
                 if (response.website_status) {
                     is_available = /^up$/i.test(response.website_status.site_status);
+                    has_message = typeof response.website_status.message === 'string';
                     if (is_available && !BinarySocket.availability()) {
                         window.location.reload();
-                    } else if (!is_available) {
+                    } else if (has_message) {
                         Header.displayNotification(response.website_status.message, true);
```
negarn: what happens if we have other messages showing on header? e.g. authentication message, tnc message, etc. will they get overridden by the website_status message all the time? if so, then we need another way of showing site message since if website is up and there is a message, it still should let client know about any pending actions they should take, e.g. authentication etc

src/javascript/app/base/socket_general.js
zakame: Yeah, that's an artifact of the previous commits, thanks for catching this :+1:

src/javascript/app/base/socket_general.js
zakame: `Header.displayNotification` seems to be used in only 3 other places, and testing my current changes shows these behaviors:

- At `BinaryLoader.init`: this would always override the website status message
- At `NetworkMonitor.updateHeaderNotification` when it detects a failing connection: this would override *any* message (even the above,) as well as make itself disappear in a very short while.
- At `Authenticate.showSuccess`: a bit trickier to test, but it also overrides the website status message

It seems better to have the statuses be separate from the existing header messages though, perhaps also working with `NetworkMonitor`...

zakame:
Superseded by https://github.com/binary-com/binary-static/pull/3682

PR #3641, opened 2018-03-28
Affiliate FAQ: show sidebar on mobile

ruskakimov:


src/sass/static/affiliates.scss
```diff
@@ -6,6 +6,15 @@
             margin-top: 3rem;
         }
     }
+    @media screen and (max-width: 480px) {
+        .sidebar-container {
```
negarn: can we use the same styling that terms and conditions menu is using? if not then it's fine to do it separately here

ruskakimov: `sidebar-collapsible` (in terms and conditions) styles are not specific to mobile :)
Also I want to preserve `sidebar` desktop styles identical to the one on get-started/binary-options and others

PR #3647, opened 2018-03-30
Affiliate FAQ back button

ruskakimov:


src/templates/static/affiliates/faq.jsx
```diff
@@ -103,6 +103,13 @@ const FAQ = () => (
                     <h3>{it.L('Where can I send my questions, comments, and suggestions?')}</h3>
                     <p>{it.L('Feel free to get in touch with us at [_1]. We’d love to hear from you.', `<a href='mailto:${it.affiliate_email}'>${it.affiliate_email}</a>`)}</p>
                 </Section>
+
+                <div className='border-bottom divider' />
```
negarn: how about using `<SeparatorLine/>`? with something like class gr-padding-30, and the element below can be a gr-parent so it has no padding on top

PR #3648, opened 2018-03-30
IB programme page

ruskakimov:


src/javascript/app/base/binary_pages.js
```diff
@@ -103,6 +103,7 @@ const pages_config = {
     self_exclusionws         : { module: SelfExclusion,              is_authenticated: true, only_real: true },
     settingsws               : { module: Settings,                   is_authenticated: true },
     signup                   : { module: TabSelector },
+    'ib-signup'              : { module: TabSelector },
```
negarn: generally all the single quoted pages are at the bottom

src/javascript/app/base/binary_pages.js
```diff
@@ -117,6 +118,7 @@ const pages_config = {
     'cryptocurrencies'       : { module: GetStarted.Cryptocurrencies },
     'deposit-jp'             : { module: CashierJP.Deposit,          is_authenticated: true, only_real: true },
     'faq'                    : { module: StaticPages.AffiliatesFAQ },
+    'ib-faq'                 : { module: StaticPages.IBProgrammeFAQ },
```
negarn: this should be alphabetical according to page url, so it should be on line 127

src/templates/static/_layout/footer.jsx
```diff
@@ -173,7 +173,8 @@ const FooterNormal = () => (
                             <FooterColumn
                                 header={it.L('Partner With Us')}
                                 items={[
-                                    { text: it.L('Affiliate Program'),       href: it.url_for('affiliate/signup') },
+                                    { text: it.L('Affiliate Programme'),       href: it.url_for('affiliate/signup') },
```
negarn: the `href` is not aligned

src/templates/static/affiliates/signup.jsx
```diff
@@ -40,6 +40,7 @@ const Signup = () => (
                 <div className='steps'>
                     <Step circle_no='1' header={it.L('Sign up')}                       text={it.L('Getting started is free and easy – just <a href=\'[_1]\'>fill out the application form</a> and wait for our approval.', it.affiliate_signup_url)} />
                     <Step circle_no='2' header={it.L('Promote [_1]', it.website_name)} text={it.L('Use your unique affiliate link and the marketing tools we provide to advertise [_1] to your audience.', it.website_name)} />
+                    <Step circle_no='3' header={it.L('Earn')}                          text={it.L('Choose from two types of  commission plans when your referred clients trade binary options on our platform.')} />
```
negarn: was this supposed to replace line 44?

src/sass/static/pages.scss
```diff
@@ -1,4 +1,80 @@
...
+            margin-bottom: 2rem;
+
+            .sidebar a {
+                line-height: 2.5rem;
+            }
+        }
+    }
+}
+
+.signup-page {
```
negarn: since it's about affiliate signup page, I would recommend a clearer class name, since we also have binary account signup page and mt account sign up page. also the block is pretty long so it would make sense to keep it in its own file as before.

ruskakimov: these styles are used by both affiliate and IB sign up pages, i guess i can use specific classes for them and declare styles like this: `.affiliate-signup, .ib-signup {`
But then do i keep it in pages.scss? 🤔

negarn: I think it's still kind of an affiliate page, so should be fine in affiliate file :P but yeah, using different class names for each and assigning the same style to both sounds good.

src/templates/static/affiliates/signup.jsx
```diff
@@ -2,13 +2,13 @@ import React from 'react';
 import { Table } from '../../_common/components/elements.jsx';
 import { TabContainer, TabContent, TabContentContainer, TabsSubtabs } from '../../_common/components/tabs.jsx';
 
-const BoxRow = ({ children, top_row, bottom_row }) => (
-    <div className={`box-row gr-row ${top_row ? 'box-row-top' : ''} ${bottom_row ? 'box-row-bottom' : ''}`}>
+export const BoxRow = ({ children, top_row, bottom_row, no_border }) => (
```
negarn: not sure if it's fine to keep these element definitions here or it's better to move them to a more common folder and import it in both affiliates and ib pages. up to you, I'm okay with both

ruskakimov: @negarn component uses hardcoded path for `images/pages/affiliates`, so I think it's better to keep it here. I don't think it can be easily reused on other pages.

negarn: sounds good to me.

ruskakimov:
@negarn could you please review the changes? :)

src/templates/static/ib_programme/ib_signup.jsx
```diff
@@ -49,6 +49,7 @@ const Signup = () => (
             <h2 className='center-text'>{it.L('Commission structure')}</h2>
             <div className='gr-padding-20 center-text'>
                 <Table
+                    className='ib-signup-page__commission-table'
```
negarn: I don't think we have this style of `__`, why not just use `commission-table` and then in sass file call it by `.ib-signup-page { .commission-table ...`

negarn: oh did you already change the code, I'm going through each commit. but there'a  lot of `--` and `__` if we want to use it we can, we just need to add documentation for it so we consistently use it everywhere

ruskakimov: also which documentation you were talking about? :) And probably we should consult with Mohammad first

negarn: documentation could be added to a readme file in the sass folder.

PR #3656, opened 2018-04-03
negar/jp_client_fixes

negarn:
add {JAPAN ONLY} to all japanese text and simplify jpClient condition


src/javascript/app/pages/user/account/statement/statement.init.js
```diff
@@ -81,7 +81,7 @@ const StatementInit = (() => {
                             .append($('<p/>', { class: 'notice-msg center-text', text: localize('Your account has no trading activity.') }))));
             } else {
                 $('#util_row').setVisibility(1);
-                if (getLanguage() === 'JA' && jpResidence()) {
+                if (getLanguage() === 'JA' && Client.get('residence') !== 'jp') {
```
kellybinary: shouldn't this be `===` ?

src/javascript/app/common/currency.js
```diff
@@ -41,7 +41,9 @@ const addComma = (num, decimal_points, is_crypto) => {
     ));
 };
 
-const getFiatDecimalPlaces = () => jpClient() ? 0 : 2;
+const isJPClient = () => (LocalStore.getObject('client.accounts')[LocalStore.get('active_loginid')] || {}).is_jp;
```
kellybinary: can we use `Client.get('is_jp')` here?

negarn: no, because client is calling currency.js, so currency.js can't call client or it will create a circular dependency. we can fix this when we change to import instead of require

PR #3677, opened 2018-04-06
ruskakimov/chart_mobile_optimize

ruskakimov:
## [Trello card](https://trello.com/c/CHJqH6iJ)

### DIFF
--- a/src/sass/app/trade.scss
+++ b/src/sass/app/trade.scss
@@ -567,28 +567,6 @@
             width: 70% !important;
         }
     }
-    // fix trading chart on small screens
-    .chart-dialog {
-        min-width: 300px !important;
-
-        .chartOptions {
-            margin: 0 !important;
-            position: relative;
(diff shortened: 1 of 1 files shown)


src/sass/app/trade.scss
```diff
@@ -668,6 +669,28 @@
...
+                top: 15px;
+            }
+        }
+        svg.highcharts-root {
+            width: 100vw !important;
+            margin: 0 -10px !important;
+            position: relative;
+            left: -10px;
+        }
+    }
```
negarn: these are all elements inside the webtrader chart container, shouldn't they be added to webtrader repo instead of binary-static? @4p00rv please comment

4p00rv: @negarn I agree. 
@ruskakimov can you please submit a separate PR to [webtrader-charts](https://github.com/binary-com/webtrader-charts) for the fix.

PR #3682, opened 2018-04-06
Oskar/websocket notifications

OskarAhl:
For Outage notifications through websocket API:
- Add 'warning' notifcation box to footer to display message
- Update socketgeneral.js to show message if server is available and there is a message


src/sass/_common/base/mixin.scss
```diff
@@ -97,3 +97,42 @@
         }
     }
 }
+// Handles styling and icon display for <StatusNotification /> in footer 
+@mixin notificationType($type) {
```
ashkanx: Do we still need different `Types`?

OskarAhl:
Removed notification mixin + unused icons (info, danger, success)

src/javascript/app/base/footer.js
```diff
@@ -0,0 +1,22 @@
+const getElementById         = require('../../_common/common_functions').getElementById;
```
negarn: the equal sign can be closer to the declaration, no need for extra space

src/javascript/app/base/footer.js
```diff
@@ -0,0 +1,22 @@
...
+
+const Footer = (() => {
+    const displayNotification = (message) => {
+        const status_notification = getElementById('status-notification');
```
negarn: usually for ID we use `_` instead of `-`, same here and the line below

src/javascript/app/base/footer.js
```diff
@@ -0,0 +1,22 @@
...
+        const status_message_text = getElementById('status-notification-text');
+        const close_icon = getElementById('status-notification-close');
+
+        $(status_notification).css('display', 'flex');
```
negarn: if it's going to be a jquery element, why not store it as jquery?

src/javascript/app/base/footer.js
```diff
@@ -0,0 +1,22 @@
...
+        status_message_text.html(message);
+
+        $(close_icon).click(() => {
```
negarn: should we also handle off click to avoid multiple event listeners on the same element?

src/templates/app/_layout/footer.jsx
```diff
@@ -18,6 +18,18 @@ const SocialIcons = ({ networks, is_centered }) => (
...
 );
 
+const StatusNotification = () => (
+    <div id='status-notification'>
+        <div id='status-notification-type'>
+            <img src={it.url_for('images/server_status/ic-warning.svg')}  alt='information icon'/>
+        </div>
+        <p id='status-notification-text'/>
+        <div id='status-notification-close'>
+            <img src={it.url_for('images/server_status/ic-close.svg')}  alt='Warning icon'/>
```
negarn: the alt here starts with an uppercase letter while on line 24 it's lowercase, better to keep it consistent

src/templates/static/_layout/footer.jsx
```diff
@@ -18,6 +18,18 @@ const SocialIcons = ({networks, is_centered}) => (
...
 );
 
+const StatusNotification = () => (
+    <div id='status-notification'>
+        <div id='status-notification-type'>
+            <img src={it.url_for('images/server_status/ic-warning.svg')}  alt='information icon'/>
+        </div>
+        <p id='status-notification-text'/>
+        <div id='status-notification-close'>
+            <img src={it.url_for('images/server_status/ic-close.svg')}  alt='Warning icon'/>
```
negarn: same here

src/javascript/app/base/footer.js
```diff
@@ -0,0 +1,20 @@
+const Footer = (() => {
+    const displayNotification = (message) => {
+        const status_notification = $('#status_notification');
+        const status_message_text = $('#status_notification_text');
+        const close_icon = $('#status_notification_close');
```
negarn: for jquery variables, please start the name with `$`, e.g. `$status_notification` etc

src/javascript/app/base/footer.js
```diff
@@ -0,0 +1,20 @@
...
+
+        $(status_notification).css('display', 'flex');
```
negarn: it's already jquery, so no need to wrap it in `$(...)`

src/javascript/app/base/footer.js
```diff
@@ -0,0 +1,20 @@
...
+        status_message_text.html(message);
+
+        $(close_icon).off('click').on('click', () => {
```
negarn: same here

src/javascript/app/base/socket_general.js
```diff
@@ -2,6 +2,7 @@ const Client               = require('./client');
 const Clock                = require('./clock');
 const GTM                  = require('./gtm');
 const Header               = require('./header');
+const Footer               = require('./footer');
```
negarn: should be alphabetical according to pathname, so should be on line 3

src/javascript/app/base/socket_general.js
```diff
@@ -37,6 +38,8 @@ const BinarySocketGeneral = (() => {
                     is_available = /^up$/i.test(response.website_status.site_status);
                     if (is_available && !BinarySocket.availability()) {
                         window.location.reload();
+                    } else if (is_available && response.website_status.message) {
+                        Footer.displayNotification(response.website_status.message);
                     } else if (!is_available) {
                         Header.displayNotification(response.website_status.message, true);
                     }
```
negarn: we could change the block, so it becomes:
```
if (is_available) {
    if (!BinarySocket.availability()) {
        window.location.reload();
    } else if (response.website_status.message)
        Footer.displayNotification(response.website_status.message);
    }
} else {
    Header.displayNotification(response.website_status.message, true);
}
```
but it's up to you, this is fine as well

src/javascript/app/base/socket_general.js
```diff
@@ -35,9 +36,13 @@ const BinarySocketGeneral = (() => {
...
-                        window.location.reload();
-                    } else if (!is_available) {
+                    if (is_available) {
+                        if (!BinarySocket.availability()) {
+                            window.location.reload();
+                        } else if (response.website_status.message) {
+                            Footer.displayNotification(response.website_status.message);
+                        }
+                    } else {
                         Header.displayNotification(response.website_status.message, true);
```
zakame: I think this line can go away now that `website_status.message` will always go to the Footer.

PR #3684, opened 2018-04-09
hide error after some time

4p00rv:


src/javascript/app/pages/cashier/account_transfer.js
```diff
@@ -118,6 +118,7 @@ const AccountTransfer = (() => {
             const el_error = getElementById('form_error');
             elementTextContent(el_error, response.error.message);
             el_error.setVisibility(1);
+            setTimeout(() => el_error.setVisibility(0), 5000);
```
jy-binary: add a comment to explain, for future reference?

PR #3687, opened 2018-04-10
kelly/fix_regulation_page_map_coordinates

kellybinary:


src/javascript/static/pages/regulation.js
```diff
@@ -19,6 +19,13 @@ const Regulation = (() => {
             if ($element && $element.length) {
                 $.scrollTo($element, 500);
             }
+
+            $accordion.on('accordionactivate', () => {
+                // if EU passport rights tab is active, call relocateLinks to initialize map coordinates
+                if ($accordion.accordion('option', 'active') === 0) {
```
gokula-krishna-dev: We can use if(!$accordion.accordion('option', 'active'))

PR #3688, opened 2018-04-10
kelly/add_Ladders_to_trade_menu

kellybinary:


ashkanx:
Show it to `CR` & `VRTC` only

PR #3695, opened 2018-04-11
negar/mam

negarn:
handle mam account


src/javascript/app/pages/user/metatrader/metatrader.js
```diff
@@ -54,15 +54,14 @@ const MetaTrader = (() => {
                 const title            = localize(`${toTitleCase(type)} ${company_info.title}`);
                 const is_demo          = type === 'demo';
 
-                if (!is_demo || !company_info.is_real_only) { // only add demo accounts for acc_type without is_real_only flag
+                if (!(is_demo && company_info.is_real_only)) {
```
kellybinary: should we keep it as `||` ?

PR #3696, opened 2018-04-11
Oskar/ux improvements

OskarAhl:
1. Add note to Account transfer 
2. Update topheader btn text
3. Add affiliate contact email to Contact
4. Add link to authenticate message withdraw_deposit


src/javascript/app/base/header.js
```diff
@@ -178,7 +178,7 @@ const Header = (() => {
                         }
                     }
                 } else if (show_upgrade_msg) {
-                    showUpgrade(upgrade_info.upgrade_link, `Open a ${toTitleCase(upgrade_info.type)} Account`);
+                    showUpgrade(upgrade_info.upgrade_link, `Click here to open a ${toTitleCase(upgrade_info.type)} Account`);
```
negarn: you also need to update translation strings in generate-static-data.js

src/templates/app/cashier/deposit_withdraw.jsx
```diff
@@ -18,7 +18,7 @@ const DepositWithdraw = () => (
                 {it.L('There was a problem validating your personal details. Please update your [_1] <a href="[_2]">here</a>.', '[_1]', it.url_for('user/settings/detailsws'))}
                 <br /> {it.L('If you need assistance feel free to contact our <a href="[_1]">Customer Support</a>.', it.url_for('contact'))}
             </p>
-            <p id='not_authenticated_message' />
+            <a id='not_authenticated_message' href={it.url_for('user/authenticate')} />
```
negarn: this means that the whole message will become a link. it's better to only highlight `authenticate` word. my suggestion is that instead of showing the back-end error message which stops us from linking just one word, use a front-end error message instead

scripts/generate-static-data.js
```diff
@@ -12,7 +12,7 @@ const texts = [
     'Please [_1]log in[_2] or [_3]sign up[_2] to view this page.',
 
     // top bar
-    'Open a Real Account',
+    'Click here to open a Real Account',
     'Open a Financial Account',
```
negarn: also here

src/templates/app/cashier/deposit_withdraw.jsx
```diff
@@ -18,7 +18,7 @@ const DepositWithdraw = () => (
                 {it.L('There was a problem validating your personal details. Please update your [_1] <a href="[_2]">here</a>.', '[_1]', it.url_for('user/settings/detailsws'))}
                 <br /> {it.L('If you need assistance feel free to contact our <a href="[_1]">Customer Support</a>.', it.url_for('contact'))}
             </p>
-            <a id='not_authenticated_message' href={it.url_for('user/authenticate')} />
+            <p id='not_authenticated_message'>{it.L('Please <a href="[_1]">authenticate</a> your account.', it.url_for('user/authenticate'))}</p>
```
negarn: It would be better to use something like this:

```
{it.L('Please [_1]authenticate[_2] your account.', '<a href=`${it.url_for('user/authenticate')}`>', '</a>')}
```

So that translators don't have to deal with any HTML tags. I know the lines above are using tags in strings, but it's from old code and we kept it since changing it means translators have to re-do them and creates extra work.

PR #3700, opened 2018-04-11
kelly/update_tnc_header

kellybinary:


negarn:
it's good if this PR also goes to translation instead of master, but this is fine too

PR #3703, opened 2018-04-12
negar/fix_config_undefined

negarn:
wait for socket init to avoid race condition


src/javascript/app/base/network_monitor.js
```diff
@@ -30,24 +30,27 @@ const NetworkMonitor = (() => {
...
+            } else { // if not supported, default to online and fallback to WS checks
+                navigator.onLine = true;
+            }
 
-        if (isOnline()) {
-            BinarySocket.init(ws_config);
-        }
+            if (isOnline()) {
+                BinarySocket.init(ws_config);
+                resolve(); // need to know when socket init is done
```
kellybinary: do we need to handle `reject` as well? 🤔

negarn: you're right, if it's not online it shouldn't wait indefinitely. it should resolve when wsReconnect manages to call binarysocket.init. I can't think of a good way to handle that. let me think about it.

negarn: I think a better way to handle this would be to move the whole thing to the onOpen event of socket. So we can wait and be sure that socket was definitely initialized and opened before calling wait for authroize. So I'll do that instead.

PR #3706, opened 2018-04-12
kelly/update_after_setting_currency_page_header

kellybinary:


src/javascript/app/pages/user/set_currency.js
```diff
@@ -93,6 +93,8 @@ const SetCurrency = (() => {
                             } else {
                                 Header.populateAccountsList(); // update account title
                                 $('.select_currency').setVisibility(0);
+                                $('#hide_new_account').setVisibility(0);
+                                $('#show_new_account').setVisibility(1);
```
negarn: should we only show this if `is_new_account`? I don't remember how this page works anymore 😅

PR #3714, opened 2018-04-13
ruskakimov/tick_trade_ui

ruskakimov:


src/javascript/app/pages/trade/purchase.js
```diff
@@ -45,7 +45,7 @@ const Purchase = (() => {
         const button             = CommonFunctions.getElementById('contract_purchase_button');
 
         const error      = details.error;
-        const show_chart = !error && passthrough.duration <= 10 && passthrough.duration_unit === 't' && (sessionStorage.formname === 'risefall' || sessionStorage.formname === 'higherlower' || sessionStorage.formname === 'asian');
+        const show_chart = !error && passthrough.duration <= 10 && passthrough.duration_unit === 't' && (sessionStorage.formname === 'risefall' || sessionStorage.formname === 'higherlower' || sessionStorage.formname === 'asian' || sessionStorage.formname === 'touchnotouch');
```
negarn: wouldn't it be easier to do a regex check instead of the long chain of `sessionStorage.formname === 'risefall' || sessionStorage.formname === 'higherlower' || sessionStorage.formname === 'asian' || sessionStorage.formname === 'touchnotouch'`? something like `/^risefall|higherlower|asian|touchnotouch$/.test(sessionStorage.formname)`

src/javascript/app/pages/user/view_popup/view_popup.js
```diff
@@ -166,13 +166,15 @@ const ViewPopup = (() => {
             current_spot_time = user_sold ? '' : contract.exit_tick_time;
         }
 
-        if (current_spot) {
+        const is_touchnotouch_tick_contract = /touch/i.test(contract.contract_type) && contract.tick_count;
+
+        if (current_spot && !is_touchnotouch_tick_contract) {
```
negarn: we don't want to hide it for all, only touch/no touch contracts that were sold before expiry

src/javascript/app/pages/user/view_popup/view_popup.js
```diff
@@ -135,7 +135,7 @@ const ViewPopup = (() => {
     const update = () => {
         const final_price       = contract.sell_price || contract.bid_price;
         const is_started        = !contract.is_forward_starting || contract.current_spot_time > contract.date_start;
-        const user_sold         = contract.sell_time && contract.sell_time < contract.date_expiry;
+        const user_sold         = contract.sell_spot_time && +contract.sell_spot_time < contract.date_expiry;
```
negarn: I'm worried this will break for other contracts, because it's being used everywhere. @binary-hamedanchi ?

binary-hamedanchi: Since this is strictly about _sold by user_, we should check `sell_time` as the time which sell command triggered.
`sell_spot_time` is the time of the spot that contract is actually sold, which can happen after expiry.
If `sell_time` is after `date_expiry` then we can assume that it's not sold _by user_.

PR #3737, opened 2018-04-18
[new-app] ruskakimov/react_trading_statement

ruskakimov:


src/javascript/app/pages/trade/common.js
```diff
@@ -389,6 +389,31 @@ const commonTrading = (() => {
         };
     };
 
+    /*
+     * Function is called only once each ${wait} seconds
+     * last call is debounced
+     */
+    const throttlebounce = (func, wait = 500) => {
```
binary-hamedanchi: Please move this function to `javascript/_common` folder, since it's not dependent and it could be used in other sections.
Thanks.

src/javascript/app_2/pages/statement/statement.jsx
```diff
@@ -80,53 +83,81 @@ class Statement extends React.PureComponent {
...
+        }
+
         return (
             <DataTable
-                {...this.props}
-                data_source={this.state.data_source}
+                data_source={this.state.data_source.slice(0, this.state.chunks * this.props.chunk_size)}
                 columns={this.state.columns}
-                onCloseToEnd={this.getNextBatch}
+                has_fixed_header={true}
```
binary-hamedanchi: Is it the same if removing the `={true}` value?

src/javascript/app_2/components/elements/data_table.jsx
```diff
@@ -166,40 +36,61 @@ class DataTable extends React.Component {
...
             </div>
         );
     }
 }
 
 DataTable.defaultProps = {
-    pagination        : true,
-    page_size         : 10,
-    pages_close_to_end: 5,
+    has_fixed_header: false,
```
binary-hamedanchi: Is this needed since it initially is false?

ruskakimov: good point, `undefined` will do as well :D

src/javascript/app_2/components/elements/data_table.jsx
```diff
@@ -166,40 +36,61 @@ class DataTable extends React.Component {
...
     renderHeaders() {
-        return this.props.columns.map(col => <th key={col.data_index}>{col.title}</th>);
+        return this.props.columns.map(col => <th className={col.data_index} key={col.data_index}>{col.title}</th>);
     }
 
-    render() {
-        const { pagination } = this.props;
+    renderTableClone() {
+        {/*
+            cloned table with one row for fixed header
```
binary-hamedanchi: Not sure cloning with only the first row will cover all cases.
i.e. when there is a longer cell in another row that causes changing the columns width.

ruskakimov: @binary-hamedanchi I have applied `table-layout: fixed` to the table. 🙂
So, according to spec:
> The table and column widths are set by the widths of table and col or by the width of the first row of cells. Cells in other rows do not affect column widths.

PR #3738, opened 2018-04-18
[new-app] set up functions for sending proposal

negarn:


negarn:
pending moving date picker logic out of the element

PR #3761, opened 2018-04-24
Oskar/portfolio

OskarAhl:
- Portfolio component


src/javascript/app_2/data/dao.js
```diff
@@ -6,24 +6,41 @@ const DAO = (() => {
...
     const getPayoutCurrencies = () => BinarySocket.send({ payout_currencies: 1 });
 
+    const getPortfolio = () => BinarySocket.send({ portfolio: 1 });
+
     const getWebsiteStatus = () => BinarySocket.send({ website_status: 1 });
 
+    const sellExpired = () => BinarySocket.send({ sell_expired: 1 });
+
     // ----- Streaming calls -----
+    const subscribeProposalOpenContract = (type, cb, should_forget_first) =>
```
binary-hamedanchi: Since the value of `proposal_open_contract ` is always `1`, I'd suggest to pass `contract_id` instead of `type`, then include it in the request object only if it's available.

src/javascript/app_2/data/dao.js
```diff
@@ -6,24 +6,41 @@ const DAO = (() => {
...
+        SubscriptionManager.subscribe('proposal_open_contract', { proposal_open_contract: type, subscribe: 1 }, cb, should_forget_first);
+
     const subscribeTicks = (symbol, cb, should_forget_first) =>
         SubscriptionManager.subscribe('ticks', { ticks: symbol, subscribe: 1 }, cb, should_forget_first);
 
+    const subscribeTransaction = (type, cb, should_forget_first) =>
```
binary-hamedanchi: Might consider removing the `type` from arguments, as it's not optional and its value is always `1`.

src/javascript/app_2/pages/portfolio/portfolio.jsx
```diff
@@ -0,0 +1,277 @@
...
+                style : '',
+            },
+        };
+    });
+    return formatted_portfolio;
+};
+
+// TODO: move to common
+const contract_type_display = {
+    ASIANU      : {name: localize('Asian Up'), icon: 'asia_up'},
```
binary-hamedanchi: How about using lowercase contract_types as filenames and put all in one folder according to theme or usage, then no need to map the `icon` everywhere?
(We need to use a different image for every single contract type. e.g. `CALL` and `CALLE` should be different)

OskarAhl: Yeah that is a good idea, could we perhaps use a sprite sheet for all the icons for binary static?

src/javascript/app_2/pages/portfolio/portfolio.jsx
```diff
@@ -0,0 +1,272 @@
+import React from 'react';
+import moment from 'moment';
+import DAO from '../../data/dao';
+import DataTable from '../../components/elements/data_table.jsx';
+import Client from '../../../app/base/client';
```
binary-hamedanchi: In order to keep them separated, please do not import anything from `app` into `app_2`. It's fine to import from `_common` folder.

src/javascript/app_2/components/elements/data_table.jsx
```diff
@@ -86,7 +86,8 @@ class DataTable extends React.PureComponent {
 DataTable.propTypes = {
     columns         : PropTypes.array,
     data_source     : PropTypes.array,
-    footer          : PropTypes.string,
+    // footer can be object || string
+    footer          : PropTypes.any,
```
negarn: can we put the comment on the same line 90, so it doesn't break up the object?

src/javascript/app_2/pages/portfolio/portfolio.jsx
```diff
@@ -0,0 +1,315 @@
+import React                from 'react';
+import moment               from 'moment';
+import PortfolioCard        from './portfolio_card.jsx';
+import DAO                  from '../../data/dao';
```
negarn: d is after c

src/javascript/app_2/pages/portfolio/portfolio.jsx
```diff
@@ -0,0 +1,315 @@
...
+import { getAppId }         from '../../../config';
+import ClientBase           from '../../../_common/base/client_base';
+import { formatMoney }      from '../../../_common/base/currency_base';
+import { localize }         from '../../../_common/localize';
+import { getPropertyValue } from '../../../_common/utility';
+import Loading              from '../../../../templates/_common/components/loading.jsx';
+
+const formatPortfolioData = (portfolio_arr) => {
+    const formatted_portfolio = portfolio_arr.map((portfolio_item) => {
+        const date_obj        = new Date(portfolio_item.expiry_time* 1000);
```
negarn: could we use moment for both? why use date in one and moment for other?

src/javascript/app_2/pages/portfolio/portfolio.jsx
```diff
@@ -0,0 +1,315 @@
...
+    EXPIRYMISS  : localize('Ends Outside'),
+    EXPIRYRANGE : localize('Ends Between'),
+    EXPIRYRANGEE: localize('Ends Between'),
+    LBFLOATCALL : localize('Close-Low'),
+    LBFLOATPUT  : localize('High-Close'),
+    LBHIGHLOW   : localize('High-Low'),
+    RANGE       : localize('Stays Between'),
+    UPORDOWN    : localize('Goes Outside'),
+    ONETOUCH    : localize('Touches'),
+    NOTOUCH     : localize('Does Not Touch'),
```
negarn: is it best to localize each separately here? or localize them at the end after fetching them? up to you, I'm not sure either

negarn: but I'm afraid localizing them like this won't work, it might get called before localize is populated

OskarAhl: Oh yeah that is correct, localizing them there won't work. Will update it 👍

src/javascript/app_2/components/elements/data_table.jsx
```diff
@@ -86,7 +86,7 @@ class DataTable extends React.PureComponent {
 DataTable.propTypes = {
     columns         : PropTypes.array,
     data_source     : PropTypes.array,
-    footer          : PropTypes.string,
+    footer          : PropTypes.any, // footer can be object || string
```
jbaek7023: I think footer is passing `DOM` as a footer. The purpose of writing PropTypes is to restrict the prop type so that we can further make the testing and verification easy. 

Solution
1) Try use `console.log(footer)` and refactor `footer` prop as much as you can
2) If there is no other way to change `footer` prop, restrict the `footer` prop to two types. `footer: PropTypes.oneOfTypes([PropTypes.object, PropTypes.string])`.

Thanks!

src/javascript/app_2/components/elements/tooltip.jsx
```diff
@@ -18,7 +18,7 @@ const Tooltip = ({
 
 Tooltip.propTypes = {
     alignment: PropTypes.string,
-    children : PropTypes.string,
+    children : PropTypes.any, // string || number
```
jbaek7023: Same goes to here.

src/javascript/app_2/pages/portfolio/portfolio.jsx
```diff
@@ -0,0 +1,315 @@
...
+
+class Portfolio extends React.PureComponent  {
+    state = {
+        columns: [
+            {
+                title     : localize('Reference No.'),
+                data_index: 'reference',
+                renderCell: (data, data_index) => {
+                    const tooltip =
+                        (data.app_id !== app_id) && this.state.oauth_apps && this.state.oauth_apps[data.app_id];
```
jbaek7023: `const { oauth_apps } = this.state` for code readability. :)

src/javascript/app_2/pages/portfolio/portfolio.jsx
```diff
@@ -0,0 +1,315 @@
...
+        obj_oauth_apps[app.app_id] = app.name;
+    });
+    return obj_oauth_apps;
+};
+
+const app_id = getAppId();
+
+class Portfolio extends React.PureComponent  {
+    state = {
+        columns: [
```
jbaek7023: You can explicitly define `columns` here. I don't see any reason to put `columns` data in `state`. 

The purpose of having data in `state` is for Component class to render / re-render. 

Solution:
1) Define `columns` array after or before 
const  columns  = [...];
2) call columns rather than `this.state.columns`.

OskarAhl: The columns are being rendered / re-rendered based on state (from the streaming calls ```subscribeTransaction``` and ```subscribeProposalOpenContract```), so I think they should be in state? :)

jbaek7023: Because you never `setState` which means, this `columns` prop is not directly related to this Component. So I thought we need to pull out the columns `variable` from `state =`. 

I gave you a wrong suggestion and here is a new suggestion.
Solution

0) Define `columns` array to as a constant.
1) In DataTable, we can put `componentWillReceiveProps` or `shouldComponentUpdate` to listen to props change. 

Major reason:
1) Having multiple `state` will slow down the PageSpeed dramatically. Why do we have to render `Portfolio` rather than `DataTable` component?

- Let's have some talks in our Lunch time :)

OskarAhl: Oh no never mind first comment is correct! We can move columns outside of state and keep it as a normal property of Portfolio since columns not rendered / updated over time. The property that handles rendering of the table is data_source

src/sass/app_2/_common/base/constants.scss
```diff
@@ -10,9 +10,11 @@ $COLOR_LIGHT_RED  : #f44336;
 $COLOR_BLUE       : #2a3052;
 $COLOR_DARK_BLUE  : #15212d;
 $COLOR_LIGHT_BLUE : #1565c0;
+$COLOR_TEXT_BLUE  : #1565c0;
```
binary-hamedanchi: Can we use `$COLOR_LIGHT_BLUE` as they share the same value?

src/sass/app_2.scss
```diff
@@ -37,5 +37,6 @@
 @import 'app_2/_common/mobile_widget';
 
 // Pages
+@import 'app_2/portfolio';
```
binary-hamedanchi: Probably moving to `pages` folder like statement and trading. :)

src/sass/app_2/_common/components/table.scss
```diff
@@ -0,0 +1,66 @@
+$side-padding: 2rem;
```
binary-hamedanchi: Seems that the same file already exists with the name [data_table.scss](https://github.com/binary-com/binary-static/blob/new-app/src/sass/app_2/_common/components/data_table.scss)?

src/javascript/app_2/data/ws_methods.js
```diff
@@ -15,6 +15,15 @@ const WS = (() => {
     const getAccountStatus = () =>
         BinarySocket.send({ get_account_status: 1 });
 
+    const getOauthApps = () =>
```
binary-hamedanchi: To be consistent, could you please remove extra `get` from functions name?
(Function names are now similar to the corresponding API method name. i.e. `oauthApps`, `portfolio` except the streaming ones which have `subscribe` prefix)

PR #3773, opened 2018-04-25
negar/remove_some_limits

negarn:
remove some limits


ashkanx:
@binary-hamedanchi review required

PR #3789, opened 2018-04-30
kelly/enable_ladders

kellybinary:


src/javascript/app/base/menu.js
```diff
@@ -1,13 +1,16 @@
...
 const findParent         = require('../../_common/utility').findParent;
+
 require('../../_common/lib/mmenu/jquery.mmenu.min.all.js');
 
 const Menu = (() => {
     const init = () => {
         const menu_top = getElementById('menu-top');
 
         applyToAllElements('li', (el) => { el.classList.remove('active', 'active-parent'); }, '', menu_top);
+        applyToAllElements('.cr-only', (el) => { el.setVisibility(Client.hasCostaricaAccount()); });
```
negarn: don't we also want to show it to virtual clients who can upgrade to CR?

PR #3793, opened 2018-04-30
zaki/add_economic_calendar

zakibinary:
Changelog
---
- Added economic calendar page.


src/templates/app/_layout/header.jsx
```diff
@@ -122,10 +122,11 @@ const Header = () => (
...
                                             subitems : [
-                                                { text: it.L('Asset Index'),   href: it.url_for('resources/asset_indexws'), className: 'ja-hide' },
-                                                { text: it.L('Trading Times'), href: it.url_for('resources/market_timesws') },
-                                                { text: it.L('TradingView'),   href: 'https://tradingview.binary.com',      className: 'ja-hide', target: '_blank' },
-                                                { text: it.L('SmartCharts'),   href: 'https://charts.binary.com/',          className: 'ja-hide', target: '_blank' },
+                                                { text: it.L('Asset Index'),       href: it.url_for('resources/asset_indexws'), className: 'ja-hide' },
+                                                { text: it.L('Trading Times'),     href: it.url_for('resources/market_timesws') },
+                                                { text: it.L('Economic Calendar'), href: it.url_for('resources/economic_calendar') },
```
negarn: do we need to hide it for japan?

scripts/pages.js
```diff
@@ -30,6 +30,8 @@ module.exports = [
     ['resources',                          'app/resources/index',                       'default',    'Resources'],
     ['resources/asset_indexws',            'app/resources/asset_index',                 'default',    'Asset Index', 'ja'],
     ['resources/market_timesws',           'app/resources/trading_times',               'default',    'Trading Times'],
+    ['resources/economic_calendar',        'app/resources/economic_calendar',           'default',    'Economic Calendar'],
```
negarn: do we need to exclude it for japan?

src/templates/app/resources/economic_calendar.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+const EconomicCalendar = () => (
+    <div id='economic_calendar' className='static_full'>
+        <div className='economic-calendar-header gr-row'>
+            <div className='gr-12 gr-centered'>
+                <h1>{it.L('Economic Calendar')}</h1>
+                <p>{it.L('Check out the latest economic events and indicators from all over the world. All data are collected and updated in real time.')}</p>
+            </div>
+        </div>
+        <div className='gr-12 gr-centered'>
```
negarn: why do we need the extra padding here and on line 16? it makes it smaller than the first paragraph

src/templates/app/resources/economic_calendar.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+const EconomicCalendar = () => (
+    <div id='economic_calendar' className='static_full'>
+        <div className='economic-calendar-header gr-row'>
+            <div className='gr-12 gr-centered'>
```
negarn: why add `gr-12` here? it negates the `gr-row` on top. we can remove both `gr-row` and `gr-12`, and even `gr-centered` as I don't see the point

scripts/pages.js
zakibinary: @ashkanx , do we need to? Since Trading Times isn't excluded I assume it's the same with Economic Calendar.

src/javascript/app/pages/resources/economic_calendar/economic_calendar.js
```diff
@@ -0,0 +1,14 @@
+const EconomicCalendar = (() => {
+
+    const onLoad = () => {
+        $.getScript( 'https://c.mql5.com/js/widgets/calendar/widget.v3.js' )
+          .done(() => {
+              new economicCalendar({ width: '100%', height: '500px', mode: 2 }); // eslint-disable-line new-cap, no-new, no-undef
```
binary-hamedanchi: Is it possible to set language as well?

zakibinary: It is possible, but they don't support some of the languages that we do.

ashkanx:
@zakibinary conflicts

PR #3808, opened 2018-05-07
remove_japan

negarn:
remove japan from languages


src/javascript/app/pages/trade/charts/webtrader_chart.js
```diff
@@ -78,7 +77,7 @@ const WebtraderChart = (() => {
             timePeriod        : getChartSettings().time_frame,
             type              : getChartSettings().chart_type,
             lang              : getLanguage().toLowerCase(),
-            timezoneOffset    : (isJPClient() ? -9 : 0) * 60,
+            timezoneOffset    : 0,
```
negarn: do we need to send it at all, if it's zero?

PR #3817, opened 2018-05-09
craig/tandc_handling

CraigLawrence:


ashkanx:
@CraigLawrence why is this changing? required by compliance team?

CraigLawrence:
@ashkanx We're improving the way that T&C versioning is handled in the backend.
We'll be able to notify clients of upcoming changes in advance, and have them accept the new T&Cs before they take effect, so they won't get blocked from trading.
This means the date on that page could be in the future, so we just have to tweak the language to make sense in that scenario.
We're also dropping the version number and just using a date now.

src/templates/static/legal/tac_sections/tac.jsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 const Tac = () => (
     <div>
         <h2 data-anchor>{it.L('Terms and Conditions')}</h2>
-        <p><i>{it.L('Version 44. Last modified 2018-04-10.')}</i></p>
+        <p><i>{it.L('Date effective 2018-04-10.')}</i></p>
```
negarn: how about the wording here? https://github.com/binary-com/binary-static/blob/master/src/templates/static/legal/metatrader/terms_and_conditions.jsx#L6 does that also need to be updated?

CraigLawrence: There's been no change to MT5, but we can change it if you want the language consistent.

ashkanx:
Thank you @CraigLawrence

PR #3847, opened 2018-05-15
oskar/gdpr_opt_in_out

OskarAhl:
Show email consent checkbox & update consent text:
 1. For virtual account signup
 2. Under settings for all account types

updates: 
- Email consent **only** under settings


src/templates/app/new_account/virtual.jsx
```diff
@@ -16,18 +16,26 @@ const Virtual = () => (
...
 
+            <Fieldset>
                 <FormRow
                     type='checkbox'
-                    checked
                     id='email_consent'
-                    row_class='invisible'
                     label_row_id='email_consent_label'
-                    label={it.L('Receive news and special offers')}
+                    label={it.L('Yes, I would like to receive emails on [_1] products, services, and events. [_2] I understand that I can unsubscribe from these emails at any time from my profile settings.', it.website_name, '\n')}
```
negarn: do we need the space after dot? there's probably no way translators will keep that space. if we really need it, I suggest adding it inside the placeholder as well

OskarAhl: Since it's a new line we can remove the space after the dot I think

PR #3857, opened 2018-05-17
ruskakimov/tick_chart_improvement

ruskakimov:


src/javascript/app/pages/trade/tick_trade.js
```diff
@@ -3,6 +3,7 @@ const requireHighstock     = require('./common').requireHighstock;
 const Tick                 = require('./tick');
 const ViewPopupUI          = require('../user/view_popup/view_popup.ui');
 const BinarySocket         = require('../../base/socket');
+const Currency             = require('../../../_common/base/currency_base');
```
negarn: why not just import addComma?

ruskakimov: @negarn will do 👍

src/javascript/app/pages/trade/tick_trade.js
```diff
@@ -361,26 +362,33 @@ const TickDisplay = (() => {
         }
     };
 
+    const addSellSpot = (contract) => {
+        // for tick trades only touchnotouch can have sell spot before exit spot
+        if (contract_category !== 'touchnotouch' || !applicable_ticks) return;
```
negarn: can we find a better way than hardcoding the touchnotouch category? does it break anything if we just remove the contract_category check?

ruskakimov: @negarn  it may override exit lines for other contracts, but if we only add sell spot if no line was there before, then it should work.

src/javascript/app/pages/trade/tick_trade.js
```diff
@@ -1,10 +1,11 @@
 const moment               = require('moment');
-const requireHighstock     = require('./common').requireHighstock;
+const { requireHighstock } = require('./common');
 const Tick                 = require('./tick');
 const ViewPopupUI          = require('../user/view_popup/view_popup.ui');
 const BinarySocket         = require('../../base/socket');
+const { addComma }         = require('../../../_common/base/currency_base');
 const CommonFunctions      = require('../../../_common/common_functions');
-const localize             = require('../../../_common/localize').localize;
+const { localize }         = require('../../../_common/localize');
```
negarn: we don't really import it this way, can you use the `.` as before?

PR #3859, opened 2018-05-17
zaki/fix_dropdown_for_stateless_countries

zakibinary:
Changelog
- Fixed error when initializing custom dropdown with stateless countries.


src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -356,6 +356,11 @@ const PersonalDetails = (() => {
                     states.forEach((state) => {
                         $field.append($('<option/>', { value: state.value, text: state.text }));
                     });
+                    $('#address_state').select2({
```
negarn: do we need to do https://github.com/binary-com/binary-static/pull/3859/files#diff-e45a23f7310d6d033d0ba4ff74b1cfa5R368 before select2?

zakibinary: I just double checked, `select2` properly initializes for both countries with and without states, so I think it's not needed before select2.

PR #3876, opened 2018-05-21
kelly/fix_dob_not_showing_in_android

kellybinary:


src/javascript/app/common/form_manager.js
```diff
@@ -54,8 +54,6 @@ const FormManager = (() => {
                     value = val || '';
                     if (field.value) {
                         value = typeof field.value === 'function' ? field.value() : field.value;
-                    } else if (field.$.attr('data-picker') === 'native') {
```
negarn: won't removing this break it for other pages that are using native date picker? e.g. trade page, self exclusion, etc

kellybinary: was there any reason we're not getting value from `data-value` attribute for datepicker?

negarn: I don't remember, but it's good if we can test all scenarios if we're removing it.

negarn: I have added a checklist to the card, will approve it for now

kellybinary: ok, hopefully, nothing breaks 🙏 😄

PR #3888, opened 2018-05-24
kelly/update_btc_min_withdrawal

kellybinary:


src/templates/app/cashier/payment_methods.jsx
```diff
@@ -193,7 +193,7 @@ const PaymentMethods = () => {
                                     { text: <PaymentLogo logo='bitcoin' /> },
                                     { text: 'BTC' },
                                     { text: '0.002' },
-                                    { text: '0.004' },
+                                    { text: '0.003' },
```
negarn: missing one more zero, I believe

kellybinary: oh yeah..

PR #3889, opened 2018-05-24
ruskakimov/tick_trade_app_button

ruskakimov:


src/sass/static/home.scss
```diff
@@ -194,9 +194,6 @@
             order: 2;
             padding-top: 30px;
         }
-        .center-text-m {
-            text-align: center;
```
negarn: this class is only for home page, any reason why we are removing it?

src/sass/_common/base/media-queries.scss
```diff
@@ -19,6 +19,9 @@
 }
 
 @media screen and (max-width: 768px) {
+    .center-text-m {
```
negarn: this class is used throughout the code with the assumption that it will only be centered in mobile screen size, can you instead create a new class `center-text-p` so we don't break others?

PR #3892, opened 2018-05-24
kelly/fixes

kellybinary:


src/javascript/app/base/client.js
```diff
@@ -108,6 +109,10 @@ const Client = (() => {
         });
     };
 
+    const hideUpgradePulser = () => {
+        $('#topbar-msg').children('a').setVisibility(!/maltainvest/.test(window.location.href));
```
negarn: this only works for financial account opening page. I think a more general solution would be better. I suggest on each account opening page we hide the pulser for that page individually, then we don't have to worry about url paths changing either

src/templates/app/resources/economic_calendar.jsx
```diff
@@ -4,14 +4,8 @@ import SeparatorLine from '../../_common/components/separator_line.jsx';
...
     <div id='economic_calendar' className='static_full'>
-        <div className='economic-calendar-header'>
-            <h1>{it.L('Economic Calendar')}</h1>
-            <p>{it.L('Check out the latest economic events and indicators from all over the world. All data are collected and updated in real time.')}</p>
-            <p>{it.L('You can monitor macroeconomic indicators straight from the MetaTrader 5 platform using release time labels displayed on [_1]financial symbol charts[_2].', '<a href="https://www.metatrader5.com/en/terminal/help/fundamental#chart" target="_blank" rel="noopener">','</a>')}</p>
-            <p>{it.L('With the Economic Calendar, you can see a breakdown of all the scheduled economic events due to take place on any given day.')}</p>
-            <p>{it.L('Click on an individual event to be presented with further information and links to more in-depth data about it. The information provided by the Economic  Calendar will help you make more informed trading decisions.')}</p>
-        </div>
         <div className='gr-padding-10'>
+            <h1>{it.L('Economic Calendar')}</h1>
```
negarn: if we put the header inside gr-padding-10 it will have more distance from the top of the page than our pages normally do.

src/javascript/_common/common_functions.js
```diff
@@ -97,6 +97,7 @@ const getElementById = (id_selector, parent = document) => parent.getElementById
 const getVisibleElement = (class_name, parent = document) =>
     Array.from(parent.getElementsByClassName(class_name)).find((el) => isVisible(el));
 
+
```
negarn: extra line :)

src/javascript/_common/utility.js
```diff
@@ -179,6 +179,15 @@ const getStaticHash = () => {
     return static_hash;
 };
 
+const upgradePulser = (() => {
```
binary-hamedanchi: The idea of `_common` folder is to serve shared logics between different sections. Therefore, if a function is only related to a particular section, better to move it there. This also prevents having extra unused code in another section.

Besides, having this here will add jQuery dependency to `app_2`, which is not the plan.

kellybinary: ok, noted. since this function is mostly related to new account opening, would move it to `common/account_opening`

PR #3898, opened 2018-05-25
zaki/contracts_dropdown_mobile_fix

zakibinary:
Changelog
- Fix for clickable area for contracts dropdown on mobile screens.


src/javascript/app/pages/trade/contracts.jsx
```diff
@@ -124,7 +124,7 @@ class Contracts extends React.Component {
                     className={`contracts_dropdown ${open ? '' : 'hidden'}`}
                     ref={this.saveRef.bind(null, 'wrapper')}
                 >
-                    <div className='mobile_close invisible'>
+                    <div className={`mobile_close invisible ${open && is_mobile ? '': 'disabled'}`}>
```
ruskakimov: invisible class never takes effect because of the `!important` [here](https://github.com/zakibinary/binary-static/blob/97e7cfe07891bddc1a47231163a6def1731e4590/src/sass/app/components/contract.scss#L59)
if we remove it, then this can just be:
`<div className={`mobile_close ${open ? '': 'invisible'}`}>`

src/javascript/app/pages/trade/contracts.jsx
```diff
@@ -107,7 +107,7 @@ class Contracts extends React.Component {
             open,
             formname,
         } = this.state;
-
+        const is_mobile = window.innerWidth <= 767;
```
ruskakimov: then we don't need this

PR #3903, opened 2018-05-25
update text

negarn:


scripts/generate-static-data.js
```diff
@@ -253,7 +253,7 @@ const texts = [
     'Step',
 
     // Guide -> trading page
-    'Select your market',
+    'Select your market and underlying asset',
     'Select your underlying asset',
```
binary-hamedanchi: This seems not in use anymore.

PR #3908, opened 2018-05-28
kelly/encourage_clients_to_extend_self_exclusion

kellybinary:
show notice to encourage MX/MLT clients' to extend their self-exclusion to other gambling operators, as per UKGC requirement


src/templates/app/user/security/self_exclusion.jsx
```diff
@@ -78,6 +78,16 @@ const SelfExclusion = () => (
                 </div>
 
                 <SubmitButton text={it.L('Update Settings')} type='submit' />
+
+                <div id='ukgc_requirement_notice' data-show='iom, malta' className='gr-row invisible'>
+                    <div className='gr-8 gr-push-4 gr-12-m gr-push-0-m'>
+                        <div className='notice-msg gr-child gr-parent'>
+                            <p>{it.L('Self-exclusion on this website only applies to your [_1] account and does not include other companies or websites.', it.website_name)}</p>
+                            <p>{it.L('Since you feel the need to self-exclude, we strongly encourage you to exclude yourself from similar services that you may have signed up for.')}</p>
+                            <p>{it.L('Thank you for trading with [_1].', it.website_name)}</p>
```
jy-binary: remove the "Thank you for trading with Binary.com" line. The client is clearly unhappy with the service as he is self-excluding, so it's a bit tone-deaf to be saying thank you for trading..

kellybinary: sure, will do

src/templates/_common/components/forms.jsx
```diff
@@ -134,6 +131,9 @@ export const SubmitButton = ({
                     {text}
                 </button>
             </div>
+            { !no_error &&
+                <p id={msg_id || 'msg_form'} className={`${is_centered ? 'center-text' : ''} error-msg no-margin invisible`} />
```
negarn: this will cause the change for all our form errors, are we sure about it?

kellybinary: yes, but @ashkanx could you please confirm?

src/javascript/app/pages/user/account/settings/self_exclusion.js
```diff
@@ -268,7 +270,10 @@ const SelfExclusion = (() => {
             return;
         }
         showFormMessage('Your changes have been updated.', true);
-        $('#ukgc_requirement_notice').setVisibility($('#exclude_until').attr('data-value'));
+        if ($('#exclude_until').attr('data-value')) {
+            $('#ukgc_requirement_notice').setVisibility(1);
+            $('#gamstop_info_bottom').setVisibility(0);
```
negarn: can we hide this before we show the paragraph? then it will be hidden from the start instead of showing for a short time before hiding. (e.g. change lines 274 and 275)

src/javascript/app/pages/user/account/settings/self_exclusion.js
```diff
@@ -89,6 +89,8 @@ const SelfExclusion = (() => {
                     if (key === 'exclude_until') {
                         setDateTimePicker(exclude_until_id, value);
                         $form.find('label[for="exclude_until"]').text('Excluded from the website until');
+                        $('#ukgc_requirement_notice').setVisibility(1);
+                        $('#ukgc_gamstop').setVisibility(is_gamstop_client);
```
negarn: same here, can we decide visibility of the sentence before we make the paragraph visible?

PR #3914, opened 2018-05-28
zaki/acc_transfer_exchange_rates

zakibinary:
Changelog
- Added exchange rates to account transfers page.
- Updated design and layout for account transfers page.


src/javascript/app/pages/cashier/account_transfer.js
```diff
@@ -145,11 +147,61 @@ const AccountTransfer = (() => {
...
+        const input_val = parseFloat(e.target.value);
+        calculateAmount(input_val);
+    };
+
+    const calculateAmount = (value) => {
+        const el_amount    = getElementById('amount');
+        const el_amount_to = getElementById('amount_to');
+        const exchange_rate = getPropertyValue(exchange_rates, curr_account_to);
+        const transferred_amount = (value * exchange_rate).toFixed(8);
+        if (transferred_amount && !isNaN(transferred_amount)) {
```
negarn: I think we can just do `!isNaN(transferred_amount)`, `!isNaN(undefined)` returns `false`, so should be fine?

negarn: and if it's undefined we have bigger issues, because on the line above `undefined.toFixed(8)` will definitely throw an error

src/javascript/app/pages/cashier/account_transfer.js
```diff
@@ -145,11 +147,61 @@ const AccountTransfer = (() => {
...
+    const onAmountInput = (e) => {
+        const input_val = parseFloat(e.target.value);
+        calculateAmount(input_val);
+    };
+
+    const calculateAmount = (value) => {
+        const el_amount    = getElementById('amount');
+        const el_amount_to = getElementById('amount_to');
+        const exchange_rate = getPropertyValue(exchange_rates, curr_account_to);
+        const transferred_amount = (value * exchange_rate).toFixed(8);
```
negarn: why is it hardcoded to 8 decimal places? can we get the decimal places from currency config?

src/javascript/app/pages/cashier/account_transfer.js
```diff
@@ -145,11 +147,61 @@ const AccountTransfer = (() => {
...
+        if (transferred_amount && !isNaN(transferred_amount)) {
+            el_amount_to.value = transferred_amount;
+        }
+        else {
+            el_amount_to.value = '';
+        }
+        const transfer_fee_value   = (parseFloat(el_amount.value) * 0.01);
+        const transfer_total_value = (parseFloat(el_amount.value) * 1.01);
+
+        elementTextContent(getElementById('transfer_fee_lbl'), `${client_currency} ${isCryptocurrency(client_currency) ? transfer_fee_value.toFixed(8) : transfer_fee_value.toFixed(2)}`);
```
negarn: the decimal places shouldn't be hardcoded. you can get them from `Currency.getDecimalPlaces` function

negarn: it will break for things like DAI which is crypto but only has 2 decimal places. better to get it from API.

src/javascript/app/pages/cashier/account_transfer.js
```diff
@@ -145,11 +153,72 @@ const AccountTransfer = (() => {
...
+
+    const getExchangeRates = () => getPropertyValue(exchange_rates, curr_account_to);
+
+    const updateExchangeMessage = () => {
+        getElementById('exchange_rate').textContent = `${localize('Exchange rate')}: ${client_currency} 1 = ${getExchangeRates()} ${curr_account_to}`;
+    };
+
+    const onAccountToChange = (e) => {
+        const el_amount = getElementById('amount');
+        curr_account_to = (e.target.value || e.target.getAttribute('data-value') || '').split(' (')[1].replace(')','');
```
binary-hamedanchi: Preferably, using RegExp to extract the value.

src/javascript/app/pages/cashier/account_transfer.js
```diff
@@ -145,11 +153,75 @@ const AccountTransfer = (() => {
...
+    const onAmountInput = (e) => {
+        const input_val = parseFloat(e.target.value);
+        calculateAmount(input_val);
+    };
+
+    const calculateAmount = (value) => {
+        const el_amount    = getElementById('amount');
+        const el_amount_to = getElementById('amount_to');
+
+        const exchange_rate  = getPropertyValue(exchange_rates, curr_account_to);
```
negarn: can use `getExchangeRates()` instead

src/sass/app/websocket_pages.scss
```diff
@@ -110,6 +110,10 @@
             height: 25px;
         }
     }
+    ul.hint-info {
+        margin-top: 0;
+        margin-bottom: 0;
```
negarn: can we use the `no-margin` class?

zakibinary: We only want to remove the `margin-top` and `margin-bottom` here and keep the left and right margin.

src/templates/app/cashier/account_transfer.jsx
```diff
@@ -72,9 +72,8 @@ const AccountTransfer = () => (
             </Fieldset>
             <Fieldset legend={it.L('To')}>
                 <FormRow label={it.L('Transfer to')}   type='select' id='transfer_to' />
-                <FormRow label={it.L('Amount')}        type='custom' id='transfer_amount'>
-                    <span id='amount_to' />
-                    <div className='hint' id='exchange_rate' />
+                <FormRow label={it.L('Exchange Rate')}        type='custom' id='transfer_amount'>
```
negarn: do we need all this space between label and type?

PR #3920, opened 2018-05-30
oskar/dai_improvements

OskarAhl:
* Payment agent withdrawal: change minimum and max input validation for DAI
* Set currency: update grid layout


src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -54,8 +54,8 @@ const PaymentAgentWithdraw = (() => {
 
             const currency = Client.get('currency');
             const form_id  = `#${$(view_ids.form).find('form').attr('id')}`;
-            const min      = isCryptocurrency(currency) ? 0.002 : 10;
-            const max      = isCryptocurrency(currency) ? 5 : 2000;
+            const min      = isCryptocurrency(currency) && currency !== 'DAI' ? 0.002 : 10;
+            const max      = isCryptocurrency(currency) && currency !== 'DAI' ? 5 : 2000;
```
negarn: can you think of a better way than this?

OskarAhl: I didn't want to update the isCryptocurrency() function since it is used in so many places throughout the application and DAI is a cryptocurrency, I could write a new function for setting the min and max that takes currency as input?

binary-hamedanchi: `getDecimalPlaces()` makes exactly the same categorization needed here, but not a good idea as it's not explicitly related.

Probably trying at least to not repeating the same condition? :)
I'd also suggest to put these kind of constatnt values at the top of every module and hopefully later refactor all to a single config file.

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -52,14 +53,15 @@ const PaymentAgentWithdraw = (() => {
             }
             setActiveView(view_ids.form);
 
-            const currency = Client.get('currency');
-            const form_id  = `#${$(view_ids.form).find('form').attr('id')}`;
-            const min      = isCryptocurrency(currency) ? 0.002 : 10;
-            const max      = isCryptocurrency(currency) ? 5 : 2000;
+            const form_id           = `#${$(view_ids.form).find('form').attr('id')}`;
+            const is_crypto_not_dai = isCryptocurrency(currency) && currency !== 'DAI';
+            const min_withdrawal    = is_crypto_not_dai ? 0.002 : 10;
```
negarn: what if we add a min_pa_withdrawl and max_pa_withdrawal to crypto_config in currency_base.js and add a function to get min pa withdrawal and max pa withdrawal? at least all the config will be in one place, and we won't need exceptions for DAI

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -1,11 +1,11 @@
...
-const getDecimalPlaces = require('../../common/currency').getDecimalPlaces;
-const isCryptocurrency = require('../../common/currency').isCryptocurrency;
-const FormManager      = require('../../common/form_manager');
-const validEmailToken  = require('../../common/form_validation').validEmailToken;
-const localize         = require('../../../_common/localize').localize;
-const getHashValue     = require('../../../_common/url').getHashValue;
+const Client             = require('../../base/client');
+const BinarySocket       = require('../../base/socket');
+const getDecimalPlaces   = require('../../common/currency').getDecimalPlaces;
+const getPaWithdrawalLimit = require('../../common/currency').getPaWithdrawalLimit;
```
zakibinary: Add space for the rest to align with here?

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -52,14 +53,14 @@ const PaymentAgentWithdraw = (() => {
             }
             setActiveView(view_ids.form);
 
-            const currency = Client.get('currency');
-            const form_id  = `#${$(view_ids.form).find('form').attr('id')}`;
-            const min      = isCryptocurrency(currency) ? 0.002 : 10;
-            const max      = isCryptocurrency(currency) ? 5 : 2000;
+            const form_id        = `#${$(view_ids.form).find('form').attr('id')}`;
+            const min_withdrawal = getPaWithdrawalLimit(currency, 'max');
```
negarn: shouldn't this be 'min' and the below 'max'?

PR #3925, opened 2018-06-01
[new-app] Write PropTypes for each components 

jbaek7023:
PropTypes makes the code more readable and testable to developers. 
Further prop types validation will be implemented. 

We can't further restrict the prop types more without refactoring.

To-do:
- Restrict remaining prop types after refactoring dropdown. 
    - IMH, Dropdown.jsx is complicate because we're over-using `mobx-store`. -> Better to use local states rather than having many states in store.


src/javascript/app_2/pages/trading/actions/helpers/contract_type.js
```diff
@@ -242,7 +242,7 @@ const ContractType = (() => {
         const start_dates_list = [];
 
         if (config.has_spot) {
-            start_dates_list.push({ text: localize('Now'), value: 'now' });
+            start_dates_list.push({ text: localize('Now'), value: 999 });
```
jbaek7023: Please give me any reasonable or efficient number which can represent `Now` **if you have any**. 
- We have to keep the prop type of 'start_date' as the type of `Number` only.

PR #3927, opened 2018-06-01
negar/check_pob

negarn:
show message and all fields for new account error


src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -425,8 +399,13 @@ const PersonalDetails = (() => {
         });
     };
 
+    const onUnload = () => {
+        is_for_new_account = false;
+    };
+
```
OskarAhl: Why are you assigning is_for_new_account as false on unload?

negarn: this is used to redirect back to accounts page, so once you get redirected or leave this page willingly it shouldn't assume that you want to be redirected back to accounts page if you come back again

PR #3928, opened 2018-06-01
Gokula/mb min max validation

gokula-krishna-dev:


src/javascript/app/pages/mb_trade/mb_event.js
```diff
@@ -106,20 +108,19 @@ const MBTradingEvents = (() => {
...
             let error_msg = '';
 
             if (!payout_amount || isNaN(payout_amount)) {
                 is_valid  = false;
                 error_msg = localize('Should be a valid number.');
-            } else if (+payout_amount < min_amount || +payout_amount > max_client_amount) {
+            } else if (+payout_amount > max_client_amount) {
                 is_valid  = false;
-                error_msg = localize('Should be between [_1] and [_2]', [min_amount, max_client_amount]);
+                error_msg = localize('Should be below [_1]', [max_client_amount]);
```
negarn: can we use `Should be less than [_1]` since we already have translation string for it

PR #3932, opened 2018-06-04
ruskakimov/call_put_spread

ruskakimov:


src/javascript/app/pages/trade/charts/highchart.ui.js
```diff
@@ -1,5 +1,5 @@
-const addComma = require('../../../common/currency').addComma;
-const localize = require('../../../../_common/localize').localize;
+const addComma      = require('../../../common/currency').addComma;
+const localize      = require('../../../../_common/localize').localize;
```
negarn: extra whitespace

src/javascript/app/pages/trade/tick_trade.js
```diff
@@ -128,6 +128,8 @@ const TickDisplay = (() => {
                 width          : config.width || (config.minimize ? 394 : null),
                 height         : config.minimize ? 143 : null,
                 backgroundColor: null,
+                // TODO: we pass UNDEFINED as load event callback,
+                // plot() is executed once
```
negarn: what is this comment?

ruskakimov: @negarn on this line
```js
events         : { load: plot(config.plot_from, config.plot_to) },
```
we pass the result of `plot(config.plot_from, config.plot_to)`, which is `undefined` in this case,
this is probably not intentional and what was meant is something like this:
```js
events         : { load: () => {plot(config.plot_from, config.plot_to)} },
```

negarn: I see, then could we fix it here? or it breaks something?

ruskakimov:
some github issues, closing and re-opening a new one

PR #3937, opened 2018-06-04
[new-app] Write PropTypes for each components

jbaek7023:
PropTypes makes the code more readable and testable to developers. 
Further prop types validation will be implemented. 

We can't further restrict the prop types more without refactoring.

To-do:
- Restrict remaining prop types after refactoring dropdown. 
    - IMH, Dropdown.jsx is complicate because we're over-using `mobx-store`. -> Better to use local states rather than having many states in store.


jbaek7023:
I removed oneOfTypes() prop types as many as possible. The rest can be removed with refactoring.

PR #3944, opened 2018-06-05
negar/callpute_only_show_if_dur

negarn:
only show checkbox if callputequals is allowed for current duration


src/javascript/app/pages/trade/process.js
```diff
@@ -217,9 +219,22 @@ const Process = (() => {
...
-    const displayEquals = (formname = Defaults.get('formname'), contracts = State.getResponse('contracts_for.available')) => {
-        const el_equals = getElementById('callputequal');
-        if (/^(callputequal|risefall)$/.test(formname) && (contracts || []).find(contract => contract.contract_category === 'callputequal')) {
+    const duration_type = {
+        t: 'tick',
+        s: 'intraday',
+        m: 'intraday',
+        h: 'intraday',
+        d: 'daily',
+    };
```
OskarAhl: Could duration_type be moved to commonTrading in the common.js file?

src/javascript/app/pages/trade/common.js
```diff
@@ -294,6 +294,14 @@ const commonTrading = (() => {
         d: 5,
     };
 
+    const duration_type = {
```
binary-hamedanchi: Can we merge this with duration_order?

src/javascript/app/pages/trade/contract.js
```diff
@@ -75,20 +75,24 @@ const Contract = (() => {
 
         const form_barrier = getFormNameBarrierCategory(form_name);
         form               = form_barrier.form_name;
-        barrier            = barrier_category = form_barrier.barrier_category;
+        if (!form) {
+            return;
+        }
+        barrier = barrier_category = form_barrier.barrier_category;
```
binary-hamedanchi: Does anything break by removing the `barrier_category` and use `barrier` instead?

negarn: no, I think in the past it was just used to show we are comparing category with category in the loop below.

PR #3972, opened 2018-06-12
negar/hide_livechat_if_unavailable

negarn:
only show live chat if during working hours


src/javascript/static/pages/contact.js
```diff
@@ -1,17 +1,44 @@
...
 
 const Contact = (() => {
+    let $chat_button,
+        $chat_unavailable;
 
     const onLoad = () => {
+        $chat_button      = $('#chat_button');
+        $chat_unavailable = $('#live_chat_unavailable');
+        showHideLiveChat();
+        setExternalTimer(showHideLiveChat);
```
binary-hamedanchi: Not sure if it's a good idea to call this every second.
How about a setTimeout with the remaining time till the next state?

src/templates/static/about/contact.jsx
```diff
@@ -77,6 +72,11 @@ const Contact = () => (
                                 </div>
                             </div>
                         </div>
+                        <div className='gr-parent ja-hide invisible' id='live_chat_unavailable'>
+                            <p className='notice-msg center-text'>
+                                {it.L('Sorry, live chat is currently unavailable. Please email us at [_1].', '<a href=\'mailto:support@binary.com\'>support@binary.com</a>')}
```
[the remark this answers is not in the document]
negarn: @OskarAhl no, I confirmed it with Reuben for the better text

src/javascript/static/pages/contact.js
negarn: Done, tried to make it as clear as possible, let me know if it can be improved.

PR #3985, opened 2018-06-18
kelly/update_logo_and_tagline

kellybinary:
- Update tagline on trading pages
- Show old logo for MF accounts


src/javascript/app/common/content_visibility.js
```diff
@@ -74,12 +74,17 @@ const ContentVisibility = (() => {
...
+        };
+
         document.querySelectorAll('[data-show]').forEach(el => {
             const attr_str              = el.dataset.show;
             const { is_exclude, names } = parseAttributeString(attr_str);
             const rule_set              = new Set(names);
 
-            const rule_set_has_current = rule_set.has(current_landing_company_shortcode);
+            const rule_set_has_current = rule_set.has(current_landing_company_shortcode) ||
+                Client.hasAccountType(map_account_type[names]);
```
binary-hamedanchi: Will this cause unwanted display issues in other places?
i.e. to display `maltainvest` contents for `malta` and `iom` clients. (Since they have another account by maltainvest)

kellybinary: yes it will 😅 I added this to handle virtual landing companies. i guess by for virtual accounts, we'll just show the default llogo altho clients might have MF accounts?

src/templates/app/_layout/header.jsx
```diff
@@ -74,7 +74,8 @@ const Header = () => (
                                         <div />
                                     </div>
                                     <div className='gr-9 gr-hide-m gr-hide-p binary-logo-text'>
-                                        <div />
+                                        <div data-show='-maltainvest' />
+                                        <span data-show='maltainvest' className='binary-type-logo-old'/>
```
binary-hamedanchi: By changing `span` to `div`:
- It receives same styles, then only `background-image` would be enough, no need to duplicate others.
- Won't need the `display: inline-block;` (this caused 3px increase on height of Header)

src/javascript/app/common/content_visibility.js
binary-hamedanchi: The current logged-in account is the one that matters.

PR #3988, opened 2018-06-18
kelly/missing_required_field_banner

kellybinary:


src/javascript/app/base/header.js
```diff
@@ -265,44 +265,61 @@ const Header = (() => {
...
 
+            const hasMissingRequiredField = () => {
+                const required_field = [
+                    'account_opening_reason',
+                    'address_line_1',
+                    'address_city',
+                    'phone',
+                    'tax_residence',
+                    'tax_identification_number',
+                    ...Client.get('residence') === 'gb' ? 'address_postcode' : '',
```
binary-hamedanchi: Probably meant:
```js
...(Client.get('residence') === 'gb' ? ['address_postcode'] : []),
```

kellybinary: thanks, good catch :)

src/javascript/app/base/header.js
```diff
@@ -265,44 +265,61 @@ const Header = (() => {
...
+                ];
+                const get_settings = State.getResponse('get_settings');
+                return required_field.some(field => get_settings[field] === '');
```
binary-hamedanchi: To handle possible API inconsistencies, I'd suggest `...some(field => !get_settings[field]);`

src/javascript/app/base/header.js
```diff
@@ -275,6 +289,7 @@ const Header = (() => {
                 document_review      : () => buildMessage('We are reviewing your documents. For more details [_1]contact us[_2].',                                                               'contact'),
                 excluded_until       : () => buildMessage('Your account is restricted. Kindly [_1]contact customer support[_2] for assistance.',                                                 'contact'),
                 financial_limit      : () => buildMessage('Please set your [_1]30-day turnover limit[_2] to remove deposit limits.',                                                             'user/security/self_exclusionws'),
+                required_fields      : () => buildMessage('Please complete your [_1]personal details[_2] before you proceed.', 'user/settings/detailsws'),
```
negarn: `req` should be after `res` alphabetically 😅

PR #3994, opened 2018-06-19
[new-app] Statement refactor (separate logic and state from UI)

ruskakimov:
[Test link](https://ruskakimov.binary.sx/br_statement_refactor/en/app.html)


src/javascript/app_2/store/pages/statement/statement_model.js
```diff
@@ -0,0 +1,84 @@
...
+            this.fetchNextBatch();
+        }
+    }
+}
+
+/*
+    RESPONSE PARSING FUNCTION
+*/
+const getStatementData = (statement, currency) => {
+    const date_obj   = new Date(statement.transaction_time * 1000);
```
morteza-binary: @ruskakimov 
There are some helper functions in [`app_2/common/date_time.js`](https://github.com/binary-com/binary-static/blob/new-app/src/javascript/app_2/common/date_time.js) to convert `epoch` to `moment` and `GMT` format.

PR #4000, opened 2018-06-19
kelly/fix_financial_assessment

kellybinary:


src/javascript/app/pages/user/account/settings/financial_assessment.js
```diff
@@ -97,6 +101,7 @@ const FinancialAssessment = (() => {
                     BinarySocket.send({ get_account_status: 1 }).then(() => {
                         Header.displayAccountStatus();
                         displayHighRiskClassification();
+                        getFinancialAssessment();
```
binary-hamedanchi: Is there a reason that it needs to wait for the response of `get_account_status` to get called? (Moving it outside the `send` will increase performance a little bit)

PR #4003, opened 2018-06-20
kelly/set_explaination_tab_as_default

kellybinary:


src/javascript/app/pages/trade/get_active_tab.js
```diff
@@ -24,7 +23,7 @@ const getActiveTab = (item) => {
     return selected_tab;
 };
 
-const getDefaultTab = (is_chart_allowed) => is_chart_allowed ? 'tab_graph' : 'tab_explanation';
+const getDefaultTab = () => 'tab_explanation';
```
negarn: do we need this function at all?

PR #4006, opened 2018-06-20
kelly/fix_tradepage

kellybinary:
- hide Reset time row if the contract is not reset
- handle undefined barrier error in the console


src/javascript/app/pages/trade/process.js
```diff
@@ -242,6 +243,7 @@ const Process = (() => {
             } else {
                 Defaults.set('selected_tick', selected_tick.value);
             }
+            Reset.hideResetTime();
```
negarn: why specifically for this type of trade? can't we hide it in a more general place?

PR #4007, opened 2018-06-20
ruskakimov/intermittent_sell_time_issue

ruskakimov:


src/javascript/app/pages/trade/charts/highchart.js
```diff
@@ -579,7 +579,7 @@ const Highchart = (() => {
     const endContract = () => {
         if (chart && !stop_streaming) {
             drawLineX({
-                value     : (isSoldBeforeExpiry() ? sell_time : end_time),
+                value     : (isSoldBeforeExpiry() ? (sell_time || sell_spot_time) : end_time),
```
negarn: we need to test this for all types of trades, to make sure it doesn't break anything.

src/javascript/app/pages/trade/charts/highchart.js
```diff
@@ -608,12 +608,14 @@ const Highchart = (() => {
                     i++;
                 }
                 const last = parseInt(last_data.x || last_data[0]);
-                if (last > (end_time * 1000) || last > ((sell_time || sell_spot_time) * 1000)) {
+                if (last >= (end_time * 1000) || last >= ((sell_time || sell_spot_time) * 1000)) {
```
negarn: I think it was `>` so that it would draw one tick after the end, is there any reason to remove that?

ruskakimov: oh it was intentional? i will revert it then 👍

src/javascript/app/pages/trade/charts/highchart.js
```diff
@@ -608,12 +608,14 @@ const Highchart = (() => {
...
                     stop_streaming = true;
                 } else {
                     // add a null point if the last tick is before end time to bring end time line into view
                     const time = isSoldBeforeExpiry() ? (sell_time || sell_spot_time) : end_time;
-                    chart.series[0].addPoint({ x: ((time || window.time.unix()) + margin) * 1000, y: null });
+                    const point = { x: time * 1000, y: null };
+                    chart.addSeries({ data: [point] });
+                    chart.zoomOut(); // sometimes navigator doesn't show new points
```
negarn: if we are zooming out instead of adding a null point with the margin value, maybe we can update the above comment

src/javascript/app/pages/trade/charts/highchart.js
```diff
@@ -488,20 +488,16 @@ const Highchart = (() => {
     // calculate where to display the maximum value of the x-axis of the chart for line chart
     const getMaxHistory = (history_times) => {
         let end = end_time;
-        if (sell_spot_time && (sell_time || sell_spot_time) < end_time) {
-            end = sell_spot_time;
+        if (sell_time && sell_time < end_time) {
+            end = sell_time;
```
ruskakimov: our line is drawn at `sell_time`, we should end on the tick after `sell_time` so the line is always visible

src/javascript/app/pages/trade/charts/highchart.js
```diff
@@ -488,20 +488,16 @@ const Highchart = (() => {
...
         }
 
         const history_times_length = history_times.length;
         if (is_settleable || is_sold) {
-            for (let i = history_times_length - 1; i >= 0; i--) {
-                if (parseInt(history_times[i]) === end) {
-                    max_point = parseInt(history_times[i === history_times_length - 1 ? i : i + 1]);
-                    break;
-                }
-            }
```
ruskakimov: since we use `sell_time` now, sometimes it is between ticks, which means `parseInt(history_times[i]) === end` is never true

src/javascript/app/pages/trade/charts/highchart.js
```diff
@@ -598,7 +594,7 @@ const Highchart = (() => {
     };
 
     const setStopStreaming = () => {
-        if (chart && (is_sold || is_settleable)) {
+        if (chart && (is_sold || is_settleable) && (isSoldBeforeExpiry() ? sell_time : end_time)) {
```
ruskakimov: `(isSoldBeforeExpiry() ? sell_time : end_time)` check ensures that sell or end line is drawn

src/javascript/app/pages/trade/charts/highchart.js
```diff
@@ -610,10 +606,6 @@ const Highchart = (() => {
                 const last = parseInt(last_data.x || last_data[0]);
                 if (last > (end_time * 1000) || last > ((sell_time || sell_spot_time) * 1000)) {
                     stop_streaming = true;
-                } else {
-                    // add a null point if the last tick is before end time to bring end time line into view
-                    const time = isSoldBeforeExpiry() ? (sell_time || sell_spot_time) : end_time;
-                    chart.series[0].addPoint({ x: ((time || window.time.unix()) + margin) * 1000, y: null });
```
ruskakimov: why not just wait for ticks to come, adding `null` point always causes some weird issues

PR #4008, opened 2018-06-20
Prevent from sending incorrect data to the Back-End (Optimization - Low Priority)

jbaek7023:
## Observation
In Trade app page,
- form(barriers and inputs) sends `proposal` data even after it didn't pass the incorrect.

## Why should we fix this?
- There is no need to pass the wrong data to the Back-End
    - Back-End returns response with an error message. 

## Solution
- We can validate user's input and send the data on our hand (in Front-End).

## Current Fix
1. Check if the higher barrier is greater than lower barrier 
2. Re-render - Send generated param to `display`)

## Further Implementation Plan
1. Check if the validation is correct (better than checking for each barrier every time).
    - If the validation is incorrect, prevent from sending the current proposal data to the Back-End.

### DIFF
--- a/src/javascript/app/pages/trade/event.js
+++ b/src/javascript/app/pages/trade/event.js
@@ -147,6 +147,7 @@ const TradingEvents = (() => {
          * bind event to change in duration units, populate duration and request price
          */
         getElementById('duration_units').addEventListener('change', (e) => {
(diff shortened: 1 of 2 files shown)


ashkanx:
@jbaek7023 conflicts

PR #4010, opened 2018-06-20
FE check for higehr/lower barriers

jbaek7023:


src/templates/app/trade/trading.jsx
```diff
@@ -109,6 +109,9 @@ const Trading = () => (
                                         <div className='big-col'>
                                             <input id='barrier_high' type='text' data-lpignore='true' name='H' autoComplete='off' />
                                             <span id='indicative_high_barrier_tooltip' data-balloon={it.L('This is an indicative barrier. Actual barrier will be the entry spot plus the barrier offset.')} data-balloon-length='xlarge' />
+                                            <div className='hint'>
+                                                <span id='barrier_high_error' className='error-msg invisible'>{it.L('High barrier must be higher than lower barrier')}</span>
```
negarn: should be `low barrier` instead of `lower barrier`

src/javascript/app/pages/trade/barriers.js
```diff
@@ -160,14 +161,75 @@ const Barriers = (() => {
         Defaults.remove('barrier', 'barrier_high', 'barrier_low');
     };
 
-    const validateBarrier = () => {
+    const showError = (barrier) => {
+        if (!barrier.classList.contains('error-field')) {
+            barrier.classList.add('error-field');
+        }
+        const error_node = barrier.parentNode.lastElementChild.firstElementChild;
```
negarn: this is too path dependent, can we use some class identifier?

src/javascript/app/pages/trade/barriers.js
```diff
@@ -160,14 +161,75 @@ const Barriers = (() => {
...
+        if (error_node.classList.contains('invisible')) {
+            error_node.classList.remove('invisible');
+        }
+    };
+
+    const resolveError = (barrier) => {
```
negarn: generally we use 'hide' as opposed to 'show'

src/javascript/app/pages/trade/barriers.js
```diff
@@ -160,14 +161,75 @@ const Barriers = (() => {
...
+        if (barrier.classList.contains('error-field')) {
+            barrier.classList.remove('error-field');
+        }
+        const error_node = barrier.parentNode.lastElementChild.firstElementChild;
+        if (!error_node.classList.contains('invisible')) {
+            error_node.classList.add('invisible');
+        }
+    };
```
negarn: it seems these two functions are exactly the same, except one uses remove, add and the other uses add, remove. could be worth combining them with a flag to hide/show the error.

src/javascript/app/pages/trade/barriers.js
```diff
@@ -160,14 +161,75 @@ const Barriers = (() => {
...
+
+    /**
+    * Show an error on the target Barrier.
+    *
+    * @param {Object} barrier_1   the target barrier object for prompting error
+    * @param {Object} barrier_2   barrier object whose errors will be resolved
+    */
+    const showThisError = (barrier_1, barrier_2) => {
```
negarn: we don't normally use `This` in names, probably better to be more specific about what `This` is.

src/javascript/app/pages/trade/event.js
```diff
@@ -367,6 +367,7 @@ const TradingEvents = (() => {
          */
         const low_barrier_element = getElementById('barrier_low');
         low_barrier_element.addEventListener('input', CommonTrading.debounce((e) => {
+            Barriers.validateBarrier();
```
negarn: why not pass if it's low or high calling validateBarrier here, instead of checking if low or high value in store is different from input field?

src/javascript/app/pages/trade/barriers.js
jbaek7023: maybe it can have another param for `remove=True`?

src/javascript/app/pages/trade/barriers.js
jbaek7023: In that way, we have to pass extra two params to `hideError` `showError`. And I wanted to avoid it.

There are two ways to make this better:
1) (Your suggestion) pass extra two params 
2) Or we can dynamically add or remove DOM. 
- Append Error DOM dynamically.`const  newDom = createElementDOM('<div>').className='hint'; 

Which way do you think it's better?

src/javascript/app/pages/trade/event.js
jbaek7023: Because We can generalize the function for uses of validation in both `barriers.jsx` and `event.js`.

- validateBarrier() is used in `display()` function in `barriers.jsx` (no e.target). 
    - there is no `e.target.value` in display() -  `barrier.js (line 76 and 146)` .

src/javascript/app/pages/trade/barriers.js
jbaek7023: Oh we can get the DOM from hideError function. Uhmm

negarn: let's keep it as is for now

negarn: wait, I got confused. this comment is about using `.parentNode.lastElementChild.firstElementChild` instead of some common class like in parent, look for child with class error-message or something like that, to avoid depending on it being the first child of the last child

src/javascript/app/pages/trade/barriers.js
negarn: this is the one that I meant to comment let's leave as it is, let's leave the show error and hide error as two different functions

src/javascript/app/pages/trade/barriers.js
jbaek7023: No, actually your first suggestion was absolutely correct and better :) Thanks

src/javascript/app/pages/trade/event.js
negarn: but in barrier.js it's not user input, so we can just default to first_time, no?

src/javascript/app/pages/trade/barriers.js
```diff
@@ -160,14 +161,73 @@ const Barriers = (() => {
         Defaults.remove('barrier', 'barrier_high', 'barrier_low');
     };
 
-    const validateBarrier = () => {
+    const showError = (barrier) => {
+        if (!barrier.classList.contains('error-field')) {
+            barrier.classList.add('error-field');
+        }
+        const error_node = barrier.parentNode.getElementsByClassName('hint')[0].getElementsByClassName('error-msg')[0];
```
jbaek7023: Could be a better way to find error_node... Anyone is welcomed to comment for suggestion!

negarn: do we need this part? `.getElementsByClassName('hint')[0]`

jbaek7023: Nice! thanks

src/javascript/app/pages/trade/barriers.js
```diff
@@ -160,14 +161,73 @@ const Barriers = (() => {
...
+            if (is_high_barrier_greater) {
+                resolveAllErrors(barrier_high_element, barrier_low_element);
+            } else {
+                showErrorOnTarget(barrier_high_element, barrier_low_element);
+            }
+        } else if (is_high_barrier_greater) {
+            resolveAllErrors(barrier_high_element, barrier_low_element);
+        } else {
+            showErrorOnTarget(barrier_low_element, barrier_high_element);
+        }
```
negarn: how about simplifying this a bit? something like:

```
if (+barrier_high_element.value > +barrier_low_element.value) {
    hideError(barrier_high_element);
    hideError(barrier_low_element);
} else {
    hideError(is_high_barrier_changed ? barrier_low_element : barrier_high_element);
    showError(is_high_barrier_changed ? barrier_high_element : barrier_low_element);
}
```

jbaek7023: oh no... it causes PageSpeed loss only when we find for elements in `if` statement :) nvm.

src/javascript/app/pages/trade/barriers.js
```diff
@@ -160,14 +161,52 @@ const Barriers = (() => {
...
+    * Validate Barriers
+    * @param {Boolean} is_high_barrier_changed Whether we're validating this barrier.
+    *                                          And the default validation is on High barrier.
+    *
+    */
+    const validateBarrier = (is_high_barrier_changed = true) => {
         const barrier_element = getElementById('barrier');
         const empty           = isNaN(parseFloat(barrier_element.value)) || parseFloat(barrier_element.value) === 0;
+        const barrier_high_element = getElementById('barrier_high');
+        const barrier_low_element = getElementById('barrier_low');
```
negarn: align by `=`

src/javascript/app/pages/trade/barriers.js
```diff
@@ -160,14 +161,52 @@ const Barriers = (() => {
...
+
```
negarn: extra empty line

src/javascript/app/pages/trade/barriers.js
```diff
@@ -160,14 +161,52 @@ const Barriers = (() => {
...
+    const hideError = (barrier) => {
+        if (barrier.classList.contains('error-field')) {
+            barrier.classList.remove('error-field');
+        }
+        const error_node = barrier.parentNode.getElementsByClassName('hint')[0].getElementsByClassName('error-msg')[0];
+        if (!error_node.classList.contains('invisible')) {
+            error_node.classList.add('invisible');
+        }
+    };
```
negarn: I still think we could combine these two, but it's up to you. it would look something like this: (I think by default it doesn't add/remove the class if it already exists/already has been removed, so it doesn't need the explicit check)

```
const showHideError = (barrier, should_show_error) => {
    barrier.classList[should_show_error ? 'add' : 'remove']('error-field');
    const error_node = barrier.parentNode.getElementsByClassName('error-msg')[0];
    error_node.classList[should_show_error ? 'remove' : 'add']('invisible');
};
```

jbaek7023: That looks good but I have to disagree with you. 

We'll pass `Boolean` to `showHideError` and developers will need to see what the function looks like before they use `showHideError`. Better to call showError or hideError instead (more logical... I like this way better!).

negarn: then I suggest keeping the showError and hideError to call this showHideError internally, so if we want to change something e.g. in error node we don't need to update two sets of the same code :)

jbaek7023: Sure, thanks!

src/javascript/app/pages/trade/barriers.js
```diff
@@ -144,10 +144,11 @@ const Barriers = (() => {
                     high_elm.value = high_elm.textContent = value_high;
                     low_elm.value  = low_elm.textContent  = value_low;
 
-                    Defaults.set('barrier_high', high_elm.value);
```
jbaek7023: There was no need to make a change here though :)

src/javascript/app/pages/trade/barriers.js
```diff
@@ -160,14 +161,27 @@ const Barriers = (() => {
...
+    /**
+    * Validate Barriers
+    */
     const validateBarrier = () => {
-        const barrier_element = getElementById('barrier');
-        const empty           = isNaN(parseFloat(barrier_element.value)) || parseFloat(barrier_element.value) === 0;
+        const barrier_element      = getElementById('barrier');
+        const empty                = isNaN(parseFloat(barrier_element.value))||parseFloat(barrier_element.value) === 0;
+        const barrier_high_element = getElementById('barrier_high');
+        const barrier_low_element  = getElementById('barrier_low');
```
negarn: do we need `barrier_low_element`?

src/javascript/app/pages/trade/barriers.js
```diff
@@ -160,14 +161,27 @@ const Barriers = (() => {
...
+        const error_node           = getElementById('barrier_high_error');
```
negarn: we can define this inside the `if (isVisible(barrier_high_element)) {` on line 181 or so

src/javascript/app/pages/trade/barriers.js
jbaek7023: Yes. Shouldn't we compare the values of low barrier and high barrier?

negarn: it can move inside the if condition as well

PR #4013, opened 2018-06-20
Fixed basename for branches

morteza-binary:


src/javascript/app_2/app.js
```diff
@@ -42,7 +42,7 @@ const initApp = () => {
  * @return {string} returns the basename of current url
  */
 const getBasename = () => {
-    const regex_string = `((${Object.keys(getAllLanguages()).join('|')})\/app\.html).*`;
+    const regex_string = `(.*(${Object.keys(getAllLanguages()).join('|')})\/app\.html).*`;
```
zakibinary: Maybe we can remove the escape character `\` before `app\.html`?

morteza-binary: @zakibinary I've removed it in another PR.

PR #4015, opened 2018-06-21
kelly/ui_updates

kellybinary:


src/javascript/app/pages/user/view_popup/view_popup.js
```diff
@@ -320,8 +320,9 @@ const ViewPopup = (() => {
     };
 
     const contractEnded = () => {
-        containerSetText('trade_details_current_title', localize(contract.status === 'sold' || (contract.sell_spot_time < contract.date_expiry) ? 'Contract Sold' : 'Contract Expiry'));
+        getElementById('trade_details_live_date').parentNode.setVisibility(0);
```
negarn: probably we should also stop updating the time, since it's now hidden?

src/javascript/app/pages/user/view_popup/view_popup.js
```diff
@@ -283,6 +283,8 @@ const ViewPopup = (() => {
             $container.find('#notice_ongoing').setVisibility(1);
         }
 
+        Clock.stopExternalTimer(is_ended); // stop/resume timer
```
negarn: why not just set external timer to blank? e.g. `Clock. setExternalTimer();` should achieve the same thing

src/javascript/app/pages/user/view_popup/view_popup.js
```diff
@@ -283,6 +283,8 @@ const ViewPopup = (() => {
             $container.find('#notice_ongoing').setVisibility(1);
         }
 
+        Clock.setExternalTimer(is_ended ? '' : updateTimers); // stop/resume timer
```
kellybinary: we need to restart the timer when visiting portfolio after visiting profit table or statement page

PR #4021, opened 2018-06-22
[new-app] purchase_and_other_ui_fixes

zakibinary:


src/javascript/app_2/components/elements/drawer/toggle_drawer.jsx
```diff
@@ -10,7 +10,7 @@ class ToggleDrawer extends PureComponent {
         const { alignment } = this.props;
         if (alignment && alignment === 'left') {
```
negarn: I think for all of these we can just check `if  (alignment === '[...]')`, no need to check if alignment exists, since if it's undefined, `undefined === 'left'` will just return false as expected

src/javascript/app_2/components/elements/menu_drawer.jsx
```diff
@@ -5,36 +5,43 @@ import { DrawerItems, DrawerItem } from '../elements/drawer/index.jsx';
...
-                <DrawerItem text={localize('Contact Us')}/>
-                {Client.isLoggedIn() &&
-                    <DrawerItem text={localize('Logout')} custom_action={requestLogout}/>
-                }
-            </div>
-        </PerfectScrollbar>
-    </div>
-);
+export const MenuDrawer = () => {
+    const is_mobile = window.innerWidth < 980;
```
negarn: isn't that more of a tablet size, than a mobile size?

zakibinary: What if I changed it to `is_desktop = window.innerWidth > 979` instead and used `!is_desktop && ..` on L31?
We don't really have a concrete reference for tablet sizes in `trading_app`.

negarn: sounds fine to me

src/javascript/app_2/pages/trading/components/purchase.jsx
```diff
@@ -1,28 +1,47 @@
...
+        <div>{localize('Start')}: {toGMTFormat(purchase_info.start_time * 1000)}</div>
+        <div>{localize('Contract ID')}: {purchase_info.contract_id}</div>
+        <div>{localize('Transaction ID')}: {purchase_info.transaction_id}</div>
+        <div>{localize('Description')}: {purchase_info.longcode}</div>
     </React.Fragment>
 );
 
+const LogInMessage = () => (
+    <div className='purchase-login-wrapper'>
+        <span className='info-text'>{localize('Please login to purchase the contract  ')}</span>
```
negarn: I think it should be `log in` rather than `login`

negarn: also we should remove the extra whitespace at the end

src/javascript/app_2/pages/trading/components/purchase.jsx
```diff
@@ -1,28 +1,47 @@
...
+        <Button
+            className='secondary orange'
+            has_effect
+            text={localize('LOGIN')}
```
negarn: same here, I think it should be `log in`. can we make the text into uppercase using CSS instead, to avoid duplication for translators?

src/javascript/app_2/pages/trading/components/purchase.jsx
```diff
@@ -1,28 +1,47 @@
...
+            onClick={redirectToLogin}
+        />
+        <p>{localize('Don\'t have a Binary.com account?')}</p>
```
negarn: are we using website_name for new app, or going to just use `Binary.com`?

src/javascript/app_2/pages/trading/components/purchase.jsx
```diff
@@ -32,36 +51,59 @@ const Purchase = ({
     trade_types,
 }) => (
     Object.keys(trade_types).map((type, idx) => {
-        const info = proposal_info[type] || {};
+        const info         = proposal_info[type] || {};
         const is_logged_in = isLoggedIn();
+        const el_currency  = <span className={`symbols ${(Client.get('currency') || 'usd').toLowerCase()}`}/>;
```
negarn: can we use `USD` in all caps, so making it lowercase makes sense? :P

zakibinary: Sure :)

src/javascript/app_2/pages/trading/components/purchase.jsx
zakibinary: I don't think we can use `website_name` as that is for static? I'll just use `localize('... [_1], ['Binary.com'])`

src/javascript/app_2/components/elements/account_balance.jsx
```diff
@@ -42,7 +45,7 @@ export const AccountBalance = connect(
                     }
                 </React.Fragment> :
                 <Button
-                    className='primary green'
+                    className='primary orange'
                     has_effect
                     text={localize('Login')}
```
negarn: `log in` as well?

zakibinary: Thanks, missed this. 😅

PR #4025, opened 2018-06-25
Set the remaining time "-" when the contract is ended

jbaek7023:


src/javascript/app/base/binary_pjax.js
```diff
@@ -166,7 +166,8 @@ const BinaryPjax = (() => {
 
         const query_params = Url.paramsHash();
         if (!query_params.anchor) {
-            $.scrollTo('body', 500);
+            // $.scrollTo('body', 500); // this jQuery function causes an issue and can be replaced by the function below
```
binary-hamedanchi: The reported issue might have a different reason, as `$.scrollTo` is being used in many places.

src/javascript/app/base/binary_pjax.js
```diff
@@ -166,7 +166,8 @@ const BinaryPjax = (() => {
...
+            $('body').animate({ scrollTop: 500 });
```
binary-hamedanchi: Probably:
```js
$('html, body').animate({ scrollTop: 0 }, 500);
```

jbaek7023: Thanks! 👍

PR #4026, opened 2018-06-26
End Between/End outside "double Barrier missing"

jbaek7023:


src/javascript/app/pages/trade/barriers.js
```diff
@@ -27,9 +26,10 @@ const Barriers = (() => {
             const unit     = getElementById('duration_units');
             const end_time = getElementById('expiry_date');
             const is_daily = (unit && isVisible(unit) && unit.value === 'd') ||
-                (end_time && isVisible(end_time) && moment(end_time.getAttribute('data-value')).isAfter(moment(), 'day'));
+                (end_time && isVisible(end_time));
```
jbaek7023: Showing barriers even if the end-time is the current date.

src/javascript/app/pages/trade/duration.js
```diff
@@ -184,7 +184,7 @@ const Durations = (() => {
         const date_start     = CommonFunctions.getElementById('date_start').value;
         const now            = !date_start || date_start === 'now';
         const current_moment = moment((now ? window.time : parseInt(date_start) * 1000)).add(5, 'minutes').utc();
-        let expiry_date      = Defaults.get('expiry_date') ? moment(Defaults.get('expiry_date')) : current_moment;
+        let expiry_date      = Defaults.get('expiry_date') ? moment(Defaults.get('expiry_date')) : current_moment.add(1, 'days').utc();
```
jbaek7023: Adjust end date so that user can see the valid end-time

src/javascript/app/pages/trade/barriers.js
negarn: this is the check for `is_daily` value, if we remove that check then `is_daily` is not being decided correctly. we use this to show intraday or daily barrier values and need to keep this check.

src/javascript/app/pages/trade/duration.js
negarn: this will add one day for all trade types, but some trades are fine to be on the same day. we need a better solution to only add one day for contracts that only allow daily duration.

src/javascript/app/pages/trade/barriers.js
ruskakimov: I presume `is_daily` means contracts lasts 1 day or more, so with this change wouldn't `is_daily` loose it's semantic meaning? So for other contracts with end time `is_daily` would be true always, even if end time is in one minute or something.

ruskakimov: (oops, duplicate)

src/javascript/app/pages/trade/duration.js
```diff
@@ -180,11 +180,26 @@ const Durations = (() => {
         return (value * seconds[from_unit]) / seconds[to_unit];
     };
 
+    const getSmallestDuration = () => {
+        const unit               = CommonFunctions.getElementById('duration_units');
```
negarn: I'm not sure if it's a good idea to get this from DOM when you can get it from something like `list` as above on line 134. we can turn that into a global variable or store the smallest value and return it here.

jbaek7023: I think we don't really need this function because we don't need `unit_name` and `unit_value` too.

jbaek7023: I'm not sure how we can compare the values without checking the value from DOM. This change compare the current value of DOM and local storage, and returns the latest (higher value) of Date.

Previously if you go to Volatility(today's date) and go to Forex tab, the date barrier sets to current date and it caused an error.

negarn: you can look at `list` above as mentioned, it already has this list of values stored.

jbaek7023: oh ok got it 👍

PR #4028, opened 2018-06-26
kelly/fix_trackjs

kellybinary:
should resolve two issues:
- Cannot read property 'setVisibility' of null
- Cannot read property 'length' of undefined


src/javascript/app/pages/trade/tick_trade.js
```diff
@@ -337,7 +337,7 @@ const TickDisplay = (() => {
     const dispatch = (data) => {
         const tick_chart = CommonFunctions.getElementById(id_render);
 
-        if (!CommonFunctions.isVisible(tick_chart) || !data || (!data.tick && !data.history)) {
+        if (!CommonFunctions.isVisible(tick_chart) || !data || (!data.tick && !data.history) || applicable_ticks === 'undefined') {
```
binary-hamedanchi: Missing `typeof`?

src/javascript/app/pages/trade/tick_trade.js
```diff
@@ -337,6 +337,10 @@ const TickDisplay = (() => {
     const dispatch = (data) => {
         const tick_chart = CommonFunctions.getElementById(id_render);
 
+        if (typeof applicable_ticks === 'undefined') {
+            applicable_ticks = [];
```
negarn: why not just initiate it as an empty array on definition?

kellybinary: currently it is initialized here: https://github.com/binary-com/binary-static/blob/master/src/javascript/app/pages/trade/tick_trade.js#L334, probably we can init it here as well: https://github.com/binary-com/binary-static/blob/master/src/javascript/app/pages/trade/tick_trade.js#L56 ?

PR #4031, opened 2018-06-26
kelly/signup_page

kellybinary:
- separate Signup page


src/javascript/static/pages/signup.js
```diff
@@ -0,0 +1,82 @@
+const BinarySocket   = require('../../app/base/socket');
+const FormManager    = require('../../app/common/form_manager');
+const getElementById = require('../../_common/common_functions').getElementById;
+const localize       = require('../../_common/localize').localize;
+const Login          = require('../../_common/base/login');
```
negarn: line 5 should be on line 3

src/javascript/static/pages/signup.js
```diff
@@ -0,0 +1,82 @@
...
+        $google_btn,
+        $login_btn;
+
+    const form_id = '#signup_form';
+
+    const onLoad = () => {
+        getElementById('footer').setVisibility(0); // always hide footer in this page
+
+        $google_btn = $('#google-signup');
+        $login_btn  = $('#login');
```
negarn: some elements are stored, others are called on the fly, e.g. `$('.error-msg')`, `$(form_id)`, `$('#verify_email')`, it would be better to use the same standard

src/javascript/static/pages/signup.js
```diff
@@ -0,0 +1,82 @@
...
+                { request_field: 'type', value: 'account_opening' },
+            ]);
+            FormManager.handleSubmit({
+                form_selector       : form_id,
+                fnc_response_handler: verifyEmailHandler,
+                fnc_additional_check: checkCountry,
+            });
+            $('.error-msg').addClass('center-text');
+
+            $google_btn.off('click').on('click', (e) => {
```
negarn: do we need the `off` here when on unload we are removing it?

src/sass/_common/common.scss
```diff
@@ -638,6 +638,21 @@ label + input {
     }
 }
 
+#signup_form {
```
negarn: does it belong under common? seems only specific to this page, might be more suited for pages.scss, or websocket_pages.scss depending

src/javascript/app/pages/signup.js
```diff
@@ -0,0 +1,85 @@
...
+            ]);
+            FormManager.handleSubmit({
+                form_selector       : form_id,
+                fnc_response_handler: verifyEmailHandler,
+                fnc_additional_check: checkCountry,
+            });
+            $google_btn   = $('#google-signup');
+            $login_btn    = $('#login');
+            $verify_email = $('#verify_email');
+            $error_msg    = $('.error-msg');
```
negarn: we should be able to declare all of these and call form manager without having to wait for website_status, so on line 18 or so, everything except for line 20 should be able to move out

PR #4035, opened 2018-06-27
kelly/hide_mt5_menu

kellybinary:


src/templates/app/_layout/header.jsx
```diff
@@ -109,7 +109,7 @@ const Header = () => (
                                                 { text: it.L('SmartTrader'),  href: it.url_for('trading'),          className: 'no-capitalize' },
                                                 { text: it.L('WebTrader'),    href: 'https://webtrader.binary.com', target: '_blank' },
                                                 { text: it.L('Binary Bot'),   href: 'https://bot.binary.com',       target: '_blank' },
-                                                { text: it.L('MetaTrader 5'), href: it.url_for('user/metatrader') },
+                                                { text: it.L('MetaTrader 5'), href: it.url_for('user/metatrader'),  className: 'invisible', id: 'user_menu_metatrader' },
```
negarn: I suggest changing it to class, now that there will be more than one element on the page with this id

PR #4036, opened 2018-06-27
ruskakimov/callputequal_end_time_fix

ruskakimov:


src/javascript/app/pages/trade/process.js
```diff
@@ -274,9 +274,9 @@ const Process = (() => {
     const setFormName = (formname) => {
         let formname_to_set    = formname;
         const has_callputequal = hasCallPutEqual();
-        if (/^(callputequal)$/.test(formname_to_set) && (!has_callputequal || !+Defaults.get('is_equal'))) {
+        if (formname_to_set === 'callputequal' && (!has_callputequal || !+Defaults.get('is_equal'))) {
             formname_to_set = 'risefall';
-        } else if (/^(risefall)$/.test(formname_to_set) && has_callputequal && +Defaults.get('is_equal')) {
+        } else if (formname_to_set === 'risefall' && has_callputequal && +Defaults.get('is_equal')) {
```
ruskakimov: It is an equivalent operation. I think it is more readable this way.

src/javascript/app/pages/trade/duration.js
```diff
@@ -334,7 +334,7 @@ const Durations = (() => {
             let $expiry_date     = $('#expiry_date');
             const date_start_val = CommonFunctions.getElementById('date_start').value || 'now';
             const is_now         = isNow(date_start_val);
-            const is_risefall    = /risefall/.test(Defaults.get('formname')) || false;
+            const is_risefall    = /^(risefall|callputequal)$/.test(Defaults.get('formname')) || false;
```
negarn: possibly change the variable name

src/javascript/app/pages/trade/duration.js
```diff
@@ -378,25 +376,30 @@ const Durations = (() => {
...
+                requested = selectEndDate(selected_date);
             }
         }
         return requested;
     };
 
-    const appendExpiryDateValues = ($expiry_date, date) => {
-        $expiry_date.append($('<option/>', { text: date.format('ddd - DD MMM, YYYY'), 'data-value': toISOFormat(date) }));
+    const appendExpiryDateValues = ($expiry_date, date, selected_date) => {
+        $expiry_date.append($('<option/>', { text: date.format('ddd - DD MMM, YYYY'), 'data-value': toISOFormat(date), 'selected': toISOFormat(selected_date) === toISOFormat(date) }));
```
ruskakimov: `selected` attr is needed cos Dropdown from binary style doesn't use `data-value`.

Also I tried `date.isSame(selected_date, 'day')` instead of `toISOFormat(selected_date) === toISOFormat(date)`, but somehow it returned `true` for different days.

negarn: you can have a look at the `isSameDay` function to see how that works.

PR #4056, opened 2018-06-28
ruskakimov/disable_past_time

ruskakimov:


src/javascript/app/pages/trade/starttime.js
```diff
@@ -102,6 +102,11 @@ const StartDates = (() => {
                 Durations.onStartDateChange(first);
             }
         } else {
+            if (start_dates && start_dates.has_spot) {
+                const now_option = createElement('option', { value: 'now', text: localize('Now') });
+                CommonIndependent.getStartDateNode().appendChild(now_option);
+                has_now = 1;
+            }
```
ruskakimov: the problem was that time picker gets starting date from the `#date_start`, but it's value was `''` since it was an empty selector with no options

src/javascript/app/pages/trade/starttime.js
```diff
@@ -102,6 +102,11 @@ const StartDates = (() => {
...
             State.remove('is_start_dates_displayed');
             getElementById('date_start_row').style.display = 'none';
             getElementById('date_start').value = 'now';
```
ruskakimov: this line had no effect basically, value was still `''`

src/javascript/app/components/time_picker.js
```diff
@@ -55,6 +55,7 @@ const TimePicker = (() => {
                 options.minTime = time_now;
             }
             if (options.useLocalTime) options.minTime = options.minTime.local();
+            options.minTime.add(1, 'minute'); // rounding up seconds
```
ruskakimov: if time is `12:05:33` it would still show `05` minute option

PR #4064, opened 2018-07-02
ruskakimov/cashier_page_improvements

ruskakimov:


src/templates/app/cashier/payment_methods.jsx
```diff
@@ -24,13 +24,39 @@ const TableValues = ({ value }) => {
     );
 };
 
+const ReferenceLink = ({ href, className = '', title = '' }) => (
+    <a
+        className={`payment-methods__reference ${className} ${!href ? 'payment-methods__reference--disabled' : ''}`}
+        href={href}
+        target='_blank'
+        aria-disabled={!href}
+        title={title}
```
negarn: do we need the `rel='noopener noreferrer'`?

negarn:
generally fine, but maybe it's easier to use simpler (shorter) file names? 😅

PR #4068, opened 2018-07-03
ruskakimov/mt5_tabs_bug

ruskakimov:


src/javascript/app/pages/user/metatrader/metatrader.ui.js
```diff
@@ -380,6 +385,14 @@ const MetaTraderUI = (() => {
         $form.find('.mt5_type_box').click(selectAccountTypeUI);
     };
 
+    // restoreUI:
+    // On deposit response setCurrentAccount is called, which updates DOM
+    // So, if you change UI (e.g. to new_account) after triggering deposit and before BE response
+    // -> UI will get messed up, since setCurrentAccount assumes you are still on cashier UI
+    const restoreUI = (acc_type) => {
+        handleNewAccountUI(current_action_UI || 'cashier', acc_type);
+    };
```
ruskakimov: `setCurrentAccount` updates displayed balance and some other things

src/javascript/app/pages/user/metatrader/metatrader.ui.js
```diff
@@ -380,6 +384,14 @@ const MetaTraderUI = (() => {
         $form.find('.mt5_type_box').click(selectAccountTypeUI);
     };
 
+    // restoreUI:
+    // On deposit response setCurrentAccount is called, which updates DOM
+    // So, if you change UI (e.g. to new_account) after triggering deposit and before BE response
```
negarn: instead of BE please use API

src/javascript/app/pages/user/metatrader/metatrader.ui.js
```diff
@@ -380,6 +384,14 @@ const MetaTraderUI = (() => {
...
+    // -> UI will get messed up, since setCurrentAccount assumes you are still on cashier UI
+    const restoreUI = (acc_type) => {
+        handleNewAccountUI(current_action_ui || 'cashier', acc_type);
```
negarn: why cashier? why not get default action?

src/javascript/app/pages/user/metatrader/metatrader.ui.js
```diff
@@ -380,6 +384,14 @@ const MetaTraderUI = (() => {
         $form.find('.mt5_type_box').click(selectAccountTypeUI);
     };
 
+    // restoreUI:
+    // On deposit response setCurrentAccount is called, which updates DOM
```
negarn: what if instead of this, we check before updating DOM if you are on the expected action?

src/javascript/app/pages/user/metatrader/metatrader.ui.js
```diff
@@ -167,7 +168,9 @@ const MetaTraderUI = (() => {
     const setCurrentAccount = (acc_type) => {
         if (Client.get('mt5_account') && Client.get('mt5_account') !== acc_type) return;
 
-        displayAccountDescription(acc_type);
+        if (current_action_ui !== 'new_account') {
+            displayAccountDescription(acc_type);
+        }
 
         if (accounts_info[acc_type].info) {
```
negarn: I think you can move line 172 here, this condition means that action is not new account, because accounts info is already updated with the info. please test it to confirm, though.

ruskakimov: image
this check still passes on `new_account` page

negarn: hm, I'm just wondering if there's a better way to know the action, other than using a global variable.

ruskakimov: another way might be to find DOM element with `act_` and selected class, like
```
if ($('.act_new_account.selected'))
```
but I think it is much worse 😄

ruskakimov: i don't think there is some other way of knowing what current action is, since we just update DOM on button click.
[Bind callback to all act_ buttons](https://github.com/ruskakimov/binary-static/blob/344c3b5fedd305d55b81b193e5bef63e3df253a2/src/javascript/app/pages/user/metatrader/metatrader.ui.js#L41)
[Parse action name from class name](https://github.com/ruskakimov/binary-static/blob/344c3b5fedd305d55b81b193e5bef63e3df253a2/src/javascript/app/pages/user/metatrader/metatrader.ui.js#L225)

PR #4075, opened 2018-07-04
negar/improve_ux_document_submit

negarn:
show submit status


src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -55,287 +61,313 @@ const Authenticate = (() => {
         });
 
         $('.file-picker').on('change', onFileSelected);
+    };
 
-        /**
```
negarn: most of these changes are just indentation, everything used to be inside the `init` function for some reason and I brought them out and did some fixes, but generally code didn't change

PR #4084, opened 2018-07-06
kelly/fix_localize

kellybinary:


src/javascript/app/pages/user/account/settings/two_factor_authentication.js
```diff
@@ -41,7 +42,7 @@ const TwoFactorAuthentication = (() => {
             next_state    = state[+(!res.account_security.totp.is_enabled)].slice(0, -1);
 
             $(`#${current_state}`).setVisibility(1);
-            $btn_submit.text(localize(next_state));
+            $btn_submit.text(localize(capitalize(next_state)));
```
negarn: why not use `toTitleCase(localize(next_state))`?

kellybinary: ahh i forgot we have this function

PR #4093, opened 2018-07-10
ruskakimov/upload_selfie

ruskakimov:


src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -167,21 +167,25 @@ const Authenticate = (() => {
...
+                const file_obj  = {
                     file     : e.files[0],
                     chunkSize: 16384, // any higher than this sends garbage data to websocket currently.
                     class    : id,
                     type,
                     name,
                 };
+                if (page_type) {
+                    file_obj.page_type = page_type;
+                }
```
negarn: why not something like `file_obj.page_type = $e.attr('data-page-type')` ? if it doesn't exist it will be undefined, which on line 266 if `f.page_type` doesn't exist it will be undefined anyway, so we don't need an extra variable assignment and if condition

PR #4097, opened 2018-07-11
kelly/authenticated_account_status

kellybinary:


src/javascript/app/base/header.js
```diff
@@ -360,7 +360,7 @@ const Header = (() => {
                     get_account_status = State.getResponse('get_account_status') || {};
                     status             = get_account_status.status;
                     checkStatus(check_statuses_real);
-                    $('.account-id')[/authenticated/.test(status) ? 'append' : 'remove'](el_account_status);
+                    $('.account-id')[+get_account_status.prompt_client_to_authenticate ? 'remove' : 'append'](el_account_status);
```
negarn: it needs to check both conditions, is authenticated and does not have to prompt_client_to_authenticate

PR #4109, opened 2018-07-12
[new-app] negar/duration_start_end_fixes

negarn:
duration & start/end time fixes


src/javascript/app_2/Stores/Modules/Trading/Helpers/contract_type.js
```diff
@@ -209,24 +209,46 @@ const ContractType = (() => {
...
+                compare_moment.minute(minutes.find(m => isSessionAvailable(sessions, compare_moment.minute(m))) || compare_moment.format('mm'));
+            }
         }
-        return { start_time: `${hour}:${minute}` };
+        return compare_moment.format('HH:mm');
+    };
+
+    const buildMoment = (date, time) => {
+        const [ hour, minute ] = time.split(':');
+        return moment.utc(isNaN(date) ? date : +date * 1000).hour(hour).minute(minute);
```
kellybinary: perhaps `moment.utc(isNaN(date) ? undefined : ...` ?

src/javascript/app_2/Stores/Modules/Trading/Helpers/contract_type.js
```diff
@@ -209,24 +209,46 @@ const ContractType = (() => {
...
+    const getEndTime = (sessions, start_date, start_time, expiry_date, expiry_time) => {
+        const start_moment = start_date ? buildMoment(start_date, start_time) : moment().utc();
+        const end_moment   = buildMoment(expiry_date, expiry_time);
+
+        let end_time = expiry_time;
+        if (sessions && !isSessionAvailable(sessions, end_moment)) {
+            end_time = getValidTime(sessions, end_moment, start_moment);
+        }
+        if (end_moment.isSameOrBefore(start_moment)) {
+            const is_end_of_day     = start_moment.format('HH:mm') === '23:55';
```
kellybinary: is it possible for the start_moment to be more than 23:55? say 23:56, 23:57? if yes, perhaps we can use greater than equality here?

src/javascript/app_2/Stores/Modules/Trading/Helpers/contract_type.js
negarn: no, in case where it's not a number it's a string format of date, so we want to change that to moment

src/javascript/app_2/Stores/Modules/Trading/Helpers/contract_type.js
negarn: I can make the change, but I'm not sure that we do

kellybinary: hmm, i guess it's fine then 😄

PR #4115, opened 2018-07-16
[new-app] ruskakimov/portfolio_refactor

ruskakimov:


src/javascript/app_2/Modules/Portfolio/Containers/portfolio.jsx
```diff
@@ -1,256 +1,89 @@
...
-        footer     : {
-            reference : 'Total',
-            payout    : '',
-            purchase  : '',
-            indicative: '',
-        },
-        is_loading: true,
-        oauth_apps: null,
+        // TODO: get currency from store, once it has been added
+        currency: ClientBase.get('currency').toLowerCase(),
```
morteza-binary: @ruskakimov 
IMO, it's better to define the `currency` property in `ClientStore` and pass to the component by `connect` instead of adding local state.

src/javascript/app_2/Modules/Portfolio/Containers/portfolio.jsx
```diff
@@ -1,256 +1,89 @@
...
-                })()}
+            <div className={classnames('portfolio container', { 'portfolio--card-view': should_show_cards })}>
+                {
+                    should_show_cards ?
+                        <CardList data={data} currency={this.state.currency} />
+                        :
+                        <DataTable
+                            columns={getTableColumnsTemplate(this.state.currency)}
+                            data_source={data}
+                            footer={data.length > 0 ? totals : undefined} // don't show footer if table is empty
```
morteza-binary: @ruskakimov 
it can be `!!data.length`

src/javascript/app_2/Modules/Portfolio/Containers/portfolio.jsx
ruskakimov: that's what TODO stands for 😅

morteza-binary: @ruskakimov 
it has been added to the `ClientStore` by @kellybinary. You can fetch the last version of `new-app`

src/sass/app_2/_common/layout/header.scss
```diff
@@ -1,216 +1,212 @@
-#header {
+.header {
     position: fixed;
     z-index: 99;
     width: 100%;
     height: $HEADER_HEIGHT;
     top: 0;
+    border-bottom: 1px solid $COLOR_GRAY;
+    background-color: $COLOR_WHITE;
```
ruskakimov: git diff is a mess here, but I only made a few changes + tab indentation

src/javascript/app_2/Stores/Modules/Portfolio/portfolio_store.js
```diff
@@ -0,0 +1,144 @@
...
+        this.clearTable();
+        WS.forgetAll('proposal_open_contract', 'transaction');
+    }
+
+    @computed
+    get data_with_remaining_time() {
+        return this.data.map((portfolio_pos) => {
+            portfolio_pos.remaining_time = formatDuration(getDiffDuration(this.common.server_time, portfolio_pos.expiry_time));
+            return portfolio_pos;
+        });
```
ruskakimov: don't use es6 spread here, since it runs every second
this approach is 20 times faster than with es6 spread (`...portfolio_pos`)
[performance benchmark](http://jsben.ch/YTUEK)

ruskakimov: @binary-hamedanchi @morteza-binary should i add this to comments? 🤔

morteza-binary: @ruskakimov 
IMO, it's better to add it to comments, so it prevents others to change it unwittingly.

ruskakimov:
@binary-hamedanchi @morteza-binary ready for review 🙂

src/javascript/app_2/Utils/Date/date_time.js
```diff
@@ -39,4 +39,29 @@ export const formatDate = (date, date_format = 'YYYY-MM-DD') => moment(date || u
...
     const diff = moment(date).utc().diff(moment().utc(), 'days');
     return (!date || diff < 0) ? '' : diff + 1;
-};
\ No newline at end of file
+};
+
+/**
+ * return moment duration between two dates
+ * @param  {Number|moment} epoch start time or moment object
+ * @param  {Number|moment} epoch end time or moment object
```
binary-hamedanchi: `epoch` or `milliseconds`?

src/javascript/app_2/Utils/Date/date_time.js
```diff
@@ -39,4 +39,29 @@ export const formatDate = (date, date_format = 'YYYY-MM-DD') => moment(date || u
...
+ * @return {String} formatted display string
+ */
+export const formatDuration = (duration) => {
+    const d = Math.floor(duration.asDays()); // duration.days() does not include months/years
+    const h = duration.hours();
+    const m = duration.minutes();
+    const s = duration.seconds();
+    let formatted_str = moment(0).hour(h).minute(m).seconds(s).format('HH:mm:ss');
+    if (d > 0) {
+        formatted_str = `${d} ${d > 1 ? 'days' : 'day'} ${formatted_str}`;
```
binary-hamedanchi: Needs localization here.

src/javascript/app_2/Stores/Modules/Portfolio/portfolio_store.js
```diff
@@ -0,0 +1,145 @@
...
+import {
+    getDiffDuration,
+    formatDuration }               from '../../../Utils/Date';
+
+export default class PortfolioStore extends BaseStore {
+    @observable data       = [];
+    @observable is_loading = false;
+    @observable error      = '';
+
+    constructor(common) {
```
binary-hamedanchi: To prevent getting mistaken by local properties and also for consistency, better to use `root_store` everywhere.

src/javascript/app_2/Utils/Date/date_time.js
ruskakimov: if I'm not mistaken [epoch is in milliseconds](https://momentjs.com/docs/#/parsing/unix-timestamp-milliseconds/) :)

ruskakimov: but yeah we pass epoch as param

binary-hamedanchi: The term `epoch` refers to [_seconds_](https://momentjs.com/docs/#/parsing/unix-timestamp/) in our API methods. I think the descriptions here will create confusion.

This function expects milliseconds to be passed to (and it seems to be done like that in this PR). However, to create a moment from epoch should be: `moment.unix(Number)`.

src/javascript/app_2/App/Components/Elements/data_table.jsx
```diff
@@ -8,88 +8,102 @@ import React                          from 'react';
...
-                    }
-
-                    <tbody className='table-body'>
-                        {this.renderBodyRows()}
-                    </tbody>
-                </table>
+            <div className='table'>
+                <div className='table__head'>
+                    <div className='table__row' ref={el => { this.el_header_row = el; }}>
+                        {this.renderHeaders()}
```
binary-hamedanchi: How about using components instead of function calls? 🙂

PR #4142, opened 2018-07-26
negar/financial_auth

negarn:
add financial auth message for mt accounts


negarn:
this is replacing https://github.com/binary-com/binary-static/pull/3618/files, I couldn't deploy that to gh-pages for some reason.

PR #4144, opened 2018-07-26
[new-app] kelly/system_notifications

kellybinary:
added `<Routes />` component that listen to `has_error` and `error` from common_store. 
It renders `<ErrorComponent />` when `has_error` is set to true, otherwise renders` <BinaryRoutes />`


src/javascript/app_2/App/Containers/Routes/routes.jsx
```diff
@@ -0,0 +1,28 @@
+import { PropTypes as MobxPropTypes } from 'mobx-react';
+import PropTypes                      from 'prop-types';
+import React                          from 'react';
+import { withRouter }                 from 'react-router';
+import ErrorComponent                 from '../../Components/Elements/Errors';
+import BinaryRoutes                   from '../../Components/Routes';
+import { connect }                    from '../../../Stores/connect';
+
+const Routes = (props) => {
+    if (props.has_error) {
```
kellybinary: in the future, perhaps can add check for `is_loading` and load `<Loading />` here instead of adding it individually in each page

PR #4145, opened 2018-07-26
kelly/upload_PL_KID

kellybinary:


src/templates/static/legal/regulation.jsx
```diff
@@ -76,8 +76,9 @@ const Area = ({items}) => {
 };
 
 const Regulation = () => {
-    const has_KID = /de|en|es|it|fr|pl|pt|ru/.test(`${it.language.toLowerCase()}`);
-    const lang = has_KID ? `${it.language.toLowerCase()}` : 'en';
+    const has_KID    = /de|en|es|it|fr|pl|pt|ru/.test(`${it.language.toLowerCase()}`);
+    const has_CFD_FX = !/fr/.test(it.language.toLowerCase()); // CFD, FX documents in FR will not be available
+    const lang       = has_KID && has_CFD_FX ? `${it.language.toLowerCase()}` : 'en';
```
negarn: this means that CFD and FX documents will show in english for all the languages that don't have KID. is that intended?

kellybinary: yeap

PR #4146, opened 2018-07-27
[new-app] Contract Details

binary-hamedanchi:


src/javascript/app_2/Stores/Modules/Contract/Helpers/chart_markers.js
```diff
@@ -0,0 +1,91 @@
...
+
+    return createMarkerConfig(
+        MARKER_TYPES_CONFIG.LINE_START.type,
+        contract_info.date_start,
+    );
+}
+
+// -------------------- Spots --------------------
+function createMarkerSpotEntry(contract_info) {
+    if (!contract_info.entry_tick_time) return false;
```
negarn: @binary-hamedanchi do we want to update the condition here to handle sold before start?

src/javascript/app_2/App/Components/Elements/DataTable/table_row.jsx
```diff
@@ -1,31 +1,43 @@
...
-                            ? renderCellContent({ cell_value, col_index, row_obj, is_footer })
-                            : cell_value
-                    }
-                </TableCell>
-            );
-        })}
-    </div>
-);
+    return (
+        <div className='table__row' onClick={onClick}>
```
ruskakimov: maybe we use `RedirectOnClick` component here? [Like for portfolio drawer card](https://github.com/binary-com/binary-static/pull/4146/files#diff-9817baa1d2dd1ac4bd3ebff18bbe0decR9). This way I guess we don't need to pass any additional props.

src/javascript/app_2/Constants/ui.js
```diff
@@ -1,2 +1,20 @@
...
+    ORANGE      : '#e98024',
+    LIGHT_RED   : '#f44336',
+    LIGHT_GREEN : '#4caf50',
+    GREEN       : '#2e8836',
+    LIGHT_BLUE  : '#1565c0',
+    BLUE        : '#2a3052',
+    DARK_BLUE   : '#15212d',
+    LIGHT_BLACK : '#333',
+    BLACK       : '#000',
+};
```
ruskakimov: If these colors are just for icons, then I believe I found a better solution [in my PR](https://github.com/binary-com/binary-static/pull/4169). There we leave the styling to CSS and color constants are only in CSS. Here we would need to keep both sass constants and COLOR_CODES updated and in sync.

ruskakimov: In my PR CSS also doesn't need to know icon's inner structure.

src/javascript/app_2/App/Components/Elements/DataTable/table_row.jsx
binary-hamedanchi: The reason I didn't use it here was that `DataTable` is a common component, using `RedirectOnClick` here will prevent it from doing other things rather than redirecting (i.e. displaying a popup). So, I used a more general way for this to cover all needs.

src/javascript/app_2/Constants/ui.js
binary-hamedanchi: Sure, thanks.

PR #4147, opened 2018-07-30
[new-app] kelly/fix_header

kellybinary:


src/javascript/app_2/App/Components/Elements/InkBar/with_inkbar.jsx
```diff
@@ -0,0 +1,89 @@
...
+        componentDidUpdate() {
+            const active_el = this.node.querySelector('a[class="active"]');
+            if (active_el) {
+                this.updateInkbarPosition(active_el);
+            } else if (this.state.left !== 0 || this.state.width !== 0) {
+                this.clearInkBar(); // clear InkBar when active element doesn't exist
+            }
+        }
+
+        componentWillMount() {
```
morteza-binary: @kellybinary 
this lifecycle method is deprecated in the latest version of React, it's better to use the `constructor` instead.

src/javascript/app_2/App/Components/Layout/Header/menu_links.jsx
```diff
@@ -1,52 +1,28 @@
...
 import { BinaryLink } from '../../Routes';
 import Url            from '../../../../../_common/url';
 
-class MenuLinks extends React.PureComponent {
-    state = {
-        width: 0,
-        left : 0,
-    };
+// components wrapped with withInkBar has to be outside of render
+const DivWithInkBar = withInkBar((props) => <div {...props} />);
```
morteza-binary: @kellybinary 
is it possible to use other kinds of components instead of `div` or it's the allowable one?

src/javascript/app_2/App/Components/Elements/InkBar/with_inkbar.jsx
kellybinary: sure, will do 👍

src/javascript/app_2/App/Components/Layout/Header/menu_links.jsx
kellybinary: yes, it's possible to use other kinds of components as well

morteza-binary: so why isn't it passed directly to the `ComponentWithInkBar` as a prop? Does it cause rendering issues?

src/javascript/app_2/App/Components/Elements/InkBar/div_with_inkbar.jsx
```diff
@@ -0,0 +1,85 @@
+import classnames from 'classnames';
+import PropTypes  from 'prop-types';
+import React      from 'react';
+import { InkBar } from './inkbar.jsx';
+
+class DivWithInkBar extends React.Component {
```
negarn: I don't really like having `With` in the name, can we remove it, or change it to InkBarDiv instead?

PR #4155, opened 2018-08-01
[new-app] Portfolio Drawer Updates

ruskakimov:


src/sass/app_2/_common/layout/header.scss
```diff
@@ -101,16 +102,16 @@
                             overflow: hidden;
                             visibility: hidden;
                         }
-                        &.trade:before {
+                        &.menu-links__trade:before {
```
zakibinary: Just a heads up, I'll be changing all these external icons to inline SVG components for the dark theme card. So most of these icons will get removed.

PR #4156, opened 2018-08-01
[new-app] zaki/dark_theme

zakibinary:
Changelog
- Added `ThemeWrapper` for enabling dark theme layout and added the necessary dark theme subclasses for components. _(This is a temporary solution until we can come up with a more efficient way to theme our layout and components)_
- Refactored external assets and converted common and trade-types icons into inline SVG component.
- Updated and added color constants as per latest design style guide.
- Minor style updates and layout bug fixes.
- Removed Side Menu Drawer items temporarily until pages are ready.


zakibinary:
@morteza-binary , understood, I shared those concerns initially (was considering either svg spritesheets or using svg external altogether) but I think for less complex icons it should be fine for most part. You can refer to this article for a benchmarking test: http://mmckinnon.com/blog/2015/do-inline-svgs-weigh-down-websites

PR #4169, opened 2018-08-06
[new-app] Please login message

ruskakimov:
Icon styles are now in one place. Changing icon style is a simple as adding a class:
`<svg class="inline-icon disabled" ...>`
or using `@extend`
```
a:hover inline-icon {
  @extend .inline-icon.active;
}
```

So icons now will have a set of predefined states (`active`, `disabled`, etc.)
Adding new state [is done in one line](https://github.com/binary-com/binary-static/pull/4169/files#diff-5eff9d0625f17e9eac946263a6d96e3eR25). Themes can just override these states.

## Added `Localized` component
- [Use example](https://github.com/binary-com/binary-static/pull/4169/files#diff-bec613486dfb1abe758f4748440b1f4bR11)


src/javascript/app_2/App/Components/Routes/route_with_sub_routes.jsx
```diff
@@ -20,15 +21,17 @@ const RouteWithSubRoutes = route => {
...
-                        </div>
-                    </div>
-                    :
-                    <route.component {...props} routes={route.routes}/>
+                should_show_login_msg && route.keep_component &&
+                    <route.component {...props} routes={route.routes}>
+                        <PleaseLoginMessage onLogin={redirectToLogin} />
+                    </route.component>
+                ||
+                should_show_login_msg && <PleaseLoginMessage onLogin={redirectToLogin} />
```
ruskakimov: we can pass `route.path`  to determine the icon, but
- portfolio drawer has `/` route


or we can pass `route.component.displayName`, but
- `displayName` of route components start with `inject-`, so we would need to use regex
- we can add file in constants folder that would map names to icon components (not sure we would need it for something else though)
- Ashkan said we might replace it with a single icon for all pages, so it was all for naught

src/sass/app_2/_common/components/pagination.scss
```diff
@@ -1,75 +0,0 @@
...
-    & a {
-        text-decoration: none;
-        font-weight: 400;
-        font-size: 16px;
-
-        &::selection {
-            background-color: rgba(0, 0, 0, 0);
-        }
-    }
-}
```
ruskakimov: these styles aren't used anymore

src/javascript/app_2/App/Components/Elements/please_login_message.jsx
```diff
@@ -0,0 +1,31 @@
...
+                {localize('log in')}
+            </a>
+            &nbsp;
+            {localize('or')}
+            &nbsp;
+            <a href='javascript:;' onClick={onSignup}>
+                {localize('sign up')}
+            </a>
+            &nbsp;
+            {localize('to view this page.')}
```
ruskakimov: couldn't think of a solution, where we pass a single string template to localize.

negarn: why not use placeholders?

negarn: we already have this string `'Please [_1]log in[_2] or [_3]sign up[_2] to view this page.'`

src/javascript/app_2/App/Components/Elements/please_login_message.jsx
```diff
@@ -0,0 +1,35 @@
+import PropTypes    from 'prop-types';
+import React        from 'react';
+import { localize } from '../../../../_common/localize';
+
+const PleaseLoginMessage = ({ onLogin, onSignup, children }) => (
```
negarn: I don't think we need `please` in the name, `LoginMessage` is enough. same for the file name.

src/javascript/app_2/App/Components/Elements/please_login_message.jsx
ruskakimov: @negarn yeah, but how to insert achors then? In binary-static we replace `[_n]` with strings, if we do the same here, then we would need to use `dangerouslySetInnerHtml`. And it would still not bind `onClick` callback correctly.

negarn: true, maybe @morteza-binary or @binary-hamedanchi have some better ideas.

src/javascript/app_2/App/Components/Elements/please_login_message.jsx
ruskakimov: i just though `LoginMessage` could be a bit ambiguous 😅

negarn: why? it's a message for logging in

zakibinary: TBH, this is the first time I've seen someone declare a constant in such polite fashion. ;)

src/javascript/app_2/App/Components/Elements/please_login_message.jsx
binary-hamedanchi: Need a component/function to handle this. Breaking a sentence into several parts and translate them separately will lose the point in many languages.

PR #4179, opened 2018-08-08
negar/remove_japan

negarn:


ashkanx:
@negarn conflicts

negarn:
@ashkanx it will have the `ja_JP.po` file conflict whenever we update translations because this PR removes that file, so I would suggest resolving that right before merging this PR instead of after every translation update.

PR #4181, opened 2018-08-08
negar/fix_withdrawal_email

negarn:
don't send multiple withdrawal requests in the same session

### DIFF
--- a/src/javascript/app/pages/cashier/deposit_withdraw.js
+++ b/src/javascript/app/pages/cashier/deposit_withdraw.js
@@ -16,6 +16,8 @@ const template          = require('../../../_common/utility').template;
 const DepositWithdraw = (() => {
     const default_iframe_height = 700;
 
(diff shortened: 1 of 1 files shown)


src/javascript/app/pages/cashier/deposit_withdraw.js
```diff
@@ -46,23 +48,32 @@ const DepositWithdraw = (() => {
...
-                verify_email: Client.get('email'),
-                type        : 'payment_withdraw',
-            }).then((response_withdraw) => {
-                if ('error' in response_withdraw) {
-                    showError('custom_error', response_withdraw.error.message);
-                } else {
-                    showMessage('check_email_message');
-                }
-            });
+            if (response_withdrawal.length) {
```
kellybinary: perhaps Object.keys(response_withdrawal).length?

PR #4183, opened 2018-08-08
ashkan/update binary-in-numbers

ashkanx:


src/templates/static/about/binary_in_numbers.jsx
```diff
@@ -45,7 +45,7 @@ const BinaryInNumbers = () => (
             <Numbers
                 className='employee'
                 header={it.L('Number of employees and contractors')}
-                text={it.L('We’ve grown in size over the years –– both in terms of manpower and offices. We currently have over 150 employees across four offices in Malaysia, Malta, and Japan. Further growth in headcount is expected this year to cater to the increased demand in our products and services.')}
+                text={it.L('We’ve grown in size over the years –– both in terms of manpower and offices. We currently have over 200 employees across three offices in Malaysia and Malta. Further growth in headcount is expected this year to cater to the increased demand in our products and services.')}
```
negarn: I would suggest a placeholder for numbers (at least the `200`, not feasible for `three`) so translation doesn't break as we may regularly update this

src/templates/static/about/index.jsx
```diff
@@ -70,7 +70,7 @@ const Index = () => (
                         </Box>
                         <Box>
                             <BoxInner className='border-right-bottom gr-padding-30' image='license' text={it.L('Licensed and regulated in Malta, the United Kingdom, the Isle of Man, Ireland, and Japan')} />
-                            <BoxInner className='gr-padding-30' image='languages' text={it.L('Published in English, Indonesian, Japanese, Chinese, Polish, German, French, Portuguese, Russian, and Thai')} />
+                            <BoxInner className='gr-padding-30' image='languages' text={it.L('Published in English, Indonesian, Chinese, Polish, German, French, Portuguese, Russian, and Thai')} />
```
negarn: should we add spanish?

src/templates/static/affiliates/signup.jsx
```diff
@@ -19,7 +19,7 @@ const Signup = () => (
             <div className='gr-padding-20'>
                 <h2 className='center-text margin-bottom-50'>{it.L('How it works')}</h2>
                 <div className='steps'>
-                    <Step circle_no='1' header={it.L('Sign up')}                       text={it.L('Getting started is free and easy – just <a href=\'[_1]\'>fill out the application form</a>, choose from two types of commission plans and wait for our approval.', it.affiliate_signup_url)} />
+                    <Step circle_no='1' header={it.L('Sign up')}                       text={it.L('Getting started is free and easy – just <a href=\'[_1]\' target="_blank">fill out the application form</a>, choose from two types of commission plans and wait for our approval.', it.affiliate_signup_url)} />
```
negarn: I would suggest putting the entire anchor into a placeholder, up to you

PR #4184, opened 2018-08-08
ruskakimov/password_field_limit

ruskakimov:


src/templates/_common/components/forms.jsx
```diff
@@ -35,15 +35,15 @@ export const FormRow = ({
...
+        if (['text', 'password', 'number', 'checkbox'].includes(type)) {
             return (
                 <React.Fragment>
                     {input_prefix}
                     <input
                         type={type}
                         className={className}
                         id={id}
-                        maxLength={type === 'password' ? 25 : undefined}
+                        maxLength={type === 'password' ? 26 : undefined}
```
negarn: if other input fields don't have a maxlength then it makes sense to me to remove it for password too, I see your point now

PR #4197, opened 2018-08-10
[new-app] morteza/pages_title

morteza-binary:


src/javascript/app_2/App/Components/Routes/route_with_sub_routes.jsx
```diff
@@ -32,6 +32,7 @@ const RouteWithSubRoutes = route => {
             );
         }
 
+        document.title = `${localize(route.title)} | Binary.com`;
```
negarn: do we want to use a placeholder or directly use `Binary.com`?

morteza-binary: @negarn 
it's possible to use placeholders but I don't exactly whether we need it or not?
@binary-hamedanchi @ashkanx What's your idea?

binary-hamedanchi: We can just move it to a constant.

@morteza-binary Do we need to handle removing ` | ` separator in case of `route.title` is unavailable?

src/javascript/app_2/App/Components/Routes/route_with_sub_routes.jsx
```diff
@@ -32,7 +33,8 @@ const RouteWithSubRoutes = route => {
             );
         }
 
-        document.title = `${localize(route.title)} | Binary.com`;
+        const title = route.title ? `${ localize(route.title) } | ` : '';
+        document.title = `${ title }${default_title}`;
```
negarn: we should either have space or no space after `{` for both variables to be consistent

src/javascript/app_2/App/Constants/app_config.js
```diff
@@ -0,0 +1 @@
+export const default_title = 'Binary.com';
```
negarn: maybe a different name? it's our website's name, not just a title

PR #4221, opened 2018-08-15
ruskakimov/proposal_display_fix

ruskakimov:


src/javascript/app/pages/trade/price.js
```diff
@@ -355,6 +345,7 @@ const Price = (() => {
 
         processForgetProposalOpenContract();
         processForgetProposals().then(() => {
+            $('.price_container').hide();
```
negarn: it looks strange that it keeps disappearing and reappearing, isn't there any other way than making it display none all the time?

ruskakimov: @negarn It doesn't really reappear, it just persists from previous contract type. Here I hide all price containers on contract type change, then `display()` shows the necessary ones.

src/javascript/app/pages/trade/price.js
```diff
@@ -355,16 +345,38 @@ const Price = (() => {
...
+            const setPriceContainersVisibility = () => {
+                Object.keys(position_is_visible).forEach(position => {
+                    const container = CommonFunctions.getElementById(`price_container_${position}`);
+                    if (position_is_visible[position]) {
+                        $(container).fadeIn(0);
+                    } else {
+                        $(container).fadeOut(0);
+                    }
+                });
+            };
```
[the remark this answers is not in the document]
ruskakimov: @negarn ready now :)

src/javascript/app/pages/trade/price.js
```diff
@@ -355,21 +345,44 @@ const Price = (() => {
 
         processForgetProposalOpenContract();
         processForgetProposals().then(() => {
+            const position_is_visible = {
+                top   : false,
+                middle: false,
+                bottom: false,
+            };
+            let first_price_proposal = true;
             Object.keys(types || {}).forEach((type_of_contract) => {
```
negarn: can't we use the index in foreach to detect first price proposal instead?

ruskakimov: @negarn price proposal is a subscription call here

ruskakimov: what i mean by `first_price_proposal` is the very first response we get from `proposal: 1`

PR #4233, opened 2018-08-17
[new-app]Morteza/url_parameters

morteza-binary:


src/javascript/app_2/Modules/Trading/Containers/trade.jsx
```diff
@@ -10,9 +10,14 @@ import { getPropertyValue } from '../../../../_common/utility';
 
 class Trade extends React.Component {
     componentDidMount() {
+        this.props.tradeComponentDidMount();
         this.props.updateQueryString();
```
binary-hamedanchi: How about having only two `onMount()` and `onUnmount()` in store and make these calls there? It'll move the logic to store and also maintains consistency.

PR #4235, opened 2018-08-17
kelly/footer_update

kellybinary:
- show CFD warnings for EU countries


src/javascript/app/base/footer.js
```diff
@@ -1,4 +1,23 @@
+const BinarySocket = require('./socket');
+const Client       = require('../base/client');
+const State        = require('../../_common/storage').State;
+
 const Footer = (() => {
+    const onLoad = () => {
+        BinarySocket.wait('website_status', 'authorize').then(() => {
+            const $footer      = $('#footer-last');
+            const $cfd_warning = $footer.find('.eu-only');
+            const show_warning = Client.get('is_eu') || /maltainvest/.test(State.getResponse('authorize.landing_company_name'));
```
negarn: do we want to check is_eu for logged out else check maltainvest for logged in? currently this will show for virtual and MLT clients too, not sure if we want that or not. could check with compliance.

src/javascript/app/base/footer.js
```diff
@@ -1,4 +1,29 @@
...
+
 const Footer = (() => {
+    const onLoad = () => {
+        BinarySocket.wait('website_status', 'authorize').then(() => {
+            const $footer      = $('#footer-last');
+            const $cfd_warning = $footer.find('.eu-only');
+            const show_warning = (Client.isLoggedIn() ?
+                (/maltainvest/.test(State.getResponse('authorize.landing_company_name'))
+                    || /financial/.test(Client.getBasicUpgradeInfo().can_upgrade_to)
+                    || Client.hasAccountType('financial'))
```
negarn: can we check has account type before checking can upgrade to financial? it takes less processing so it's better to check that first

kellybinary: yes, sure 👍

PR #4244, opened 2018-08-20
[new-app] Morteza/url_parameters

morteza-binary:


src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
```diff
@@ -74,3 +74,34 @@ const createProposalRequestForContract = (store, type_of_contract) => {
         ),
     };
 };
+
+export const getProposalParametersName = (proposal_requests, alternatives, removable_properties) => {
+    const is_digit = Object.keys(proposal_requests)
```
binary-hamedanchi: Can we call [`isDigitContract()`](https://github.com/binary-com/binary-static/blob/new-app/src/javascript/app_2/Stores/Modules/Contract/Helpers/digits.js#L1) instead?

src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
```diff
@@ -74,3 +74,34 @@ const createProposalRequestForContract = (store, type_of_contract) => {
...
+        .findIndex(i => i.toUpperCase().indexOf('DIGIT') > -1) > -1;
+
+    const keys = Object.keys(proposal_requests[Object.keys(proposal_requests)[0]]);
```
binary-hamedanchi: This probably does the same:
```js
const keys = Object.keys(Object.values(proposal_requests)[0]);
```

src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
```diff
@@ -74,3 +74,34 @@ const createProposalRequestForContract = (store, type_of_contract) => {
         ),
     };
 };
+
+export const getProposalParametersName = (proposal_requests, alternatives, removable_properties) => {
```
binary-hamedanchi: I believe it's possible to simplify the whole function. Let's check this together.

src/javascript/app_2/Stores/Modules/Trading/trade_store.js
```diff
@@ -231,6 +239,19 @@ export default class TradeStore extends BaseStore {
     requestProposal() {
         const requests = createProposalRequests(this);
         if (!isEmptyObject(requests)) {
+            const proper_proposal_params_for_query_string = getProposalParametersName(
+                requests,
+                proposal_properties_alternative_names,
+                removable_proposal_properties
```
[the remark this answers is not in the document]
morteza-binary: @binary-hamedanchi 
it's possible, but as the `proposal.js` is a helper function it's better, not depend on some hardcoded stuff.

src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
morteza-binary: @binary-hamedanchi 
IMO it's not appropriate to use it since `proposal.js` and `digits.js` are in different modules unless we move the `digit.js` in a more common directory.

src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
morteza-binary: @binary-hamedanchi 
Sure let talk about it whenever you're free.

PR #4245, opened 2018-08-20
kelly/fix_status_check

kellybinary:
- fix check for cashier_locked / withdrawal_locked statuses

### DIFF
--- a/src/javascript/app/base/header.js
+++ b/src/javascript/app/base/header.js
@@ -283,8 +283,7 @@ const Header = (() => {
             };
 
             const buildMessage = (string, path, hash = '') => localize(string, [`<a href="${Url.urlFor(path)}${hash}">`, '</a>']);
-            const checkCashierStatus = (string) =>
-                status.findIndex(s => s === string) < 0 ? Boolean(false) : Boolean(true);
+            const hasStatus = (string) => status.findIndex(s => s === string) < 0 ? Boolean(false) : Boolean(true);
 
             const messages = {
                 authenticate         : () => buildMessage('[_1]Authenticate your account[_2] now to take full advantage of all payment methods available.',                                      'user/authenticate'),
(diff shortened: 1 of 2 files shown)


src/javascript/app/base/header.js
```diff
@@ -303,7 +303,7 @@ const Header = (() => {
 
             const validations = {
                 authenticate         : () => +get_account_status.prompt_client_to_authenticate,
-                cashier_locked       : () => /cashier_locked/.test(status),
+                cashier_locked       : () => /^cashier_locked$/.test(status),
```
negarn: this doesn't work unless the only status in the array is `cashier_locked`. usually we will have multiple items in the status array.

src/javascript/app/base/header.js
```diff
@@ -315,7 +315,7 @@ const Header = (() => {
                 tax                  : () => Client.shouldCompleteTax(),
                 tnc                  : () => Client.shouldAcceptTnc(),
                 unwelcome            : () => /unwelcome/.test(status),
-                withdrawal_locked    : () => /withdrawal_locked/.test(status),
+                withdrawal_locked    : () => /^withdrawal_locked$/.test(status),
```
negarn: same here.

src/javascript/app/base/header.js
```diff
@@ -299,11 +300,12 @@ const Header = (() => {
                 tnc                  : () => buildMessage('Please [_1]accept the updated Terms and Conditions[_2] to lift your withdrawal and trading limits.',                                  'user/tnc_approvalws'),
                 unwelcome            : () => buildMessage('Trading and deposits have been disabled on your account. Kindly [_1]contact customer support[_2] for assistance.',                    'contact'),
                 withdrawal_locked    : () => localize('Withdrawals have been disabled on your account. Please check your email for more details.'),
+                mt5_withdrawal_locked: () => localize('MT5 withdrawals have been disabled on your account. Please check your email for more details.'),
```
negarn: this string needs to be added to translations

src/javascript/app/base/header.js
```diff
@@ -299,11 +301,12 @@ const Header = (() => {
...
             };
 
             const validations = {
                 authenticate         : () => +get_account_status.prompt_client_to_authenticate,
-                cashier_locked       : () => /cashier_locked/.test(status),
+                cashier_locked       : () => checkCashierStatus('cashier_locked'),
                 currency             : () => !Client.get('currency'),
                 document_needs_action: () => /document_needs_action/.test(status),
```
negarn: perhaps we can update all of these to use the same function, and rename it to something more general, like `hasStatus`

kellybinary: sure, good idea :)

src/javascript/app/base/header.js
```diff
@@ -283,6 +283,7 @@ const Header = (() => {
             };
 
             const buildMessage = (string, path, hash = '') => localize(string, [`<a href="${Url.urlFor(path)}${hash}">`, '</a>']);
+            const hasStatus = (string) => status.findIndex(s => s === string) < 0 ? Boolean(false) : Boolean(true);
```
negarn: can also be used on line 368

PR #4247, opened 2018-08-21
[new-app] mohammad/upgrade_smartcharts

binary-hamedanchi:
- Upgrade to v0.2.8
- Use chart id


src/javascript/app_2/Modules/Trading/Containers/trade.jsx
```diff
@@ -27,6 +27,7 @@ class Trade extends React.Component {
                 <div className='chart-container notice-msg'>
                     { this.props.symbol &&
                         <SmartChart
+                            chart_id={1}
```
morteza-binary: @binary-hamedanchi 
Probably it's better to use a constant instead of hardcoding the id or use a random unique numbers e.g. uuid or guid

binary-hamedanchi: @morteza-binary It shouldn't be a random number, but has to be a unique number for trade page. So, every time clients visit the trade page, chart loads the latest settings they set.

Other pages (Statement/Portfolio) don't have an `id`, that causes to display a clean chart every time and prevent storing any settings there.

https://github.com/binary-com/SmartCharts#api

morteza-binary: @binary-hamedanchi 
I see, but IMHO, it's better to save the id in `trade_store` or in a `constant` variable in `constant` file. By keeping the id in `trade_store` we can generate it randomly for each user and save it in the `localStorage` when the app is run for the first time.

BTW, it's just an idea to prevent id collision in future if we decide to add id for some other `smartchart` instances

PR #4248, opened 2018-08-21
[new-app] zaki/quick_portfolio_animation_fix

zakibinary:
Changelog
- Improved performance of quick portfolio drawer transition.


src/sass/app_2/_common/layout/layouts.scss
```diff
@@ -4,29 +4,76 @@
...
+                        transform: translateX(#{$PORTFOLIO_DRAWER_WIDTH}) !important;
+                    }
+                    .info-box {
+                        transform: translateX(-#{$PORTFOLIO_DRAWER_WIDTH}) !important;
                     }
-                    .sidebar-container, .contract-wrapper {
-                        grid-column: col-start 10 / span 3;
+                }
+            }
+            .portfolio, .statement {
```
binary-hamedanchi: Do we have to update this (and another one on [L#74](https://github.com/binary-com/binary-static/pull/4248/files#diff-d482fe7751bf830c863c2f96d09eacd9R74)) every time adding a new module in the future?

src/sass/app_2/_common/layout/layouts.scss
```diff
@@ -4,29 +4,76 @@
...
-            @media screen and (max-width: 980px) {
-                .trade-container {
-                    .chart-container {
-                        grid-column: col-start 1 / span 8;
+            .ciq-chart {
+                .cq-top-ui-widgets {
+                    transition: transform 0.25s ease;
+
+                    .info-box {
+                        transition: transform 0.25s ease;
```
binary-hamedanchi: Can get merged with the above one:
```css
&, .info-box {
    transition: transform 0.25s ease;
}
```

src/sass/app_2/_common/layout/layouts.scss
zakibinary: Yes, I assume for cashier and others we'll add this as well, maybe we can unify them with a more common class, so that instead of using `margin`, we use `translate` and `max-width` to make the animation consistent with the `trade` app layout.

PR #4253, opened 2018-08-21
[new-app] Quick portfolio refreshes after contract purchase

ruskakimov:


src/javascript/app_2/Stores/Modules/Portfolio/portfolio_store.js
```diff
@@ -30,71 +31,77 @@ export default class PortfolioStore extends BaseStore {
...
+        const { contract_id, action: act } = response.transaction;
+
+        if (act === 'buy') {
+            WS.portfolio().then((res) => {
+                const new_pos = res.portfolio.contracts.find(pos => +pos.contract_id === +contract_id);
+                if (!new_pos) return;
+                this.pushNewPosition(new_pos);
+            });
+            // subscribe to new contract:
+            WS.subscribeProposalOpenContract(contract_id, this.proposalOpenContractHandler, false);
```
binary-hamedanchi: Subscribing to a particular `contract_id` causes a bug here, unless it handles forgetting it on contract expiry.
2nd option: It can be handled in Subscription Manager by removing the subscription when it receives a response without `id` (and not the first response for some calls!)

ruskakimov: @binary-hamedanchi by bug, you mean that expired contracts remain in `subscriptions`?

ruskakimov: @binary-hamedanchi i have gone with an option 2 :)

PR #4268, opened 2018-08-27
kelly/mf_retail

kellybinary:
- since MF retail now can deposit but not allowed to trade, we need to show `Trading has been disabled..` banner for MF retail clients.


src/javascript/app/base/header.js
```diff
@@ -311,14 +312,15 @@ const Header = (() => {
                 document_review      : () => hasStatus('document_under_review'),
                 excluded_until       : () => Client.get('excluded_until'),
                 financial_limit      : () => hasStatus('ukrts_max_turnover_limit_not_set'),
+                mf_retail            : () => State.getResponse('authorize.landing_company_name') === 'maltainvest' && !hasStatus('professional'),
```
negarn: we can also use `Client.get('landing_company_shortcode')`

src/javascript/app/base/header.js
```diff
@@ -347,7 +350,10 @@ const Header = (() => {
             const checkStatus = (check_statuses) => {
                 const notified = check_statuses.some((check_type) => {
                     if (validations[check_type]()) {
-                        displayNotification(messages[check_type]());
+                        if (check_type === 'mf_retail' && !/trading/.test(window.location.href)) {
```
negarn: better to check `State.get('is_trading')`

negarn: if you also need ladders we have a similar `is_mb_trading` for that

src/javascript/app/pages/mb_trade/mb_tradepage.js
```diff
@@ -17,6 +18,7 @@ const State               = require('../../../_common/storage').State;
 const urlFor              = require('../../../_common/url').urlFor;
 const findParent          = require('../../../_common/utility').findParent;
 
+
```
negarn: extra space?

src/javascript/app/base/header.js
```diff
@@ -347,7 +350,10 @@ const Header = (() => {
             const checkStatus = (check_statuses) => {
                 const notified = check_statuses.some((check_type) => {
                     if (validations[check_type]()) {
-                        displayNotification(messages[check_type]());
+                        if (check_type === 'mf_retail' && !(State.get('is_trading') || State.get('is_mb_trading'))) {
```
negarn: would be good to include a comment to explain this

src/javascript/app/pages/mb_trade/mb_tradepage.js
```diff
@@ -32,6 +33,7 @@ const MBTradePage = (() => {
     };
 
     const init = () => {
+        Header.displayAccountStatus();
```
negarn: do we need this? it should ideally have updated the notification in header before needing to call it here

PR #4273, opened 2018-08-28
zaki/career_location_values

zakibinary:
Changelog
- Added values section `careers` page.
- Added values section to `about-us` page.
- Added career locations pages.


src/templates/static/about/careers.jsx
```diff
@@ -44,12 +44,26 @@ const ColumnSM = ({ gr, header, paragraph }) => (
...
+const ColumnLG = ({ center, image, text, link }) => (
+    <div className={`gr-4 gr-12-m${center ? ' center-text' : ''}`}>
+        <div>
+            <p className='column-margin'>{text}</p>
+        </div>
         <img className='responsive' src={it.url_for(`images/pages/careers/${image}.jpg`)} />
-        <div className='white-bg-color'>
-            <p className='column-margin fill-text'>{text}</p>
+        <div className='fill-text'>
+            <a href={link}><span>{it.L('Learn more')}</span></a>
```
kellybinary: perhaps can add a check if link exists then return `<a/>` else return `<p/>`

zakibinary: @kellybinary , is this better? Since if we don't have `link` prop it's pointless to return anything else in place.

```{!!link &&
            <div className='fill-text'>
                <a href={link}><span>{it.L('Learn more')}</span></a>
            </div>
        }```

Comment (kellybinary) [reply src/templates/static/about/careers.jsx]: ahh right, but what about the `Learn more` text? shouldn't it be `{it.L(text)}` instead?

Comment (zakibinary) [reply src/templates/static/about/careers.jsx]: `text` prop is for the header on top of the image. The `button link` will always use `Learn More` in this use case.

File: src/images/pages/careers/ic-competence.svg [hunk 2]
```diff
@@ -0,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="60" height="60" viewBox="0 0 60 60">
```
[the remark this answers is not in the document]
zakibinary: Sure. 👍

PR #4283, opened 2018-08-28
negar/maltainvest_mt_acct_comparison

negarn:
show different types to maltainvest

### DIFF
--- a/src/javascript/app/pages/user/metatrader/types_of_accounts.js
+++ b/src/javascript/app/pages/user/metatrader/types_of_accounts.js
@@ -1,14 +1,15 @@
 const Client       = require('../../../base/client');
 const BinarySocket = require('../../../base/socket');
 const Scroll       = require('../../../../_common/scroll');
+const State        = require('../../../../_common/storage').State;
 
 const TypesOfAccounts = (() => {
     const onLoad = () => {
         Scroll.goToHashSection();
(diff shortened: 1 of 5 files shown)


src/javascript/app/pages/user/metatrader/types_of_accounts.js
```diff
@@ -0,0 +1,31 @@
...
+const Scroll       = require('../../../../_common/scroll');
+
+const TypesOfAccounts = (() => {
+    const onLoad = () => {
+        Scroll.goToHashSection();
+
+        if (Client.isLoggedIn()) {
+            BinarySocket.wait('landing_company').then((response) => {
+                const should_show_maltainvest_content = response.landing_company.financial_company.shortcode === 'maltainvest'; // for VRTC, MLT, or MF
+                $('.hide-maltainvest').setVisibility(!should_show_maltainvest_content);
```
kellybinary: nitpick: we can wait for `authorize`, then use `Client.get('landing_company_shortcode')` as well?

negarn: no, because we want it for VRTC and MLT as well and their landing company shortcode would be `virtual` and `malta`, not `maltainvest`. we want to show it to any client that has or can open MF

PR #4287, opened 2018-08-29
kelly/fixes

kellybinary:
- add check for maximum Lookback multiplier
- add Call/Put Spreads in Get Started
- fixes for mobile


src/javascript/app/pages/trade/price.js
```diff
@@ -61,6 +61,11 @@ const Price = (() => {
 
         if (multiplier && CommonFunctions.isVisible(multiplier) && multiplier.value) {
             proposal.amount = multiplier.value;
+            if (multiplier.value > 1000) {
+                proposal.error = {
+                    message: 'Maximum multiplier of 1000.',
```
negarn: this is called with localize() somewhere?

PR #4299, opened 2018-09-03
[new-app] zaki/accounts_header_popup

zakibinary:
Changelog
- Changed accounts switcher to dialog modal.
- Removed old accounts switcher in menu drawer.
- Moved logout button into accounts switcher modal.
- Updated `smartcharts` library to `0.2.9`.


src/javascript/app_2/App/Components/Elements/account_switcher.jsx
```diff
@@ -14,12 +16,11 @@ const getAccountInfo = (loginid) => {
         loginid,
         is_virtual,
         icon : account_type.toLowerCase(), // TODO: display the icon
-        title: localize('[_1] Account', [account_type]),
+        title: account_type.toLowerCase() === 'virtual' ? localize('demo') : account_type,
```
binary-hamedanchi: Can we localize the exact word we're going to display (`DEMO`)? Because using CSS to capitalize text will result in inconsistency and issues in different languages.

src/javascript/app_2/App/Components/Layout/Header/account_info.jsx
```diff
@@ -1,12 +1,32 @@
...
 
-const AccountInfo = ({ balance, currency, loginid }) => (
+const AccountInfo = ({ balance, currency, loginid, is_dialog_on, toggleDialog }) => (
     <div className='acc-balance'>
-        <div className='acc-info'>
-            <p className='acc-balance-currency'>{`${(currency || '').toUpperCase()} ${localize('Account')}`}</p>
-            <p className='acc-balance-accountid'>{loginid}</p>
+        <div className='acc-switcher-container'>
+            <div className={classNames('acc-info', { 'show': is_dialog_on })} onClick={toggleDialog}>
+                <p className='acc-balance-currency'>{`${(currency || '').toUpperCase()} ${localize('Account')}`}</p>
```
binary-hamedanchi: 1. To achieve a proper translation, better to not take apart a phrase.
So it can be like `localize('[_1] Account', [...])`

2. According to the description on the card, we'll display the account type here instead of currency. (the class name seems inappropriate too)

zakibinary: @binary-hamedanchi , regarding 1) I assume something like this would be fine? `{localize('[_1] Account', [(currency || '').toUpperCase()])}`

PR #4331, opened 2018-09-12
kelly/change_OTC_Indices_to_Indices

kellybinary:
- change OTC Indices to Indices


negarn:
probably our active symbols test will start failing, but we can update it after BE releases

ashkanx:
I think BE released the changed

negarn:
@kellybinary then can you update the test as well please?

kellybinary:
@negarn which test are we talking about?

negarn:
https://github.com/binary-com/binary-static/blob/master/src/javascript/app/common/__tests__/active_symbols.js#L9

PR #4359, opened 2018-09-20
kelly/add_1.1.1.1

kellybinary:


src/templates/app/user/security/cloudflare_dns.jsx
```diff
@@ -0,0 +1,30 @@
...
+                <img className='responsive' src={it.url_for('images/pages/settings/connect-1111.svg')} />
+            </div>
+        </div>
+        <div className='gr-row'>
+            <div className='gr-12 gr-padding-10'>
+                <h2>{it.L('What is DNS')}</h2>
+                <p>{it.L('The Domain Name System (DNS) allows you to access a website through a user-friendly web address such as [_1], instead of a complicated IP address.', `<a href="${it.url_for('/')}">www.binary.com</a>`)}</p>
+            </div>
+            <div className='gr-12 gr-padding-10'>
+                <h2>{it.L('Why you should use 1.1.1.1 instead of your ISP\'s default DNS service')}</h2>
```
negarn: if we're not translating 1.1.1.1 in other strings, we should also use a placeholder to not translate it here

src/templates/app/user/security/cloudflare_dns.jsx
```diff
@@ -0,0 +1,30 @@
...
+                <p>{it.L('The default DNS services provided by your ISP can be slow and insecure. [_1] recommends using Cloudflare\'s 1.1.1.1 DNS service that gives you a faster and more private way to browse the Internet.', it.website_name)}</p>
```
negarn: same here

PR #4375, opened 2018-09-25
Add Missing languages label

easteregg:


src/templates/static/about/index.jsx
```diff
@@ -128,7 +128,7 @@ const Index = () => (
                         </Box>
                         <Box>
                             <BoxInner className='border-right-bottom gr-padding-30' image='license' text={it.L('Licensed and regulated in Malta, the United Kingdom, the Isle of Man, and Ireland')} />
-                            <BoxInner className='gr-padding-30' image='languages' text={it.L('Published in English, Indonesian, Chinese, Polish, German, French, Spanish, Portuguese, Russian, and Thai')} />
+                            <BoxInner className='gr-padding-30' image='languages' text={it.L('Published in English, Indonesian, Chinese, Polish, German, French, Spanish, Portuguese, Russian, Italian, Vietnamese and Thai')} />
```
negarn: can we change the order to follow the same order as the drop-down? (it should be alphabetical) then it will be also easier to see what is missing in the future

PR #4377, opened 2018-09-25
kelly/mt5_redirect

kellybinary:


src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -228,8 +228,10 @@ const PersonalDetails = (() => {
                     return;
                 }
                 const get_settings    = data.get_settings;
-                const has_required_mt = get_settings.tax_residence && get_settings.tax_identification_number &&
-                    get_settings.citizen;
+                const has_required_mt = (Client.get('landing_company_shortcode') === 'costarica' ?
```
negarn: isn't it better to check with the new field `landing_company.config.tax_details_required` instead of CR?

kellybinary: sure, it it only for mt accounts?

negarn: that's my understanding. you can check with the QA team

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -229,8 +229,10 @@ const PersonalDetails = (() => {
                     return;
                 }
                 const get_settings    = data.get_settings;
-                const has_required_mt = get_settings.tax_residence && get_settings.tax_identification_number &&
-                    get_settings.citizen;
+                const has_required_mt = (State.getResponse('landing_company.config.tax_details_required') !== 1 ?
```
negarn: can we check `=== 1`?

negarn: also remember the `+` before `State` to turn the value into a number

PR #4379, opened 2018-09-25
easteregg/descriptive_date_label_on_mobile_devices

easteregg:


src/javascript/app/pages/resources/trading_times/trading_times.ui.js
```diff
@@ -37,6 +37,18 @@ const TradingTimesUI = (() => {
             maxDate : 364,
         });
         $date.val(localize('Today'));
+        // Create a label to be friendlier
+        const $label = $('label[for=trading-date]');
+        $label.append('<span class=\'ux-date foot-note\'></span>');
+        if ($date.val() === '') {
+            $('span.ux-date').text(localize('Today'));
+            $date.val(moment().format('YYYY-MM-DD'));
```
negarn: have a look at the function `toISOFormat()`, it's used so that we don't have to hardcode the format everywhere

src/javascript/app/pages/resources/trading_times/trading_times.ui.js
```diff
@@ -37,6 +37,18 @@ const TradingTimesUI = (() => {
...
+            $date.attr('value', moment.utc().toISOString().substring(0,10));
```
negarn: if the value is going through `.utc()` then val above and below should also be done with `.utc()`.

src/javascript/app/pages/resources/trading_times/trading_times.ui.js
```diff
@@ -30,13 +30,26 @@ const TradingTimesUI = (() => {
...
+        $date.attr('data-value', isoFormattedDate);
         DatePicker.init({
             selector: '#trading-date',
             minDate : 0,
             maxDate : 364,
         });
         $date.val(localize('Today'));
+        // Create a label to be friendlier
+        const $label = $('label[for=trading-date]');
+        $label.append('<span class=\'ux-date foot-note\'></span>');
```
negarn: can you create the span using jquery element creation? it's more common within our codebase

src/javascript/app/pages/resources/trading_times/trading_times.ui.js
```diff
@@ -30,13 +30,26 @@ const TradingTimesUI = (() => {
...
+        if ($date.val() === '') {
```
negarn: how about using `!$date.val()` instead to handle other cases?

src/javascript/app/pages/resources/trading_times/trading_times.ui.js
```diff
@@ -30,13 +30,26 @@ const TradingTimesUI = (() => {
...
+            $('span.ux-date').text(localize('Today'));
+            $date.val(isoFormattedDate);
+            $date.attr('value', isoFormattedDate);
+        }
+        $date.change(() => {
+            $('span.ux-date').text(moment($date.val()).format('LL'));
```
negarn: do we need `.utc` here too?

src/javascript/app/pages/resources/trading_times/trading_times.ui.js
easteregg: Sure, Right away :)

src/javascript/app/pages/resources/trading_times/trading_times.ui.js
easteregg: I couldn't quite catch that, you mean line 50?

negarn: yes, something like `moment.utc($date.val()`

easteregg: I changed my time zone drastically to see if it changes. It didn't. I think because we are using a date field, changing time doesn't affect its behavior. So the only time that this can come to play is when we are initiating a date field, and that is passed through a UTC filter.

negarn: as discussed and tested with @binary-hamedanchi it seems you're right and it only affects things like `.valueOf()` which need the timezone, no change needed then

src/javascript/app/pages/resources/trading_times/trading_times.ui.js
```diff
@@ -30,13 +30,28 @@ const TradingTimesUI = (() => {
...
             selector: '#trading-date',
             minDate : 0,
             maxDate : 364,
         });
         $date.val(localize('Today'));
+        // Create a label to be friendlier
+        const $label = $('label[for=trading-date]');
+        $('<span/>')
+            .class('ux-date foot-note')
+            .appendTo($label);
```
negarn: why not something like `$label.append($('<span/>', { class: 'ux-date foot-note' });`?

PR #4391, opened 2018-09-27
negar/fix_acc_transfer_decimals

negarn:
fix decimal places allowed and min withdrawal amount for dai & tether


jy-binary:
please check conflicts

PR #4393, opened 2018-09-27
ruskakimov/remove_browser_support_page

ruskakimov:


scripts/sitemap.js
```diff
@@ -35,7 +35,6 @@ const urls = [
     ['regulation',                         'monthly', 0.80, 'id'],
     ['responsible-trading',                'monthly', 0.80],
     ['terms-and-conditions',               'monthly', 0.80],
-    // ['user/browser-support',               'monthly', 0.80],
```
negarn: if we're deleting it then please remove it rather than commenting. if we might return it, then we should comment the line in pages.js as well

PR #4404, opened 2018-10-01
[new-app] kelly/fix_incorrect_endtime_bought

kellybinary:
- use server_time if selected start time is 'now'


src/javascript/app_2/Stores/Modules/Trading/trade_store.js
```diff
@@ -259,6 +259,7 @@ export default class TradeStore extends BaseStore {
 
     @action.bound
     requestProposal() {
+        this.server_time = this.root_store.common.server_time; // pass server_time from common_store
```
binary-hamedanchi: This causes updating `this.server_time` _only_ when `requestProposal()` being called. Better to access it directly where needed.

PR #4408, opened 2018-10-01
show_os_specific_download_link

easteregg:


src/javascript/app/base/binary_pages.js
```diff
@@ -1,9 +1,10 @@
 // ==================== _common ====================
-const TabSelector = require('../../_common/tab_selector'); // eslint-disable-line import/order
+// eslint-disable-line import/order
 
 // ==================== app ====================
 const LoggedInHandler         = require('./logged_in');
 const Redirect                = require('./redirect');
+const TabSelector             = require('../../_common/tab_selector');
```
negarn: why move this inside the app section? it has it's own _common section above

src/javascript/app/base/binary_pages.js
```diff
@@ -47,6 +48,7 @@ const SetCurrency             = require('../pages/user/set_currency');
 const TelegramBot             = require('../pages/user/telegram_bot');
 const TNCApproval             = require('../pages/user/tnc_approval');
 const VideoFacility           = require('../pages/user/video_facility');
+const MetatraderDownloadUI    = require('../pages/resources/metatrader/download.ui');
```
negarn: this should be on line 16 to follow the alphabetical path order

src/javascript/app/base/binary_pages.js
```diff
@@ -140,6 +142,7 @@ const pages_config = {
     'video-facility'         : { module: VideoFacility,              is_authenticated: true, only_real: true },
     'why-us'                 : { module: WhyUs },
     'telegram-bot'           : { module: TelegramBot,                is_authenticated: true },
+    download                 : { module: MetatraderDownloadUI },
```
negarn: this should be on line 80 to follow the alphabetical page order

src/javascript/app/pages/resources/metatrader/download.ui.js
```diff
@@ -0,0 +1,54 @@
+import OSDetect from '../../../../../templates/_common/os_detect';
+
+const toggleDownloadPage = target => {
+    document.querySelectorAll('.download-block').forEach(block => {
+        if (block.getAttribute('id') === target) {
+            block.classList.add('visible');
```
negarn: instead of `.classList.add('visible')` and `.classList.remove('invisible')` you can use `.setVisibility(0)` and `.setVisibility(1)` respectively

src/templates/static/metatrader/download.jsx
```diff
@@ -1,64 +1,111 @@
...
-    badge,
-    desc,
-    download,
-    href,
-    image,
-    os,
-    target,
-    title,
-}) => (
+const DownloadBlock = ({ badge, desc, download, href, image, os, target, title }) => (
```
negarn: if there are four or more arguments we break it into multiline, you can find it here https://github.com/binary-com/binary-static/blob/master/src/templates/README.md#jsx-rules

src/templates/static/metatrader/download.jsx
```diff
@@ -1,64 +1,111 @@
...
+            href={os ? `https://download.mql5.com/cdn/mobile/mt5/${os}?server=Binary.com-Server` : href}
+            download={download || undefined}
+            target={target || undefined}
+        >
             <span className={badge || undefined}>{badge ? '' : it.L('Download')}</span>
         </a>
     </div>
 );
 
+const DesktopDownloadBlock = ({ badge, download, href, image, os, target, title, id }) => (
```
negarn: same here

src/templates/static/metatrader/download.jsx
```diff
@@ -1,64 +1,111 @@
...
+    <div className='gr-12 gr-12-m gr-padding-10 invisible download-block' id={id}>
+        <img className='gr-6 gr-centered' src={it.url_for(`images/pages/metatrader/icons/${image}.svg`)} />
+        <a
+            className={!badge ? 'button' : undefined}
+            href={os ? `https://download.mql5.com/cdn/mobile/mt5/${os}?server=Binary.com-Server` : href}
+            download={download || undefined}
+            target={target || undefined}
+        >
```
negarn: why not store this anchor element and reuse it in both blocks?

src/templates/static/metatrader/download.jsx
```diff
@@ -1,64 +1,111 @@
...
+                    'Looking for [_1] or [_2] Desktop apps?',
+                    `<a data-type='alt-link' data-target='macos' id='enable-mac-alt-link'>${it.L('Mac')}</a>`,
+                    `<a data-type='alt-link' data-target='linux' id='enable-linux-alt-link'>${it.L('Linux')}</a>`
+                )}
+            </p>
+            <p className='alternative-download-description invisible' id='linux-alternative-description'>
+                {it.L(
+                    'Looking for [_1] or [_2] Desktop apps?',
+                    `<a data-type='alt-link' data-target='windows' id='enable-windows-alt-link'>${it.L('Windows')}</a>`,
+                    `<a data-type='alt-link' data-target='macos' id='enable-mac-alt-link'>${it.L('Mac')}</a>`
```
negarn: I would suggest to store each anchor as it seems to be reusable

src/templates/static/metatrader/download.jsx
```diff
@@ -1,64 +1,111 @@
...
+            <h1 className='invisible download-heading' id='macos-heading'>
+                {it.L('MT5 for Mac')}
+            </h1>
+            <h1 className='invisible download-heading' id='linux-heading'>
+                {it.L('MT5 for Linux')}
+            </h1>
+            <p>
+                {it.L(
+                    'Download MT5 for your desktop or laptop to access the powerful tools and features enjoyed by millions of traders.'
+                )}
```
negarn: we don't really break localization strings into multiline, I think you can put it in one line instead of three

src/javascript/app/base/binary_pages.js
easteregg: It's a typo. I'll change this back.

src/javascript/app/pages/resources/metatrader/download.ui.js
easteregg: The only reason behind using `visible` class is to create a possibility to hook on visible items to do transitions and stuff. There is no other logic behind this. 🤔 

I was looking for `.setVisibility()` API in MDN 😃. I did find it on our very own `plugins.js`. Yeah that's a good shortcut. I'll use it.

src/templates/static/metatrader/download.jsx
easteregg: I couldn't do this. By refactoring the string part only, `it.L` will be undefined, and changing it to `localize` function would result in `ReferenceError: Element is not defined`.

I'm not sure, but by removing localize import the error goes away.

```
at Object.<anonymous> (src/javascript/_common/lib/polyfills/element.matches.js:1:63)
```

negarn: look at something like this https://github.com/binary-com/binary-static/blob/master/src/templates/app/cashier/payment_methods.jsx#L78, we declare the variable outside of the return function and then use it inside, localization works fine

src/templates/static/metatrader/download.jsx
easteregg: This string is too long (more than 80 characters and reading it requires horizontal scrolling. Vertical texts are easier on the eye. 

@negar-binary Let me know if we have to do it in one-line only.

negarn: it's more consistent with the rest of the code base, but it's up to you.

negarn:
please run `svgo` on all your images so that they are compressed to one line

src/javascript/_common/os_detect.js
```diff
@@ -0,0 +1,23 @@
+const OSDetect = () => {
+    const Systems = {
+        macos  : ['Mac68K', 'MacIntel', 'MacPPC'],
+        linux  : ['HP-UX', 'Linux i686', 'Linux armv7l', ''],
+        ios    : ['iPhone', 'iPod', 'iPad'],
+        android: ['Android'],
+        windows: ['Win16', 'Win32', 'WinCE'],
+    };
+    if (typeof navigator !== 'undefined') {
```
negarn: not sure if it's better to change this to `if (navigator && navigator.platform)`

negarn: or perhaps `getPropertyValue(navigator, 'platform')`

src/javascript/app/base/binary_pages.js
```diff
@@ -1,5 +1,5 @@
 // ==================== _common ====================
-const TabSelector = require('../../_common/tab_selector'); // eslint-disable-line import/order
+const TabSelector             = require('../../_common/tab_selector'); // eslint-disable-line import/order
```
negarn: extra whitespace

src/javascript/_common/os_detect.js
```diff
@@ -0,0 +1,23 @@
...
+        return Object.keys(Systems)
+            .map(os => {
+                if (Systems[os].some(platform => navigator.platform === platform)) {
+                    return os;
+                }
+                return false;
```
negarn: still suggest changing this to `.map(os => Systems[os].some(platform => navigator.platform === platform))`

src/javascript/_common/os_detect.js
```diff
@@ -0,0 +1,23 @@
...
+            })
+            .filter(os => os)[0];
+    }
+
+    return 'Unknown OS';
```
negarn: still suggest localizing this

src/javascript/_common/os_detect.js
```diff
@@ -0,0 +1,23 @@
+const OSDetect = () => {
+    const Systems = {
+        macos  : ['Mac68K', 'MacIntel', 'MacPPC'],
+        linux  : ['HP-UX', 'Linux i686', 'Linux armv7l', ''],
+        ios    : ['iPhone', 'iPod', 'iPad'],
+        android: ['Android'],
+        windows: ['Win16', 'Win32', 'WinCE'],
+    };
```
negarn: might be better to declare this outside, then we don't have to re-declare it every time the function is called

src/javascript/app/base/binary_pages.js
negarn: still not resolved...?

src/javascript/app/pages/resources/metatrader/download.ui.js
```diff
@@ -0,0 +1,48 @@
+import OSDetect from '../../../../_common/os_detect';
+
+const toggleDownloadPage = target => {
+    document.querySelectorAll('.download-block').forEach(block => {
+        if (block.getAttribute('id') === target) {
+            block.setVisibility(1);
+        } else {
+            block.setVisibility(0);
+        }
```
negarn: why not `block.setVisibility(block.getAttribute('id') === target)`?

src/javascript/app/pages/resources/metatrader/download.ui.js
```diff
@@ -0,0 +1,48 @@
...
+        } else {
+            text.setVisibility(0);
+        }
+    });
+};
+const DownloadMetatrader = (() => {
+    const onLoad = () => {
+        const os = OSDetect();
+
+        // Hide or show a default Item based on navigator.platform
```
negarn: nitpick, `item` should be lowercase

src/sass/static/pages.scss
```diff
@@ -1010,3 +1010,48 @@ body #not_authenticated_financial ul.checked > li {
...
+    height: 46px;
+
+    & img {
+        max-width: 100%;
+        width: 32px;
+    }
+    & .mt-link {
+        font-size: smaller;
+    }
+}
```
negarn: it's better to wrap all these in the id of the page that it's being used in

negarn: in fact, we should already have a metatrader scss file

src/templates/static/metatrader/download.jsx
```diff
@@ -1,108 +1,228 @@
...
+        desc,
+        download,
+        href,
+        image,
+        os,
+        target,
+        title,
+        children,
+    }
+) => (
```
negarn: if you put the curly brackets on the same line as the brackets and indent everything back once, it will be the same as it used to be and consistent with our other files

negarn: same for the blocks below

src/templates/static/metatrader/download.jsx
```diff
@@ -1,108 +1,228 @@
...
+            </DownloadButton>
         </div>
+    </div>
+);
 
-        <SeparatorLine no_wrapper />
+const Download = () => {
+    const windows_link = `<a data-type='alt-link' data-target='windows' id='enable-windows-alt-link'>${it.L('Windows')}</a>`;
+    const linux_link = `<a data-type='alt-link' data-target='linux' id='enable-linux-alt-link'>${it.L('Linux')}</a>`;;
+    const mac_link = `<a data-type='alt-link' data-target='mac' id='enable-mac-alt-link'>${it.L('Mac')}</a>`;;
```
negarn: could you align these by `=`

src/templates/static/metatrader/download.jsx
```diff
@@ -1,108 +1,228 @@
...
+            <div className='gr-padding-20'>
+                <h1 className='invisible download-heading' id='windows-heading'>
+                    {it.L('MT5 for Windows')}
+                </h1>
+                <h1 className='invisible download-heading' id='macos-heading'>
+                    {it.L('MT5 for Mac')}
+                </h1>
+                <h1 className='invisible download-heading' id='linux-heading'>
+                    {it.L('MT5 for Linux')}
+                </h1>
```
negarn: these h1 elements also look like they could be turned into a block and reused. if you change macos to mac, you can only pass in one parameter to use both for the id and the header text, just by making it lowercase in id

src/templates/static/metatrader/download.jsx
```diff
@@ -1,108 +1,228 @@
...
+                <p>
+                    {it.L(
+                        'Download MT5 for your desktop or laptop to access the powerful tools and features enjoyed by millions of traders.'
+                    )}
+                </p>
+                <div className='gr-row'>
+                    <DesktopDownloadBlock
+                        id='windows'
+                        title={it.L('Windows')}
+                        image='windows.png'
```
negarn: if you make the id and image have the  same name you don't need to pass it twice. the .png can be assumed as it's true for all of them right now.

negarn: I suggest also using the title to derive the id

src/templates/static/metatrader/download.jsx
```diff
@@ -1,108 +1,228 @@
...
+                        linux_link
+                    )}
+                </p>
+                <p className='alternative-download-description invisible' id='linux-alternative-description'>
+                    {it.L(
+                        'Looking for [_1] or [_2] Desktop apps?',
+                        windows_link,
+                        mac_link
+                    )}
+                </p>
```
negarn: these can also be a block

src/javascript/app/pages/resources/metatrader/download.ui.js
easteregg: Done.

src/javascript/_common/os_detect.js
easteregg: Navigator support is so good that we can almost remove this. But, if it is not defined, using it in any other context than `typeof` would throw a reference error. 

Maybe the safer choice would be `type of navigator !== 'undefined' && navigator.hasOwnProperty('platform');`

src/templates/static/metatrader/download.jsx
easteregg: I'm getting confused by [this](https://github.com/binary-com/binary-static/pull/4408#pullrequestreview-160991464)

src/javascript/_common/os_detect.js
easteregg: This only return boolean that which `index` of the `Object.keys()`'s array has the item. We have to search the array again to fetch `os` name

src/javascript/app/pages/resources/metatrader/download.ui.js
easteregg: I'm personally against one-liners. I think one level `if` statements are more readable, especially when it comes to doing something based on the result if the statement. Although using a short-hand version is fine in this case too (and I am going to use it), It is not future proof as adding more logic, later on, will make the code more ugly.

src/javascript/_common/os_detect.js
negarn: if navigator is undefined the `if (navigator` part will return false and it won't check `&& navigator.platform)` part at all, but it's up to you

src/templates/static/metatrader/download.jsx
negarn: for variables that are grouped together we usually align by `=`, if they are grouped by themselves then there's no need for extra whitespace

src/javascript/_common/os_detect.js
easteregg: screen shot 2018-10-03 at 12 01 43

negarn: I was thinking about a declared variable with undefined value :) sorry about that, you're way is fine.

src/javascript/app/base/binary_pages.js
```diff
@@ -120,6 +121,7 @@ const pages_config = {
     'cfds'                   : { module: GetStarted.CFDs },
     'contract-specifications': { module: TabSelector },
     'cryptocurrencies'       : { module: GetStarted.Cryptocurrencies },
+    'download'               : { module: MetatraderDownloadUI },
```
negarn: we only need quotation marks for items with a dash in them, e.g. `'contract-specifications'`. all the others that don't have a dash but have quotation marks around them were mistakenly put here 😞

src/templates/static/metatrader/download.jsx
```diff
@@ -10,99 +28,180 @@ const DownloadBlock = ({
...
     target,
     title,
+    children,
 }) => (
-    <div className='gr-4 gr-6-m gr-padding-10'>
-        <img className='gr-6 gr-centered' src={it.url_for(`images/pages/metatrader/icons/${image}.svg`)} />
+    <div className='gr-4 gr-12-m gr-padding-10'>
+        <img
+            className='gr-12 gr-centered mt-mobile-phone'
+            src={/[.]/.test(image) ? it.url_for(`images/pages/metatrader/icons/${image}`) : it.url_for(`images/pages/metatrader/icons/${image}.svg`)}
```
negarn: is it only about having `.svg` at the end or not? if so we can move the ternary inside so we don't repeat the path, what do you think?

src/javascript/app/base/binary_pages.js
easteregg: Yeah, I don't usually do that myself, but it is aligned with the rest and looks good too.

src/javascript/_common/os_detect.js
```diff
@@ -0,0 +1,69 @@
...
+            .map(os => {
+                if (systems[os].some(platform => navigator.platform === platform)) {
+                    return os;
+                }
+                return false;
+            })
+            .filter(os => os)[0];
+    }
+
+    return localize('Unknown OS');
```
negarn: the string needs to be added to `generate-static-data.js` as well

src/javascript/app/pages/resources/metatrader/download.ui.js
```diff
@@ -0,0 +1,46 @@
+import { OSDetect, isDesktop } from '../../../../_common/os_detect';
+
+
```
negarn: extra white line

src/javascript/_common/os_detect.js
```diff
@@ -0,0 +1,75 @@
...
+    return !!['windows', 'mac', 'linux'].find(system => system === os);
+};
+
+const isMobile = () => {
+    const os = OSDetect();
+    return !!['ios', 'android'].find(system => system === os);
+};
+
+const OSDetect = () => {
+    // For testing purposes or more compatibility, if we set 'config.os'
```
negarn: where are we setting this value?

easteregg: Manually in developer tools.

PR #4426, opened 2018-10-05
ruskakimov/mt5_form_optimization

ruskakimov:


src/templates/app/user/profit_table.jsx
```diff
@@ -11,7 +11,7 @@ const ProfitTable = () => (
             <div className='gr-padding-10 invisible' id='util_row'>
                 <div className='gr-row gr-gutter-right gr-row-align-right'>
                     <div className='label_form gr-gutter-right'>
-                        <label htmlFor='date_to'>{it.L('Show all historical transactions up to')}:</label>
+                        <label>{it.L('Show all historical transactions up to')}:</label>
```
negarn: ?

ruskakimov: @negarn previously Kelly removed label interaction with CSS (`pointer-events: none`). It was for the purpose of preventing datapicker on label click. These CSS changes then broke radio buttons. So, we decided to just remove the binding of label and datapicker.

PR #4431, opened 2018-10-08
easteregg/update_eslint

easteregg:


negarn:
you will need to fix the eslint errors in the same PR 🙂

src/javascript/landing_pages/common.js
```diff
@@ -98,7 +98,7 @@ function isBinaryApp() {
 }
 
 function getAppId() {
-    return localStorage.getItem('config.app_id') || (isBinaryApp() ? '14473' : (/staging\.binary\.com/i.test(window.location.hostname) ? '1098' : '1'));
+    return localStorage.getItem('config.app_id') || (isBinaryApp() ? '14473' : (/staging\.binary\.com/i.test(window.location.hostname) ? '1098' : '1')); // eslint-disable-line no-nested-ternary
```
negarn: better to fix it than ignore it?

.eslintrc.js
```diff
@@ -32,6 +32,7 @@ module.exports = {
         'space-in-parens'                   : ['error', 'never'],
         'space-infix-ops'                   : 'error',
         'space-unary-ops'                   : 'error',
+        'no-multiple-empty-lines'           : ["error", { "max": 1, "maxEOF": 1 }],
```
binary-hamedanchi: Not an issue, but would appreciate it if it follows the same style currently used here. 🙂 
(to put in alphabetical order and use single quotes)

PR #4435, opened 2018-10-08
easteregg/add_rtp_lookback_table

easteregg:


src/templates/static/get_started/index.jsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import SeparatorLine from '../../_common/components/separator_line.jsx';
 import { TabContainer, TabContent, TabContentContainer, TabsSubtabs } from '../../_common/components/tabs.jsx';
 
+
```
negarn: empty line? :)

src/templates/static/get_started/index.jsx
```diff
@@ -42,6 +43,70 @@ const CTA = () => (
     </div>
 );
 
+const VolatilityIndicesTable = () => {
+    const active_symbols = [
+        {
+            key             : 'r10',
```
negarn: why do we need the key? why name it active symbols? it's about the RTP values

src/templates/static/get_started/index.jsx
```diff
@@ -42,6 +43,70 @@ const CTA = () => (
...
+            label           : it.L('Volatility Indices 10'),
+            high_close_value: 86.85,
```
negarn: as I understand it needs to be rounded to the nearest integer, i.e. `87`

src/templates/static/get_started/index.jsx
```diff
@@ -42,6 +43,70 @@ const CTA = () => (
...
+                        <th className='align-end'>{it.L('High - Close')}</th>
+                        <th className='align-end'>{it.L('Close - Low')}</th>
+                        <th className='align-end'>{it.L('High - Low')}</th>
+                    </tr>
+                </thead>
+                <tbody>
+                    {active_symbols.map(item => (
+                        <tr key={item.key}>
+                            <td>{item.label}</td>
+                            <td className='align-end'>{Math.round(item.high_close_value)}</td>
```
negarn: why not centred? it looks a bit weird with align end

src/templates/static/get_started/index.jsx
```diff
@@ -310,6 +375,7 @@ const Index = () => (
                             <span>{it.L('2 * (6,000 – 5,200) = $1600')}</span>
                         </div>
                         <p>{it.L('Lookbacks options are currently only available for [_1]Volatility Indices[_2].', `<a href=${it.url_for('get-started/binary-options')}#range-of-markets>`, '</a>')}</p>
+                        <VolatilityIndicesTable />
```
negarn: we also need the sentence `Return to player (RTP) % for lookbacks for a multiplier of 1 and 1m duration is around 87% on average` above the table, as per the card

src/templates/static/get_started/index.jsx
negarn: ah, I see that you round it later. up to you, I don't see a point.

easteregg: I was not sure if these values are going to change later or not. If by any chance we make this table dynamic, Math.round will help us later.

src/templates/static/get_started/index.jsx
easteregg: The text label is way wider than the values, From the excel days, we always formatted the numbers that matter to the right. I can do center too but I preferred this one :smile:. 
If you think we should go with the center, say so.

src/templates/static/get_started/index.jsx
easteregg: When looping over components, we need to define a unique identifier for each component, (see `src/templates/static/get_started/index.jsx:97`). 

About the second question, I don't know I thought it is active symbols. renaming it to `rtp_values`

src/templates/static/get_started/index.jsx
```diff
@@ -310,6 +310,7 @@ const Index = () => (
                             <span>{it.L('2 * (6,000 – 5,200) = $1600')}</span>
                         </div>
                         <p>{it.L('Lookbacks options are currently only available for [_1]Volatility Indices[_2].', `<a href=${it.url_for('get-started/binary-options')}#range-of-markets>`, '</a>')}</p>
+                        <p>{it.L('Return to player (RTP) % for lookbacks for a multiplier of 1 and 1m duration is around 87% on average')}</p>
```
jy-binary: missing fullstop at the end of the sentence

PR #4438, opened 2018-10-09
ruskakimov/app_download_gtm

ruskakimov:


src/javascript/static/pages/platforms.js
```diff
@@ -1,30 +1,41 @@
...
                 }
-                if (/\.dmg$/.test(asset.browser_download_url)) {
-                    link_mac = asset.browser_download_url;
-                } else if (/\.exe$/.test(asset.browser_download_url)) {
-                    link_windows = asset.browser_download_url;
-                    // } else if (/x86_64\.AppImage$/.test(asset.browser_download_url)) {
-                    //     link_linux = asset.browser_download_url;
-                }
+                os_list.forEach(os => {
+                    if (os.download_url) return;
```
negarn: there is no return in foreach as far as I know

src/javascript/_common/base/gtm.js
```diff
@@ -12,35 +12,20 @@ const getAppId       = require('../../config').getAppId;
...
-            event,
-            data: data_layer_info,
-        };
-    };
+    const getCommonVariables = () => ({
+        language : getLanguage(),
+        pageTitle: pageTitle(),
+        pjax     : State.get('is_loaded_by_pjax'),
+        url      : document.URL,
+        ...ClientBase.isLoggedIn() && { visitorId: ClientBase.get('loginid') },
```
negarn: we don't need event anymore?

ruskakimov: it's from `new-app`, we will push event explicitly with `pushDataLayer({ event: 'smth' } )`

ruskakimov: forgot to pass `page_load` though in binary_loader 😅

PR #4448, opened 2018-10-10
negar/check_reality_error

negarn:
show error and stop client from trading if reality check gives error


src/javascript/app/pages/user/reality_check/reality_check.ui.js
```diff
@@ -120,7 +120,11 @@ const RealityCheckUI = (() => {
     const getSummaryAsync = () => {
         if (RealityCheckUI.shouldShowPopup()) {
             BinarySocket.send({ reality_check: 1 }).then((response) => {
-                getAjax(RealityCheckData.summaryData(response.reality_check));
+                if (response.reality_check) {
+                    getAjax(RealityCheckData.summaryData(response.reality_check));
+                } else if (response.error && !/user\/statementws\.html/.test(window.location.pathname)) { // don't block statement page for reality check error, but block all other pages
```
kellybinary: nitpick: i think its better to check for error first :)

PR #4459, opened 2018-10-12
Extract js texts

binary-hamedanchi:


src/javascript/app/base/binary_loader.js
```diff
@@ -124,7 +124,7 @@ const BinaryLoader = (() => {
         }
 
         const div_container = createElement('div', { class: 'logged_out_title_container', html: content.getElementsByTagName('h1')[0] });
-        const div_notice    = createElement('p', { class: 'center-text notice-msg', html: localize(message) });
+        const div_notice    = createElement('p', { class: 'center-text notice-msg', html: message });
```
negarn: do we want to use `localized_text` instead of `message` here too?

src/javascript/app/common/attach_dom/dialog.js
```diff
@@ -19,7 +18,7 @@ const Dialog = (() => {
                     if (!el_dialog) return;
 
                     const message = Array.isArray(options.message) ? options.message.join('<p />') : options.message;
-                    elementInnerHtml(container.querySelector('#dialog_message'), localize(message));
+                    elementInnerHtml(container.querySelector('#dialog_message'), message);
```
negarn: same here

scripts/README.md
```diff
@@ -17,7 +17,34 @@
   * Commit changes and push to origin
   * Open github to submit the PR
 
-## Update sitemap
+### Extracting texts from js code:
+Texts that used in js code would automatically be extracted during the translation update process (3rd step above: Updating the source file).
```
negarn: Texts that **are** used in js code **will** automatically

scripts/README.md
```diff
@@ -17,7 +17,34 @@
...
 
-* List of paths to include in `sitemap.xml` is here: [sitemap.js](sitemap.js)
+In order to make sure that the extractor code won't miss any string to push to translations, please:
```
negarn: perhaps clearer?
`In order to make sure that no strings are missed by the extractor code while pushing to translations, please`

scripts/README.md
```diff
@@ -17,7 +17,34 @@
...
+
+1. Refactor the code so that the first argument passed to `localize()` method, be a string literal.
```
negarn: no need for a comma, and I think `is` instead of `be`, so:
```
the first argument passed to `localize()` method is a string literal
```

scripts/README.md
```diff
@@ -17,7 +17,34 @@
...
+    i.e.
+    ```js
+    const text = localize(is_started ? 'Sell at market' : 'Sell');
+    ```
+    could change to:
```
negarn: I think `would` instead of `could`, as it's not an optional change but a required change

scripts/README.md
```diff
@@ -17,7 +17,34 @@
...
+    ```js
+    const text = is_started ? localize('Sell at market') : localize('Sell');
+    ```
+2. If there is no way to have the string literal in js code (i.e. API texts which are not translated), add them to `scripts/js_texts/static_strings_app.js`.
+3. At the end, when you're sure that a strings already been taken care of somewhere else, just put this comment `/* localize-ignore */` right after the first argument of `localize()` call to ignore it from tests.
```
negarn: `a string has already been taken care of`

scripts/README.md
```diff
@@ -17,7 +17,34 @@
...
+
+### `js_texts` folder contents:
+```
+js_texts:
+    ├── extracted_strings_app.js
+    └── static_strings_app.js
+```
+- `extracted_strings_app.js`: Contains extracted strings. It's for debugging purpose only and shouldn't change manually.
```
negarn: I think `shouldn't be updated manually` or `shouldn't be changed manually` would be clearer

scripts/README.md
```diff
@@ -17,7 +17,34 @@
...
+1. Refactor the code so that the first argument passed to `localize()` method, be a string literal.
+    i.e.
+    ```js
+    const text = localize(is_started ? 'Sell at market' : 'Sell');
+    ```
+    could change to:
+    ```js
+    const text = is_started ? localize('Sell at market') : localize('Sell');
+    ```
+2. If there is no way to have the string literal in js code (i.e. API texts which are not translated), add them to `scripts/js_texts/static_strings_app.js`.
```
negarn: should we mention that even if it's a string that's translated automatically somewhere else we should still include it here to avoid missing it if that other string gets removed?

scripts/README.md
```diff
@@ -17,7 +17,34 @@
...
+### `js_texts` folder contents:
+```
+js_texts:
+    ├── extracted_strings_app.js
+    └── static_strings_app.js
+```
+- `extracted_strings_app.js`: Contains extracted strings. It's for debugging purpose only and shouldn't change manually.
+- `static_strings_app.js` to handle those strings that don't exist in js code. e.g. API texts that are not localised, etc.
+
+During the translation update process, the source file `messages.pot` would be updated with all texts from both the above files.
```
negarn: I think `will` instead of `would`

scripts/README.md
binary-hamedanchi: @negarn I added this as a [note](https://github.com/binary-com/binary-static/pull/4459/files#diff-efc0b58e058662dd930679e42e8517ecR36). Could you please review again?
Thanks.

negarn: looks good, thank you

PR #4466, opened 2018-10-15
easteregg/add_mobx_react_devtools

easteregg:
I have added mobx react Devtools to app.jsx. 
Also, added an error boundary element wrapper to prevent the application from crashing when hitting an error.


easteregg:
This doesn't look healthy. closing and creating another one.

PR #4472, opened 2018-10-15
[new-app] Add jsx support

easteregg:
This provides the ability to import `jsx` files without the need for `index.js`. If you approve this, then we have another PR incoming. In that pull-request, I will fix the code and use direct module imports to make the code easier to understand.


src/javascript/app/pages/trade/common.js
```diff
@@ -1,8 +1,8 @@
 const Defaults         = require('./defaults');
 const Symbols          = require('./symbols');
 const Tick             = require('./tick');
-const contractsElement = require('./contracts.jsx');
-const marketsElement   = require('./markets.jsx');
+const contractsElement = require('./contracts');
```
binary-hamedanchi: IMO, this reduces the readability as you can't tell the type of contents by just looking at the import statements.

Better to know everyone's opinions.

easteregg:
I'll close this PR, maybe in the future, we open it again.

PR #4479, opened 2018-10-16
kelly/update_platforms

kellybinary:
- show Desktop App download section for non-eu clients

### DIFF
--- a/src/javascript/static/pages/platforms.js
+++ b/src/javascript/static/pages/platforms.js
@@ -19,6 +19,7 @@ const os_list = [
 
 const Platforms = (() => {
     const onLoad = () => {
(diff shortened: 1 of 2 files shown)


src/javascript/static/pages/platforms.js
```diff
@@ -39,11 +38,7 @@ const Platforms = (() => {
             });
         });
 
-        BinarySocket.wait('website_status', 'authorize', 'landing_company').then(() => {
-            if (isEuCountry()) {
-                $('.eu-hide').setVisibility(0);
-            }
-        });
+        hideEU();
```
negarn: sorry, could you move this to be the first thing called on load?

PR #4484, opened 2018-10-17
Update SmartTrader 2.0 for SmartCharts chunked implementation (v0.3)

brucebinary:


src/javascript/app_2/Modules/SmartChart/Containers/smart_chart.jsx
```diff
@@ -7,6 +7,8 @@ import ChartMarker      from '../Components/Markers/marker.jsx';
 import TopWidgets       from '../Components/top_widgets.jsx';
 import { symbolChange } from '../Helpers/symbol';
 
+setSmartChartsPublicPath('/js/app_2/');
```
brucebinary: @binary-hamedanchi this will not work for branch deployments `grunt dev --branch featureA`. I'm not quite sure if there is a global variable exposed for this?

binary-hamedanchi: @brucebinary Probably [`urlForStatic()`](https://github.com/binary-com/binary-static/blob/master/src/javascript/_common/url.js#L58) would be helpful here.

PR #4486, opened 2018-10-17
kelly/desktop_app_ID_only

kellybinary:
- show Desktop App download section to ID clients only for now


src/javascript/app/common/common_functions.js
```diff
@@ -8,6 +9,7 @@ const hideEU = () => {
             $('.eu-show').setVisibility(1);
             $('.eu-hide-parent').parent().setVisibility(0);
         }
+        $('.id-show').setVisibility(isIndonesia());
```
negarn: if we only need this in platforms page, we don't need to call it in such a common function and can only call it in platforms page

kellybinary: okay sure

PR #4517, opened 2018-10-23
Usbpage

sashcy:


src/templates/landing_pages/introducing_usb.jsx
```diff
@@ -0,0 +1,356 @@
...
+                                </div>
+                            </a>
+                        </div>
+                    </div>
+                    <div className='col-xl-9'>
+                        <div id='push' className='grill grill-ico' />
+                        <div className='row'>
+                            <div className='col'>
+                                <div id='push-nav' className='topnav'>
+                                    <a className='page-scroll' href='#video'>{('Overview')}</a>
```
negarn: if you don't want translations, you don't need this brackets at all. you can just do something like `<a className='page-scroll' href='#video'>Overview</a>`

src/templates/landing_pages/introducing_usb.jsx
```diff
@@ -0,0 +1,356 @@
+import React from 'react';
+import Layout from './_common/layout.jsx';
+
+const UsbPage = () => (
+    <Layout
+        meta_description={it.L('[_1] Introducing USB', it.broker_name)}
```
negarn: do we need translations for this? if not it can just be ```{`${it.broker_name} Introducing USB`}```

src/templates/landing_pages/introducing_usb.jsx
```diff
@@ -0,0 +1,356 @@
...
+            </div>
+        </div>
+        <div className='section-md primary-bg-color coin-section'>
+            <div className='container center-text'>
+                <div className='row align-items-center'>
+                    <div className='col-lg-2 col-md-3 col-sm-4'>
+                        <img className='usb-coin' src={it.url_for('images/usb_page/usb-coin-logo.png')} alt='USB' />
+                    </div>
+                    <div className='col-lg-10 col-md-9 col-sm-8'>
+                        <p className='content-inverse-color ft-400'>{('Guaranteed by Binary.com, USB is a digital currency pegged to the US Dollar on a one-to-one basis. Send and receive payments with the peace of mind that your currency is backed by the strength of Binary.com’s balance sheet.')}</p>
```
negarn: please replace all instances of Binary.com with [_1] and it.broker_name

src/sass/landing_pages/usbstyle.scss
```diff
@@ -0,0 +1,1993 @@
...
+    margin-right: 0;
+    margin-left: 0;
+}
+
+.no-gutters > .col, .no-gutters > [class*='col-'] {
+    padding-right: 0;
+    padding-left: 0;
+}
+
+.col-1, .col-2, .col-3, .col-4, .col-5, .col-6, .col-7, .col-8, .col-9, .col-10, .col-11, .col-12, .col, .col-auto, .col-sm-1, .col-sm-2, .col-sm-3, .col-sm-4, .col-sm-5, .col-sm-6, .col-sm-7, .col-sm-8, .col-sm-9, .col-sm-10, .col-sm-11, .col-sm-12, .col-sm, .col-sm-auto, .col-md-1, .col-md-2, .col-md-3, .col-md-4, .col-md-5, .col-md-6, .col-md-7, .col-md-8, .col-md-9, .col-md-10, .col-md-11, .col-md-12, .col-md, .col-md-auto, .col-lg-1, .col-lg-2, .col-lg-3, .col-lg-4, .col-lg-5, .col-lg-6, .col-lg-7, .col-lg-8, .col-lg-9, .col-lg-10, .col-lg-11, .col-lg-12, .col-lg, .col-lg-auto, .col-xl-1, .col-xl-2, .col-xl-3, .col-xl-4, .col-xl-5, .col-xl-6, .col-xl-7, .col-xl-8, .col-xl-9, .col-xl-10, .col-xl-11, .col-xl-12, .col-xl, .col-xl-auto {
```
negarn: it would be good if you could use our existing classes instead of adding all these new classes. we are using a mixture of grid and flexbox.

src/templates/landing_pages/introducing_usb.jsx
sashcy: Actually, we do since ESlint has max length 120

src/templates/landing_pages/introducing_usb.jsx
sashcy: I am not sure if we do

src/sass/landing_pages/usbstyle.scss
sashcy: Cleaned this up and loading from academy. 
Since we use mobile-first flexbox grid for all marketing projects I used this grid, not knowing that this page will be a part of the binary-static. 
If this is a problem I can redo using front-end grid.

src/templates/landing_pages/introducing_usb.jsx
negarn: as per this line https://github.com/binary-com/binary-static/pull/4517/files#diff-77228b84b3d4618d6fa968aeff50f04cR117 we are not showing this page in any language other than English, and you're not translating any other strings.

sashcy: noted

PR #4524, opened 2018-10-25
[new-app] kelly/fix_indicative_price_not_updating

kellybinary:


src/javascript/app_2/Services/subscription_manager.js
```diff
@@ -68,7 +68,8 @@ const SubscriptionManager = (() => {
         const stream_id = getPropertyValue(response, [response.msg_type, 'id']);
 
         if (!subscriptions[sub_id]) {
-            if (!forget_requested[stream_id]) {
+            const is_expired = getPropertyValue(response, [response.msg_type, 'is_expired']); // forget proposal_open_contract only if expired
+            if (!forget_requested[stream_id] && is_expired) {
```
negarn: as far as I know this is a general function used for all, not only proposal open contract, so is_expired will be false for all other API calls and not send forget for them. am I wrong?

PR #4529, opened 2018-10-28
negar/fix_translations

negarn:
update api translations


src/javascript/app/pages/trade/markets.jsx
```diff
@@ -4,7 +4,7 @@ import Symbols from './symbols';
 // Should be remove in the future
 import Defaults from './defaults';
 import { getElementById } from '../../../_common/common_functions';
-import { localize } from '../../../_common/localize';
+// import { localize } from '../../../_common/localize';
```
kellybinary: any reason to comment this out instead of removing this line?

src/javascript/app/pages/trade/contracts.jsx
```diff
@@ -127,7 +124,7 @@ class Contracts extends React.Component {
                     ref={this.saveRef.bind(null, 'wrapper')}
                 >
                     <div className={`mobile_close invisible ${open && is_mobile ? '' : 'disabled'}`}>
-                        <span>{localize('Select Trade Type')}</span>
+                        <span>{it.L('Select Trade Type')}</span>
```
binary-hamedanchi: This is a client side component. `it.L()` is only available on `src/templates/` folder for server side rendering.

Same applies to the changes on [`markets.jsx`](https://github.com/binary-com/binary-static/pull/4529/files#diff-7a3d234a678624c37405926dfebf6429).

negarn:
@binary-hamedanchi @kellybinary I have reverted the last commit, thanks for checking :)

PR #4530, opened 2018-10-28
[new-app] change regex to be compatible with most browsers

easteregg:


src/javascript/app_2/App/Components/Form/input_field.jsx
```diff
@@ -32,7 +32,7 @@ const InputField = ({
             const is_empty = !e.target.value || e.target.value === '';
             const signed_regex = is_signed ? '[\\+-]?' : '';
 
-            const is_number = new RegExp(`^${signed_regex}(\\d*)?${is_float ? '(\\.\\d+)?' : ''}(?<=\\d)(?<!-0)$`)
+            const is_number = new RegExp(`^${signed_regex}(\\d*)?${is_float ? '(\\.\\d+)?' : ''}(\\d*)?(?!([-0])).|^$`)
```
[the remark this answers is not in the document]
easteregg: @morteza-binary I didn't quite understand the first bullet point. As you know, look-behind is not supported in all major browsers, so we have to imitate that behavior with other rules. 
Are we validating numbers only?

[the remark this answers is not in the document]
easteregg: @morteza-binary Would you check these examples and the new regex that handles all that as well? (this one doesn't use lookbehind and supports safari too). 
```javascript
const regex = /(?!^\-0$)(^[\-+]?(\d*)?(\.\d+)?$)/gm;
const str = `*3213.
.341341
e10923
asdada
-123
090)()91231
-0
-0.134
-.13451
2345.
1234
+1234
-3
-123
-12345161
12314151
+123e12346
0.2313
--123
123151511513510
00.00
123.`;
let m;

while ((m = regex.exec(str)) !== null) {
    // This is necessary to avoid infinite loops with zero-width matches
    if (m.index === regex.lastIndex) {
        regex.lastIndex++;
    }
    
    // The result can be accessed through the `m`-variable.
    m.forEach((match, groupIndex) => {
        console.log(`Found match, group ${groupIndex}: ${match}`);
    });
}
```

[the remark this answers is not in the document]
morteza-binary: @easteregg 
It seems that it supports a wide range of templates, but there is an issue in this regex that asserts `-` or `+` as a valid number.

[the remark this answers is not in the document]
easteregg: You are correct. I'll work on that case as well.

src/javascript/app_2/App/Components/Form/input_field.jsx
```diff
@@ -30,9 +30,9 @@ const InputField = ({
     const changeValue = (e) => {
         if (type === 'number') {
             const is_empty = !e.target.value || e.target.value === '';
-            const signed_regex = is_signed ? '[\\+-]?' : '';
+            const signed_regex = is_signed ? '(?!^\\-0$)(?!^\\-$)(?!^\\+$)^[\\+-]?' : '^';
```
morteza-binary: @easteregg 

```suggestion
            const signed_regex = is_signed ? '(?!^[-+]0?$)^[+-]?' : '^';
```

src/javascript/app_2/App/Components/Form/input_field.jsx
```diff
@@ -30,9 +30,9 @@ const InputField = ({
...
 
-            const is_number = new RegExp(`^${signed_regex}(\\d*)?${is_float ? '(\\.\\d+)?' : ''}(?<=\\d)(?<!-0)$`)
+            const is_number = new RegExp(`${signed_regex}(\\d*)?${is_float ? '(\\.\\d+)?' : ''}$`)
```
morteza-binary: @easteregg 
it asserts empty string `''` as a valid number.

morteza-binary: @easteregg 
It still returns true when the `e.target.value` is an empty string.

src/javascript/app_2/App/Components/Form/input_field.jsx
```diff
@@ -30,22 +30,22 @@ const InputField = ({
     const changeValue = (e) => {
         if (type === 'number') {
             const is_empty = !e.target.value || e.target.value === '';
-            const signed_regex = is_signed ? '[\\+-]?' : '';
+            const signed_regex = is_signed ? '(?!^([-]0|[-+]?)$)^[+-]?' : '^';
```
morteza-binary: @easteregg 
Would you plz change `(?!^([-]0|[-+]?)$)` to `(?!^[-+]0?)$)` since it shouldn't accept `+0` and `-0` as a valid number.

src/javascript/app_2/App/Components/Form/input_field.jsx
easteregg: Can't reproduce this.

src/javascript/app_2/App/Components/Form/input_field.jsx
easteregg: The proposed regex change will accept the empty string as a valid number. please check this out: 

https://regex101.com/r/2zZLns/1

morteza-binary: > The proposed regex change will accept the empty string as a valid number. please check this out:
> 
> https://regex101.com/r/2zZLns/1

yes, it needs a `?` after `[-+]`. it should be `(?!^[-+]?0?)$)`

src/javascript/app_2/App/Components/Form/input_field.jsx
morteza-binary: @easteregg 
> Can't reproduce this.

when `is_signed` is false the value of `signed_regex` will be `^` and the regex in line 35 will be `/^(\d*)?$/` or `/^(\d*)?(\.\d+)?$/`.

easteregg: This line is handling the empty string, `const is_empty = !e.target.value || e.target.value === '';` 
We don't need to repeat ourselves.

easteregg: I can remove the second rule that lookahead and denies empty string in signed values as well. didn't notice this line exists before.

PR #4531, opened 2018-10-28
[new-app] forgetting error on proposal_open_contract

easteregg:


src/javascript/app_2/Stores/Modules/Contract/contract_store.js
```diff
@@ -111,7 +119,7 @@ export default class ContractStore extends BaseStore {
     @action.bound
     handleSell(response) {
         if (response.error) {
-            this.sell_info = {
+            this.sell_info         = {
```
morteza-binary: @easteregg 
it doesn't need to add extra space when the statement is not settled in the one line.

src/javascript/app_2/Stores/Modules/Contract/contract_store.js
```diff
@@ -36,6 +40,7 @@ export default class ContractStore extends BaseStore {
     @observable chart_config  = observable.object({});
 
     @observable has_error         = false;
```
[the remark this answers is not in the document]
easteregg: that is creating additional dependency on the `error_message` to be always present. That can arise situations like when `error_message` is empty (bad backend code or unset state change), but the error occurs.

src/javascript/app_2/Stores/Modules/Contract/contract_store.js
easteregg: I'll fix my web storm code style formatter.

src/javascript/app_2/Stores/Modules/Contract/contract_store.js
```diff
@@ -80,7 +87,8 @@ export default class ContractStore extends BaseStore {
     @action.bound
     updateProposal(response) {
         if ('error' in response) {
-            this.has_error = true;
+            this.has_error     = true;
+            this.error_message = response.error.message;
```
binary-hamedanchi: There are cases that `proposal_open_contract` does not return an error but an empty object in response: `proposal_open_contract: {}`.

This happens when trying to view a `contract_id` which doesn't exist or doesn't belong to the current client.
We need to handle those cases as well.

Thanks.

src/javascript/app_2/Stores/Modules/Contract/contract_store.js
```diff
@@ -80,7 +88,14 @@ export default class ContractStore extends BaseStore {
...
+            this.contract_info = {};
+            return;
+        }
+        if (isEmptyObject(response)) {
```
binary-hamedanchi: `response.proposal_open_contract` 🙂

PR #4533, opened 2018-10-29
kelly/citizenship

kellybinary:
- checks for `tax_details_required`

### DIFF
--- a/src/javascript/app/pages/user/account/settings/personal_details.js
+++ b/src/javascript/app/pages/user/account/settings/personal_details.js
@@ -220,8 +220,10 @@ const PersonalDetails = (() => {
                     BinaryPjax.loadPreviousUrl();
                     return;
                 }
-                const get_settings    = data.get_settings;
-                const has_required_mt = (/real_vanuatu_(standard|advanced)/.test(redirect_url) ?
+                const get_settings        = data.get_settings;
+                const is_tax_req          = +State.getResponse('landing_company.config.tax_details_required') === 1;
+                const is_for_mt_financial = /real_vanuatu_(standard|advanced)/.test(redirect_url);
+                const has_required_mt = ((is_for_mt_financial && is_tax_req) ?
                     (get_settings.tax_residence && get_settings.tax_identification_number && get_settings.citizen)
                     :
                     get_settings.citizen // only check Citizen if user selects mt volatility account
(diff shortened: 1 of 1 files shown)


src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -221,10 +221,10 @@ const PersonalDetails = (() => {
                     return;
                 }
                 const get_settings    = data.get_settings;
-                const has_required_mt = (/real_vanuatu_(standard|advanced)/.test(redirect_url) ?
-                    (get_settings.tax_residence && get_settings.tax_identification_number && get_settings.citizen)
+                const has_required_mt = (+State.getResponse('landing_company.config.tax_details_required') === 1 ?
+                    (get_settings.tax_residence && get_settings.tax_identification_number)
                     :
-                    get_settings.citizen // only check Citizen if user selects mt volatility account
+                    get_settings.citizen
```
negarn: I'm confused, now we are only checking the current binary accounts needs tax details, but as per the MT5 check there are binary accounts that don't need tax details but for those MT5 account openings they need tax details. doesn't this break that check, or am I missing something?

kellybinary: i've checked with QA and compliance, the same applies to MT5 account openings as well, we need to ask for tax info only if their country require tax info (check for `tax_details_required`), and citizen is needed for all MT5 accounts

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -220,8 +220,10 @@ const PersonalDetails = (() => {
                     BinaryPjax.loadPreviousUrl();
                     return;
                 }
-                const get_settings    = data.get_settings;
-                const has_required_mt = (/real_vanuatu_(standard|advanced)/.test(redirect_url) ?
+                const get_settings        = data.get_settings;
+                const is_tax_req          = +State.getResponse('landing_company.config.tax_details_required') === 1;
+                const is_for_mt_financial = /real_vanuatu_(standard|advanced)/.test(redirect_url);
+                const has_required_mt = ((is_for_mt_financial && is_tax_req) ?
```
negarn: equal sign not aligned with above :)

PR #4538, opened 2018-10-30
zaki/compress_images

zakibinary:
Changelog
- Compresses submitted files with `image` MIME-type into `image/jpeg` MIME.
- Renames all image files to `jpg` extension after compression.
- Resizes insanely high resolution files to not exceed QHD resolution.


src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -206,52 +209,85 @@ const Authenticate = (() => {
...
         let idx_to_upload     = 0;
         let is_any_file_error = false;
-        readFiles(files).then((response) => {
-            response.forEach((file) => {
-                if (file.message) {
-                    is_any_file_error = true;
-                    showError(file);
+
+        compressImageFiles(files).then((files_arr) => {
+            files_to_process = files_arr;
```
negarn: why do we need this instead of simply using `files_arr`?

PR #4548, opened 2018-10-31
Nazanin/blocked-text

nazaninreihani:


src/sass/_common/base/language_base.scss
```diff
@@ -54,6 +54,14 @@ body.FR {
             font-size: $FONT_SIZE_XXS;
         }
     }
+    #trading_socket_container #contract_prices_container {
+        .price_container {
+            .contract_heading {
+                font-size: $FONT_SIZE_XXS;
```
negarn: xxs is too small, could you find a different solution?

nazaninreihani: @negar-binary XS works either, but causes buttons to have different widths, is it okay or buttons should always be the same width?

PR #4560, opened 2018-11-01
[new-app] No client reload while switching

easteregg:


src/javascript/app_2/App/Components/Elements/AccountSwitcher/account_switcher.jsx
```diff
@@ -105,10 +94,18 @@ class AccountSwitcher extends React.Component {
...
+    client            : PropTypes.object,
     is_upgrade_enabled: PropTypes.bool,
     is_visible        : PropTypes.bool,
+    modules           : PropTypes.object,
     onClickUpgrade    : PropTypes.func,
     toggle            : PropTypes.func,
 };
 
-export { AccountSwitcher };
+const account_switcher = connect(
```
morteza-binary: @easteregg 
Regarding the [StyleGuid](https://github.com/binary-com/binary-static/blob/new-app/src/javascript/app_2/Documents/STYLE_GUIDE.MD) just container are allowed to connect to stores.

src/javascript/app_2/App/Components/Elements/AccountSwitcher/account_switcher.jsx
```diff
@@ -83,17 +72,17 @@ class AccountSwitcher extends React.Component {
                     <React.Fragment key={account.loginid}>
                         <div
                             className={classNames('acc-switcher-account', account.icon)}
-                            onClick={this.switchAccount.bind(null, account.loginid)}
+                            onClick={() => this.doSwitch(account.loginid, this.props.client)}
```
morteza-binary: @easteregg 
IMO, it's better to define the function outside of `onClick`.

src/javascript/app_2/App/Components/Elements/DenialOfServiceModal/denial_of_service.jsx
```diff
@@ -0,0 +1,39 @@
+import React                from 'react';
+import PropTypes            from 'prop-types';
+import { getAccountOfType } from '_common/base/client_base';
+import { localize }         from '_common/localize';
+import URL                  from '_common/url';
+import FullPageModal        from 'App/Components/Elements/FullPageModal/full_page_modal.jsx';
```
morteza-binary: @easteregg 
what's your idea ?

```suggestion
import FullPageModal        from '../FullPageModal/full_page_modal.jsx';
```

src/javascript/app_2/App/Components/Elements/DenialOfServiceModal/denial_of_service.jsx
```diff
@@ -0,0 +1,39 @@
...
+        onCancel={onCancel}
+        show={show}
+    />
+);
+
+DenialOfServiceModal.propTypes = {
+    show: PropTypes.bool,
+};
+
+const denial_of_service = connect(
```
morteza-binary: @easteregg 
ditto

src/javascript/app_2/App/Components/Elements/FullPageModal/full_page_modal.jsx
```diff
@@ -0,0 +1,51 @@
...
+    return <React.Fragment />;
+};
+
+FullPageModal.propTypes = {
+    body               : PropTypes.string,
+    cancel_button_text : PropTypes.string,
+    confirm_button_text: PropTypes.string,
+    onCancel           : PropTypes.func,
+    onConfirm          : PropTypes.func,
+    show               : PropTypes.bool,
```
morteza-binary: @easteregg 
Would you plz follow the [naming conventions style guide](https://github.com/binary-com/binary-static/tree/new-app/src/javascript#naming-conventions-boolean) in order to name boolean variables?

src/javascript/app_2/App/Components/Elements/FullPageModal/full_page_modal.jsx
```diff
@@ -0,0 +1,51 @@
...
+                    <div className='modal-footer'>
+                        <div
+                            className='btn flat effect primary'
+                            onClick={onCancel}
+                        >
+                            <span>{cancel_button_text}</span>
+                        </div>
+                        <div
+                            className='btn flat effect primary'
+                            onClick={() => onConfirm()}
```
morteza-binary: @easteregg 
Since the `onCofirm` is a function and doesn't need to pass any parameters to it, it's sufficient to assign its name to `onClick`;

src/javascript/app_2/App/Components/Elements/AccountSwitcher/account_switcher.jsx
easteregg: I think this component has more than enough responsibilities to be considered a presentational component. already it creates a list for the user (via direct calls to utilities). the nearest container component (`header.jsx`) is far up on the tree and it just creates a chain of props that are just being passed down the chain to get to this component. It does not result in a cleaner code IMO, nor it separates the concerns.

src/javascript/app_2/App/Components/Elements/DenialOfServiceModal/denial_of_service.jsx
easteregg: They don't need to be relatively close to each other in order to work. so refactoring a relative address is always harder IMO.

morteza-binary: @easteregg 
Since both of them are wrapped by a supreme package (App) so it's better to use the relative address.

easteregg: I don't think being wrapped by another package is enough of a reason to make them relative. Can you elaborate on what you mean by that? how is that makes the code better?

src/javascript/app_2/App/Components/Elements/DenialOfServiceModal/denial_of_service.jsx
easteregg: This is a container.

morteza-binary: So shouldn't be in `Containers` directory?

src/javascript/app_2/App/Containers/DenialOfServiceModal/denial_of_service.jsx
```diff
@@ -1,34 +1,40 @@
...
         is_visible={is_visible}
     >
-        <Localize str='You are not allowed to access this feature with your real money account at the moment.' />
+        <Localize str='Sorry, Only virtual accounts can access this feature at the moment.' />
     </FullPageModal>
 );
 
 DenialOfServiceModal.propTypes = {
-    is_visible: PropTypes.bool,
+    show: PropTypes.bool,
```
negarn: same here, don't we want to keep is_visible?

src/javascript/app_2/Stores/Modules/Statement/statement_store.js
```diff
@@ -84,22 +99,18 @@ export default class StatementStore extends BaseStore {
 
     @action.bound
     onMount() {
+        eventBus.listen('ClientAccountHasSwitched', () => {
```
morteza-binary: @easteregg 
is it not better to use [`when`](https://mobx.js.org/refguide/when.html) or [`reaction`](https://mobx.js.org/refguide/reaction.html) instead?

src/javascript/app_2/Stores/client_store.js
```diff
@@ -1,12 +1,299 @@
...
     @observable balance;
-    @observable is_logged_in = !!Client.isLoggedIn();
-    @observable loginid      = Client.get('loginid');
-    @observable currency     = Client.get('currency') || '';
-    @observable upgrade_info = Client.getBasicUpgradeInfo();
-    @observable can_upgrade  = this.upgrade_info.can_upgrade || this.upgrade_info.can_open_multi;
+    @observable storage_key = 'client.accounts';
+
+    /**
+     * Burrowed from `Client_base::getBasicUpgradeInfo()`
```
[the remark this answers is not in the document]
easteregg: RIP English

src/javascript/app_2/Stores/client_store.js
```diff
@@ -1,12 +1,299 @@
...
+    @computed
+    get is_logged_in() {
+        return !!(!isEmptyObject(this.accounts) &&
+            this.loginid &&
+            this.current_account.token);
+    }
+
+    @computed
+    get is_virtual() {
+        return !!this.current_account.is_virtual;
```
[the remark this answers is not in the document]
easteregg: it is not. it is integer.

src/javascript/app_2/Stores/Modules/Statement/statement_store.js
easteregg: Would you elaborate on what you mean by that? using reaction makes these 2 stores to be more coupled together, (let's say we create a "flag" to activate when change happens to client), our statement store and trade store and everything else (that need to do something is going to be reliant on that particular `variable` from the store). Is this going to be a better approach, instead of doing it the pub/sub way?

morteza-binary: by using pub/sub event, those stores are imposed to depend on client_store too. 
BTW the client_store is the store that other stores are based on it since they need login client information. 
IMO, using mobx facilities is more reliable to handle the flow of changes.

src/javascript/app_2/Stores/client_store.js
```diff
@@ -1,12 +1,327 @@
...
+    @observable upgrade_info;
+    @observable accounts;
+    storage_key = 'client.accounts';
+    @observable switched    = '';
+
+    /**
+     * Borrowed from `Client_base::getBasicUpgradeInfo()`
+     * @returns {{type: *, can_upgrade: boolean, can_upgrade_to: *, can_open_multi: boolean}}
+     */
+    getBasicUpgradeInfo = () => {
```
morteza-binary: @easteregg @negar-binary 
isn't it better to keep the business logic for both the legacy and the new app in one place instead of duplicating them?
Whereas the logic of keeping both projects in one repository was to maintain them better and reduce the duplication.
It seems that is better to refactor the Client_base and extract logics which are common between both projects, in this way, there is just one code_base for client logic and both `Client_base` and `client_store` use it.

negarn: makes sense, since we are going to continue keeping them in the same repo for now.

morteza-binary: yes, by extracting common business logic to another place as a black-box functions, both `Client_base` and `client_store` can use them just by passing required parameters to them. Also, there is just one code-base for client business logic for both of them.

src/javascript/app_2/Stores/client_store.js
```diff
@@ -1,12 +1,285 @@
...
+
+    @computed
+    get account_title() {
+        return getAccountTitle(this.loginid);
+    }
+
+    @computed
+    get currency() {
+        return this.accounts[this.loginid] ?
+            this.accounts[this.loginid].currency : '';
```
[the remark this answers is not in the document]
easteregg: Where is the default currency stored other than user accounts? Do you mean a hardcoded `USD`?

src/javascript/app_2/Stores/client_store.js
```diff
@@ -1,12 +1,285 @@
...
+     * @param {string} loginid
+     */
+    @action.bound
+    switchAccount(loginid) {
+        this.switched = loginid;
+    }
+
+    @action.bound
+    switchEndSignal() {
+        this.switch_broadcast = false;
```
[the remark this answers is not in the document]
easteregg: I think the third one is just like the first one, just switching problems. ended up using `when` which actually made a cleaner output. thank you.

src/javascript/app_2/Stores/client_store.js
```diff
@@ -1,12 +1,285 @@
...
-    @observable balance;
-    @observable is_logged_in = !!Client.isLoggedIn();
-    @observable loginid      = Client.get('loginid');
-    @observable currency     = Client.get('currency') || '';
-    @observable upgrade_info = Client.getBasicUpgradeInfo();
-    @observable can_upgrade  = this.upgrade_info.can_upgrade || this.upgrade_info.can_open_multi;
+    @observable loginid;
+    @observable upgrade_info;
+    @observable accounts;
+    storage_key                  = 'client.accounts';
```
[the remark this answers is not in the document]
easteregg: done

src/javascript/app_2/Stores/client_store.js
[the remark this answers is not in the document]
morteza-binary: @easteregg I think the issue is caused by the changes in trade_store by removing the currency property from there. let have a talk about it.

src/javascript/app_2/Stores/client_store.js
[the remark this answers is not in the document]
morteza-binary: @easteregg Actually, they are not as same, let talk about it.

src/javascript/app_2/Stores/Modules/Portfolio/portfolio_store.js
```diff
@@ -1,10 +1,10 @@
...
     action,
     computed,
-    observable }                   from 'mobx';
-import { WS }                      from 'Services';
-import { formatPortfolioPosition } from './Helpers/format_response';
-import BaseStore                   from '../../base_store';
+    observable }                        from 'mobx';
+import { WS }                           from 'Services';
+import { formatPortfolioPosition }      from './Helpers/format_response';
+import BaseStore                        from '../../base_store';
```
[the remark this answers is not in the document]
easteregg: yes.

morteza-binary:
@easteregg Would you plz solve the conflicts?

src/javascript/app_2/Stores/client_store.js
```diff
@@ -1,12 +1,366 @@
...
+
+    @computed
+    get account_list() {
+        return this.all_loginids.map(id => (
+            id !== this.loginid &&
+            !this.isDisabled(id) &&
+            this.getToken(id) ?
+                this.getAccountInfo(id) :
+                undefined
+        )).filter(account => account);
```
cakasuma: i think we can remove this filter since it does nothing ?

src/javascript/app_2/Stores/client_store.js
```diff
@@ -1,12 +1,366 @@
...
-    @observable currency     = Client.get('currency') || '';
-    @observable upgrade_info = Client.getBasicUpgradeInfo();
-    @observable can_upgrade  = this.upgrade_info.can_upgrade || this.upgrade_info.can_open_multi;
+    @observable loginid;
+    @observable upgrade_info;
+    @observable accounts;
+    @observable switched         = '';
+    @observable switch_broadcast = false;
+    @observable currencies_list  = {};
+    @observable selected_currency = '';
```
cakasuma: nitpick, can make the same space for the = symbol

src/javascript/app_2/Stores/base_store.js
```diff
@@ -235,5 +240,44 @@ export default class BaseStore {
...
+    disposeSwitchAccount() {
+        this.switch_account_disposer();
+        this.switch_account_listener = null;
+    }
+
+    @action.bound
+    onUnmount() {
+        this.disposeSwitchAccount();
+    }
+
```
cakasuma: i think, can remove space here as well

src/javascript/app_2/Stores/client_store.js
easteregg: it removes falsies.

PR #4563, opened 2018-11-02
[new-app][WIP] Add keypress events to select dropdown

brucebinary:


src/sass/app_2/_common/form/dropdown.scss
```diff
@@ -92,8 +95,8 @@ div.dropdown-container {
                     display: inline-block;
                     vertical-align: middle;
                 }
-                &:hover {
-                    background-color: $COLOR_BRANDBLUE_5;
+                &:hover, &.highlighted {
```
brucebinary: I felt that there should a difference between hover+highlighted color and selected. UX can decide on a specific color, but I thought this would look fine

zakibinary: On `binary-static` the color of keypress highlighted item and hovered item is the same. I think this is fine, since we shouldn't have to handle the scenario where user's cursor is left situated on the area of dropdown element.

PR #4564, opened 2018-11-02
kelly/telegram_popup

kellybinary:


src/javascript/_common/third_party_links.js
```diff
@@ -11,6 +11,29 @@ const ThirdPartyLinks = (() => {
...
+            // show a popup to remind clients to have Telegram app installed on their device
+            Dialog.confirm({
+                id               : 'telegram',
+                localized_message: localize(['Please ensure that you have the Telegram app installed on your device.', 'Click OK to proceed.']),
+            }).then((should_proceed) => {
+                if (should_proceed) {
+                    openThirdPartyLink(el_link.href);
+                }
+            });
+        }
```
negarn: this has many things in common with clickHandler. perhaps we can refactor them to use a similar function with a few passed in variables? i.e. keep `checkTelegram` and `clickHandler` functions so our event listeners can call a function without passing variables, but inside them call another function that checks the link and handles showing the popup.

src/javascript/app/common/attach_dom/popup.js
```diff
@@ -25,7 +26,19 @@ const callback = (options) => {
...
+    });
+
+    const has_lightbox = document.querySelector('.lightbox');
+    if (has_lightbox) {
+        // store this popup in queue to show it after the previous popup has been removed
+        // to avoid having multiple popup showing at the same time
+        popup_queue.push(lightbox);
+    } else {
+        document.body.appendChild(lightbox);
+    }
```
negarn: I hate having to create a queue of popups, but nice workaround.

negarn: we should show the third party redirect first, then the telegram installation popup. as of now it shows the telegram popup, then opens the new tab, then when you go back to binary tab you see the third party redirect popup.

negarn: not sure if it's better to combine them rather than show two separate ones. can you comment as well @ashkanx

kellybinary: yeah, i think it's makes more sense to show third party redirect first then the telegram installation popup too. better to combine them in this case

PR #4568, opened 2018-11-03
Nazanin/affiliate

nazaninreihani:


src/templates/static/affiliates/signup.jsx
```diff
@@ -13,13 +13,13 @@ const Signup = () => (
         <div className='container'>
             <h1>{it.L('[_1] Affiliate Programme', it.website_name)}</h1>
             <p>
-                {it.L('Earn up to 35% commission with an award-winning binary options trading platform.')}
+                {it.L('Join the Binary.com Affiliate Programme and earn commission by referring new clients to our award-winning binary options trading platform.')}
```
negarn: instead of `Binary.com` please use a placeholder `[_1]` with replacement as `it.website_name`

src/templates/static/affiliates/signup.jsx
```diff
@@ -62,13 +62,14 @@ const Signup = () => (
                     items={[
                         { id: 'revenue_share',   text: it.L('Revenue share') },
                         { id: 'turnover',        text: it.L('Turnover') },
+                        { id: 'cpa_eu_only',     text: it.L('CPA (EU only)') },
```
negarn: why not use our classes to only show this section to EU clients? can you check with Karen?

src/templates/static/affiliates/signup.jsx
```diff
@@ -148,6 +149,13 @@ const Signup = () => (
                             </div>
                             <p className='mobile-text-small'>{it.L('All commissions are credited into your account by the 15th of every month.')}</p>
                         </TabContent>
+                        <TabContent id='cpa_eu_only'>
+                            <h3 className='gr-padding-20'>
+                                <span>{it.L('Cost per acquisition (CPA) for EU affiliates only')}</span>
+                            </h3>
+                            <p className='mobile-text-small'>{it.L('Earn USD 100 in commission for each successful referral. Your referred client must open a <span className=\'bold\'>real money investment account</span> through your unique affiliate link and deposit a total of USD 100 or more (one-time or accumulative) in the account. This commission plan is only available to affiliates based in the EU.')}</p>
```
negarn: instead of `<span className=\'bold\'>` why not use `<strong>`? also please don't include HTML tags in translation strings, better use placeholders for them

src/templates/static/affiliates/signup.jsx
nazaninreihani: I talked to Karen and she said we can show it to all users, as it says in the title that it's EU only.

PR #4569, opened 2018-11-04
[new-app] Display an error message when active_symbols is empty

easteregg:


src/javascript/app_2/Stores/Modules/Trading/trade_store.js
```diff
@@ -112,18 +116,26 @@ export default class TradeStore extends BaseStore {
...
     async prepareTradeStore() {
         let query_string_values = this.updateQueryString();
-        this.smart_chart = this.root_store.modules.smart_chart;
-        const active_symbols = await WS.activeSymbols();
+        this.smart_chart        = this.root_store.modules.smart_chart;
+        const active_symbols    = await WS.activeSymbols();
+
+        if (!active_symbols.active_symbols || active_symbols.active_symbols.length === 0) {
+            runInAction(() => {
+                this.root_store.common.has_error     = true;
```
morteza-binary: @easteregg 
isn't it better to change the `root_store` properties by itself instead of changing them directly in the other store?
IMO, it's better to define an action for it and just invoke it here without any `runInAction` higher-order function.

easteregg: You are right, I'll push a new one for this.

PR #4574, opened 2018-11-05
kelly/amend_text

kellybinary:
- amend text in authentication page


src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -127,6 +127,8 @@ const Authenticate = (() => {
         const $files             = $not_authenticated.find('input[type="file"]');
         $button = $not_authenticated.find('#btn_submit');
 
+        hideSuccess();
```
negarn: better to call it elsewhere, otherwise it gets called everywhere that wants to disable the button, what do you think?

src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -440,11 +442,18 @@ const Authenticate = (() => {
             [`<a href="${Url.urlFor('contact')}">`, '</a>']);
         displayNotification(msg, false, 'document_under_review');
         setTimeout(() => {
-            $('#not_authenticated, #not_authenticated_financial').setVisibility(0); // Just hide it
+            removeButtonLoading();
+            $button.setVisibility(0);
```
negarn: not sure if we want to hide the button, why not show disabled button? user might think they're not allowed to submit anymore

src/javascript/app/pages/user/account/authenticate.js
kellybinary: yeap i agree, i've updated the location :)

PR #4575, opened 2018-11-05
kelly/cashier_note

kellybinary:
- add Cashier note.


src/templates/app/cashier/index.jsx
```diff
@@ -20,6 +20,12 @@ const DepositWithdraw = ({ id, is_payment_agent, show_upgrade }) => (
     </div>
 );
 
+const CashierNote = ({ text }) => (
+    <div className='invisible client_logged_in cashier-note'>
+        <p>{it.L(text)}</p>
```
negarn: we can't pass variables to `it.L` in this way. they need to be passed as translated text.

src/templates/app/cashier/index.jsx
```diff
@@ -85,9 +93,9 @@ const Cashier = () => (
                         <img className='responsive' id='payment_agent_icon' src={it.url_for('images/pages/cashier/payment-agents.svg')} />
                     </a>
                 </div>
-                <div className='gr-5 gr-12-m'>
+                <div className='gr-6 gr-12-m'>
                     <span>{it.L('For e-wallets or local currencies not supported by [_1].', it.website_name)}</span>
-                    <p className='faded'>{it.L('Note: Withdrawal via payment agent is available only if you deposit exclusively via payment agent')}</p>
+                    <CashierNote text='Withdrawal via payment agent is available only if you deposit exclusively via payment agent.' />
```
negarn: it needs to be `text={it.L('....')}`

src/templates/app/cashier/index.jsx
```diff
@@ -64,12 +70,14 @@ const Cashier = () => (
...
-                <div className='gr-5 gr-12-m'>
+                <div className='gr-6 gr-12-m'>
                     <span className='invisible normal_currency client_logged_out'>{it.L('Deposit or withdraw to your account via bank wire, credit card, or e-wallet.')}</span>
                     <span className='invisible crypto_currency'>{it.L('Manage the funds in your cryptocurrency account.')}</span>
+                    &nbsp;
                     <a className='invisible normal_currency client_logged_out' href={it.url_for('cashier/payment_methods')} id='view_payment_methods'>
-                        <p>{it.L('View available payment methods')}</p>
+                        <span>{it.L('View available payment methods')}</span>
                     </a>
+                    <CashierNote text='Sharing your payment method with another client is prohibited and can cause delays in your withdrawals.' />
```
negarn: same here

src/sass/_common/common.scss
```diff
@@ -791,3 +791,19 @@ label + input {
...
+
+.cashier-note {
+    background-color: $COLOR_LIGHT_GRAY;
+    color: $COLOR_BLUE;
+    display: inline-flex;
+    align-items: center;
+    margin: 1em 0;
+
+    &:before {
+        content: 'ⓘ';
```
negarn: can you check the class we have for `notice-circle` and see how that's implemented? would it be better to reuse that instead of adding a second class for this info circle?

src/templates/app/cashier/index.jsx
```diff
@@ -85,9 +104,9 @@ const Cashier = () => (
                         <img className='responsive' id='payment_agent_icon' src={it.url_for('images/pages/cashier/payment-agents.svg')} />
                     </a>
                 </div>
-                <div className='gr-5 gr-12-m'>
+                <div className='gr-6 gr-12-m'>
                     <span>{it.L('For e-wallets or local currencies not supported by [_1].', it.website_name)}</span>
-                    <p className='faded'>{it.L('Note: Withdrawal via payment agent is available only if you deposit exclusively via payment agent')}</p>
+                    <CashierNote text={it.L('Withdrawal via payment agent is available only if you deposit exclusively via payment agent.')} />
```
negarn: doesn't that mean this will be hidden for all except those four countries too? do we want that?

kellybinary: let me check

kellybinary: @negarn we hide this except for those 4 countries too

src/javascript/app/pages/cashier/cashier.js
```diff
@@ -28,6 +29,14 @@ const Cashier = (() => {
             $toggler.children().toggleClass('active');
             $toggler.toggleClass('open');
         });
+        BinarySocket.wait('website_status').then(() => {
+            $('.cashier_note').setVisibility(
+                Client.isLoggedIn() &&                       // only show to logged-in clients
+                !Client.get('is_virtual') &&                 // only show to real accounts
+                !isCryptocurrency(Client.get('currency')) && // only show to fiat currencies
+                /^(vn|ng|lk|id)$/.test(State.getResponse('website_status.clients_country')) // only show to Vietnam, Nigeria, Sri Lanka, Indonesia
```
negarn: you should be checking the residence probably

PR #4596, opened 2018-11-08
[new-app] add bom_ui variable to differentiate between old and new ui

ruskakimov:


src/javascript/_common/base/gtm.js
```diff
@@ -94,6 +94,7 @@ const GTM = (() => {
         const req  = response.echo_req.passthrough;
         const data = {
             event             : 'buy_contract',
+            bom_ui            : 'old',
```
ashkanx:
```suggestion
            bom_ui            : 'legacy',
```

PR #4598, opened 2018-11-08
[new-app] Morteza/pwa

morteza-binary:


index.html
```diff
@@ -1,6 +1,9 @@
 <!DOCTYPE html>
 <html>
     <head>
+        <!-- Prompt a message in the browser if users disabled JS -->
+        <noscript>Your browser does not support JavaScript!</noscript>
```
ashkanx: should we translate this?

binary-hamedanchi: We can't localize texts here as there can be only one `/index.html` file (in root).

morteza-binary: Since the service_worker's start_url has been set to `<lang>/app/index.html`, it doesn't need any more to have this tag here in this page. it will be moved to `src/templates/app_2/head.jsx` since there is accessibility to the localization method there.

build/webpack/plugins.js
```diff
@@ -16,6 +20,30 @@ const getPlugins = (app, grunt) => ([
...
+                fileName: 'asset-manifest.json',
+            }),
+
+            ...(languages.map(lang => new SWPrecachedWebpackPlugin({
+                cachedId                 : 'app_2',
+                dontCacheBustUrlsMatching: /\.\w{8}\./,
+                minify                   : false,
+                navigateFallback         : `${lang.toLowerCase()}/app/index.html`,
+                filepath                 : path.resolve(PATHS.DIST, `${lang.toLowerCase()}/app/service-worker.js`),
+                // TODO uncomment the below lines when this card(https://trello.com/c/FHvQREm8) has been done.
```
negarn: should we reference the open source PR instead of the private trello card?

src/javascript/app_2/App/Components/Layout/Header/install_pwa_button.jsx
```diff
@@ -0,0 +1,38 @@
...
+                    if (choice_result.outcome === 'accepted') {
+                        onClick();
+                    }
+                });
+        }
+    };
+
+    return (
+        <Button
+            className='primary orang'
```
kellybinary: this looks like a typo: should be `orange`

src/javascript/app_2/Utils/pwa/register_service_worker.js
```diff
@@ -0,0 +1,43 @@
+export default function register () { // Register the service worker
+    if (/* process.env.NODE_ENV === 'production' && */ 'serviceWorker' in navigator) {
+        window.addEventListener('load', () => {
+            let path_name = window.location.pathname;
+            path_name = /index\.html/g.test(path_name) ? window.location.pathname.replace('index.html', '') : '';
+            const swUrl = `${path_name}service-worker.js`;
```
OskarAhl: should this be `sw_url`?

morteza-binary:
```suggestion
            const sw_url = `${path_name}service-worker.js`;
```

PR #4614, opened 2018-11-12
Nazanin/ladder_page_prompt_trade

nazaninreihani:


src/javascript/app/pages/mb_trade/mb_notifications.js
```diff
@@ -2,6 +2,8 @@
  * Notifications manages various notification messages
  *
  */
+const Login           = require('../../../_common/base/login');
+const CommonFunctions = require('../../../_common/common_functions');
```
negarn: these should be above the comment, on lines 1 and 2

PR #4639, opened 2018-11-15
add domain exceptions

aaron-binary:


src/javascript/_common/url.js
```diff
@@ -58,8 +58,11 @@ const Url = (() => {
 
     const default_domain = 'binary.com';
     const host_map = { // the exceptions regarding updating the URLs
-        'bot.binary.com'       : 'www.binary.bot',
-        'developers.binary.com': 'developers.binary.com', // same, shouldn't change
+        'bot.binary.com'      : 'www.binary.bot',
+        'developers.binary.me': 'developers.binary.com', // same, shouldn't change
+        'academy.binary.me'   : 'academy.binary.com',
```
binary-hamedanchi: The keys are the actual link we have in the codebase, and there is no `*.me` link in the codebase.
Please revert to the [first commit](https://github.com/binary-com/binary-static/pull/4639/commits/cbc43b3a56df878c06371b04cbe17ab17d78a785) as that was correct. The production issue is because this is not released yet!

PR #4644, opened 2018-11-16
Aaron/Add CTA

aaron-binary:


src/javascript/app/pages/trade/tradepage.js
```diff
@@ -57,6 +58,21 @@ const TradePage = (() => {
...
                 }
             });
+
+            BinarySocket.wait('website_status', 'landing_company').then(() => {
+                const is_logged_in = Client.isLoggedIn();
+                const is_eu = isEuCountry();
+                const is_malta = State.getResponse('landing_company.financial_company.shortcode') === 'maltainvest';
+
+                if ((!is_logged_in && is_eu) || (is_logged_in && is_malta)) {
+                    if ((is_logged_in && is_eu) || (is_logged_in && is_malta)) {
```
negarn: no, we only want to show this CTA to MF accounts, meaning that the boolean `is_maltainvest` is enough, as it will return false for logged out clients and only be returning true for authorized MF clients.

src/javascript/app/pages/trade/tradepage.js
```diff
@@ -57,6 +58,21 @@ const TradePage = (() => {
...
+                        $('#professional-cta').setVisibility(1);
+                    }
+                    $('.mfsa_message')
+                        .removeClass('container')
+                        .attr('style', 'margin-bottom: 40px'); // hello
```
negarn: probably meant to remove this comment? 🙂

negarn: we don't want to add the styling in javascript, it becomes hard to track. it's better to add a class that has margin-bottom: 40px defined in scss file and simply add that class here instead.

src/templates/_common/includes/mfsa.jsx
```diff
@@ -7,7 +7,8 @@ const MFSA = () => {
     return (
         <div style={style} className='container mfsa_message'>
             <div className='notice-msg center-text gr-parent gr-child'>
-                <p>{it.L('In the EU, financial binary options are only available to professional investors.')}</p>
+                <p>{it.L('In the EU, financial binary options are only available to professional investors.')} <a id='professional-cta' className='invisible' href={it.url_for('user/settings/professional')}>{it.L('Apply now as a professional investor')}</a>
```
negarn: would be better to use `&nbsp;` instead of the white space before `<a`

aaron-binary: Do HTML entities get rendered in JSX components?

negarn: can you look through the code and find the answer to that? 🙂

src/templates/_common/includes/mfsa.jsx
```diff
@@ -7,7 +7,8 @@ const MFSA = () => {
     return (
         <div style={style} className='container mfsa_message'>
             <div className='notice-msg center-text gr-parent gr-child'>
-                <p>{it.L('In the EU, financial binary options are only available to professional investors.')}</p>
+                <p>{it.L('In the EU, financial binary options are only available to professional investors.')}&nbsp;<a id='professional-cta' className='invisible' href={it.url_for('user/settings/professional')}>{it.L('Apply now as a professional investor')}</a>
```
ashkanx: better to have a CTA for `Apply now as a professional investor` , please get help from UI team

src/javascript/app/pages/trade/tradepage.js
```diff
@@ -57,6 +58,22 @@ const TradePage = (() => {
                     if ($currency.next().attr('id') === $currency.attr('id')) $currency.next().eq(0).remove();
                 }
             });
+
+            BinarySocket.wait('website_status', 'landing_company').then(() => {
+                const is_logged_in = Client.isLoggedIn();
+                const is_eu = isEuCountry();
+                const is_maltainvest = Client.get('landing_company_shortcode') === 'maltainvest';
+
+                if (is_maltainvest) {
```
ashkanx: @aaron-binary @negar-binary  we also need to keep this hidden for `MF professionals`

negarn: yes, thank you

src/javascript/app/pages/trade/purchase.js
```diff
@@ -63,17 +65,37 @@ const Purchase = (() => {
...
-                    additional_message = localize('Try our [_1]Volatility Indices[_2].', [`<a href="${urlFor('get-started/binary-options', 'anchor=volatility-indices#range-of-markets')}" >`, '</a>']);
-                } else if (/Random/.test(error.code)) {
-                    additional_message = localize('Try our other markets.');
+            BinarySocket.wait(...required_api_calls).then(() => {
+                const is_eu = isEuCountry();
+                const is_maltainvest = Client.get('landing_company_shortcode') === 'maltainvest';
+                const is_no_mfp_client_error = /NoMFProfessionalClient/.test(error.code);
+               
+                if (is_eu && is_no_mfp_client_error) {
+                    if (is_logged_in && is_maltainvest) {
```
negarn: if we are checking the error code `NoMFProfessionalClient` from API, I don't think we need to check `is_logged_in` and `is_maltainvest` anymore, so we should be able to remove this line entirely.

src/javascript/app/pages/trade/purchase.js
```diff
@@ -63,17 +65,37 @@ const Purchase = (() => {
...
-                if (/FinancialBinaries/.test(error.code)) {
-                    additional_message = localize('Try our [_1]Volatility Indices[_2].', [`<a href="${urlFor('get-started/binary-options', 'anchor=volatility-indices#range-of-markets')}" >`, '</a>']);
-                } else if (/Random/.test(error.code)) {
-                    additional_message = localize('Try our other markets.');
+            BinarySocket.wait(...required_api_calls).then(() => {
+                const is_eu = isEuCountry();
+                const is_maltainvest = Client.get('landing_company_shortcode') === 'maltainvest';
+                const is_no_mfp_client_error = /NoMFProfessionalClient/.test(error.code);
+               
+                if (is_eu && is_no_mfp_client_error) {
```
negarn: perhaps we don't need to check `is_eu`, since the error code `NoMFProfessionalClient` should only be coming from API for MF non-professional clients, which means they are definitely EU.

src/javascript/app/pages/trade/purchase.js
```diff
@@ -63,17 +65,37 @@ const Purchase = (() => {
...
+                    if (is_logged_in && is_maltainvest) {
+                        const is_professional = (State.getResponse('get_account_status.status') || []).includes('professional');
+                        if (!is_professional) {
```
negarn: same logic here, if the API is already telling us that client can't purchase this client because of `NoMFProfessionalClient`, i.e. current client is a non-professional MF, we don't need to check it again in FE if current client is professional of not, so this if condition can be removed entirely as well.

src/javascript/app/pages/trade/tradepage.js
```diff
@@ -57,6 +59,32 @@ const TradePage = (() => {
                     if ($currency.next().attr('id') === $currency.attr('id')) $currency.next().eq(0).remove();
                 }
             });
+
+            BinarySocket.wait('website_status', 'landing_company').then(() => {
+                const is_logged_in = Client.isLoggedIn();
+                const is_eu = isEuCountry();
+                const is_maltainvest = Client.get('landing_company_shortcode') === 'maltainvest';
+                const show_msfa_message = () => {
```
negarn: function names should be camelCase

src/javascript/app/pages/trade/tradepage.js
```diff
@@ -57,6 +59,32 @@ const TradePage = (() => {
...
+                        .addClass('margin-bottom-40')
+                        .slideDown(300);
+                };
+                
+                if (is_eu) {
+                    if (is_logged_in && is_maltainvest) {
+                        BinarySocket.wait('get_account_status').then((response) => {
+                            const is_professional = (getPropertyValue(response, ['get_account_status', 'status']) || []).includes('professional');
+                            if (!is_professional) {
+                                $('#professional-cta').setVisibility(1);
```
negarn: I suggest using `client_logged_in` class name instead of this ID for the CTA element so we can remove this line and let the common code handle showing/hiding the CTA

aaron-binary: Could you elaborate? The _client\_logged\_in_ class should show elements for clients that are logged in, in this case we only want to show the CTA to logged in, non-professionals, if we would use the _client\_logged\_in_ it would still show for logged in professionals?

src/javascript/app/pages/trade/purchase.js
```diff
@@ -63,15 +63,20 @@ const Purchase = (() => {
             container.style.display = 'block';
             message_container.hide();
             confirmation_error.show();
-            let message = error.message;
-            if (/RestrictedCountry/.test(error.code)) {
+
+            let message;
+            if (/NoMFProfessionalClient/.test(error.code)) {
+                message = localize('[_1][_2]In the EU, financial binary options are only available to professional investors.[_3][_4][_5]Apply now as a professional investor[_6][_7][_8]', ['<div class="gr-row font-style-normal">', '<div class="gr-12 gr-padding-20">', '</div>', '<div class="gr-12 gr-padding-20">', `<a id="btn_professional" class="button" href="${urlFor('user/settings/professional')}"><span>`, '</span></a>', '</div>', '</div>']);
```
negarn: I'm confused, why not just combine placeholders together? e.g. `[_1][_2]` can become `[_1]` and replaced by `'<div class="gr-row font-style-normal"><div class="gr-12 gr-padding-20">'` etc so translators' job becomes easier as well.

We can even separate them to be completely without placeholders, i.e. make the divs using jquery elements, then set the inner text to be the localized sentence 1 and then localized sentence 2.

src/javascript/app/pages/trade/tradepage.js
```diff
@@ -69,10 +71,19 @@ const TradePage = (() => {
...
                 if (isEuCountry()) {
                     const isMaltainvest = () => Client.get('landing_company_shortcode') === 'maltainvest';
-                    const isProfessional = () => (getPropertyValue(response, ['get_account_status', 'status']) || []).includes('professional');
+                    const account_status = (getPropertyValue(response, ['get_account_status', 'status']) || []);
+                    const isProfessional = () => account_status.includes('professional');
+                    const hasRequestedProfessional = () => account_status.includes('professional_requested');
                     // show MFSA message to MF non-professional clients or logged out EU clients
                     if (!is_logged_in || (isMaltainvest() && !isProfessional())) {
-                        $('.mfsa_message')
+                        const mfsa_message = $('.mfsa_message');
```
negarn: for jquery elements we name them starting with `$` to make it obvious

PR #4647, opened 2018-11-16
zaki/view_popup_changes

zakibinary:
Changelog
- Added extra condition to show message when `contract.sell_price` is not available and contract has not yet ended.
- Update label for `profit/loss` depending on contract's status.


src/javascript/app/pages/user/view_popup/view_popup.js
```diff
@@ -244,6 +244,10 @@ const ViewPopup = (() => {
             containerSetText('trade_details_profit_loss', '-', { class: 'loss' });
         }
 
+        if (!contract.sell_price && is_ended) {
```
negarn: I would rather update the above if else to also include this third condition, rather than override it at the end. something like

```
if (!is_ended && final_price) {
...
} else if (is_ended && !contract.sell_price) {
...
} else {
...
}
```

src/javascript/app/pages/user/view_popup/view_popup.js
```diff
@@ -235,7 +235,7 @@ const ViewPopup = (() => {
         containerSetText('trade_details_ref_id', `${contract.transaction_ids.buy} (${localize('Buy')}) ${contract.transaction_ids.sell ? `<br>${contract.transaction_ids.sell} (${localize('Sell')})` : ''}`);
         containerSetText('trade_details_indicative_price', indicative_price ? formatMoney(contract.currency, indicative_price) : '-');
 
-        if (!is_ended && final_price) {
+        if (final_price) {
```
negarn: won't this always be true? because we will always have indicative price, so original issue doesn't get resolved

PR #4653, opened 2018-11-19
Nazanin/remove_dai

nazaninreihani:


negarn:
we need to also remove the DAI symbol from our fonts @nazaninreihani

PR #4654, opened 2018-11-20
zaki/testing_fixes

zakibinary:
Changelog
- Fixed EU Popups not triggering for third party links on map in regulation page.
- Updated text for staff information and updated offices section in `about-us`.
- Removed link to`whatbrowser.org` as it will be removed soon. Replaced with text and link recommending clients to use/download `Google Chrome`.


src/templates/static/about/index.jsx
```diff
@@ -116,14 +116,14 @@ const Index = () => (
...
+                            <BoxInner className='border-right-top' href='/careers' image='staff' text={it.L('Over [_1] <a href=\'[_2]\'>staff</a> and contractors worldwide', '200', it.url_for('careers'))} />
                             <BoxInner image='1mil' text={it.L('Over 1 million registered accounts worldwide')} />
                         </Box>
                     </div>
                     <div className='gr-row gr-parent'>
                         <Box>
                             <BoxInner className='border-right-bottom gr-padding-30' image='transacts' text={it.L('Binary\'s platform transacts on average [_1] [_2]transactions[_3] per second, 24/7', '20', '<a href=\'https://binarycom.statuspage.io/\' target=\'_blank\' rel=\'noopener noreferrer\'>', '</a>')} />
-                            <BoxInner className='border-right-bottom gr-padding-30' href='/contact' image='locations' text={it.L('<a href=\'[_1]\'>Offices</a> in Malta and Malaysia', it.url_for('contact'))} />
+                            <BoxInner className='border-right-bottom gr-padding-30' image='locations' text={it.L('Offices in Malta and Malaysia')} />
```
negarn: why not just link it to the correct pages instead?

zakibinary: @negarn , do you mean the office locations section in the career page? Previously I was thinking of using this but it seemed a bit out of context for regular clients after discussing with @ashkanx .

PR #4655, opened 2018-11-20
zaki/new_window_logout_fix

zakibinary:
…form

### DIFF
--- a/src/javascript/app/base/client.js
+++ b/src/javascript/app/base/client.js
@@ -47,6 +47,12 @@ const Client = (() => {
             applyToAllElements('.client_logged_out', (el) => {
                 el.setVisibility(1);
             }, '', el_section);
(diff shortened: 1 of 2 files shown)


src/javascript/app/base/client.js
```diff
@@ -47,6 +47,15 @@ const Client = (() => {
             applyToAllElements('.client_logged_out', (el) => {
                 el.setVisibility(1);
             }, '', el_section);
+            applyToAllElements('.client_logged_in', (el) => {
+                el.setVisibility(0);
+            }, '', el_section);
+            applyToAllElements('#client-logged-in', (el) => {
+                el.setVisibility(0);
+            }, '', el_section);
+            applyToAllElements('#topbar-msg', (el) => {
```
zakibinary: There's probably only one instance of `#topbar-msg` element so probably don't need to use `applyToAllElements`.

PR #4671, opened 2018-11-22
[new-app] easteregg/fix-start-date-value-on-page-reload

easteregg:


src/javascript/app_2/Modules/Trading/Components/Form/TradeParams/start_date.jsx
```diff
@@ -42,7 +42,7 @@ const StartDate = ({
         >
             <Dropdown
                 name='start_date'
-                value={start_date}
+                value={parseInt(start_date)}
```
morteza-binary: @easteregg 
What is your idea if we use the syntactic sugar `+start_data` instead of parseInt?

src/javascript/app_2/App/Components/Form/time_picker.jsx
```diff
@@ -239,11 +239,8 @@ TimePicker.propTypes = {
     padding        : PropTypes.string,
     placeholder    : PropTypes.string,
     sessions       : MobxPropTypes.arrayOrObservableArray,
-    start_date     : PropTypes.oneOfType([
-        PropTypes.number,
-        PropTypes.string,
-    ]),
-    value: PropTypes.string,
+    start_date     : PropTypes.number,
```
morteza-binary: @easteregg 
AFAIK, the value of `start_date` is not always as a number, sometimes the API returns it as a string.
IMO, it's better to handle it in the store and keep it as a number instead of handling it in components.

@negarn Would you plz confirm that I'm right or not since you added that propTypes before in [this commit](https://github.com/binary-com/binary-static/commit/a8d55766792ed98d7b9f715987257a4a06878625#diff-ad9770f97c3ed3130087a62755367212R243)?

negarn: your suggestion sounds better, I don't remember why string was allowed before.

src/javascript/app_2/Modules/Trading/Components/Form/TradeParams/start_date.jsx
easteregg: I prefer `parseInt` it covers more cases and it is more strict comparing to unary plus operator.

src/javascript/app_2/App/Components/Form/time_picker.jsx
easteregg: @morteza-binary Which API call returns `start_date`?

morteza-binary: @easteregg 
Actually, there isn't a direct API call for it, the response of `contracts_for` contains `forward_starting_options` that we use it to create `trade_store.start_dates_list`. 
It seems that the best place to fix the issue is in [line 15 of `start_date.js`](https://github.com/binary-com/binary-static/blob/master/src/javascript/app_2/Stores/Modules/Trading/Helpers/start_date.js#L15) just by adding a `+` before `option.date`.

src/javascript/app_2/Stores/Modules/Trading/trade_store.js
```diff
@@ -1,39 +1,33 @@
...
+    isCryptocurrency }                                                        from '_common/base/currency_base';
+import BinarySocket                                                           from '_common/base/socket_base';
+import { localize }                                                           from '_common/localize';
+import { cloneObject, isEmptyObject }                                         from '_common/utility';
+import { WS }                                                                 from 'Services';
+import GTM                                                                    from 'Utils/gtm';
+import URLHelper                                                              from 'Utils/URL/url_helper';
+import { processPurchase }                                                    from './Actions/purchase';
+import * as Symbol                                                            from './Actions/symbol';
+import { allowed_query_string_variables, non_proposal_query_string_variable } from './Constants/query_string';
```
morteza-binary: @easteregg would you plz reformat the import statement?

PR #4685, opened 2018-11-26
zaki/logout_invalid_token

zakibinary:
Changelog
- Logout / Reload page when token is no longer valid.


src/javascript/app/pages/user/reality_check/reality_check.ui.js
```diff
@@ -122,6 +122,7 @@ const RealityCheckUI = (() => {
             BinarySocket.send({ reality_check: 1 }).then((response) => {
                 if (response.error && !/user\/statementws\.html/.test(window.location.pathname)) { // don't block statement page for reality check error, but block all other pages
                     $('#content').empty().html($('<div/>', { class: 'container' }).append($('<p/>', { class: 'notice-msg center-text', text: response.error.message })));
+                    window.location.reload();
```
negarn: won't we get stuck in an endless loop here? e.g. if server is down etc it will keep reloading the page on user's behalf

src/javascript/app/pages/user/view_popup/view_popup.js
```diff
@@ -767,6 +767,9 @@ const ViewPopup = (() => {
         if (response.error) {
             if (response.error.code !== 'AlreadySubscribed' && response.echo_req.contract_id === contract_id) {
                 showErrorPopup(response, response.error.message);
+                if (response.error.code === 'InvalidToken') {
```
negarn: I think it should be more general, can't we check this in socket file for response of any call and refresh it there?

src/javascript/app/pages/user/reality_check/reality_check.ui.js
zakibinary: Noted, I'll add a condition to check if token is still valid / exists before triggering the reload.

negarn: I'll revert it for now since we want to do a release today

PR #4686, opened 2018-11-26
kelly/msedge_account_balance

kellybinary:
- increase z-index to display account_balance div in MSEdge


src/sass/app/trade.scss
```diff
@@ -467,6 +467,7 @@
             display: flex;  /* fix for iOS */
             justify-content: center;
             align-items: center;
+            z-index: 9; /* fix for MSEdge */
```
zakibinary: Not sure if we need the comments here, since it's going to be applied on all browsers either way.

PR #4687, opened 2018-11-26
Gtm account switch event

ruskakimov:


src/javascript/_common/base/gtm.js
```diff
@@ -41,8 +41,9 @@ const GTM = (() => {
 
     const eventHandler = (get_settings) => {
         if (!isGtmApplicable()) return;
-        const is_login       = localStorage.getItem('GTM_login')       === '1';
-        const is_new_account = localStorage.getItem('GTM_new_account') === '1';
+        const is_login          = localStorage.getItem('GTM_login')       === '1';
+        const is_account_switch = localStorage.getItem('GTM_login')       === '2';
```
easteregg: I think the `1` is considered as a flag (0/1). If my guess is right, I think it would be better to create a totally new flag for `is_account_switch`, like `GTM_switch` or something.

binary-hamedanchi: Yes, using 1, 2, ... leads to confusion.

@ruskakimov As we discussed yesterday, I still suggest to rename the `flag` [here](https://github.com/binary-com/binary-static/pull/4687/files#diff-e95072729306fd8884ac7dfff2323782R174) to `action` or `event`, and send the actual event name (`log_in`, `account_switch`) where it's been called.

ashkanx: yes, better to send the actual event name.

PR #4697, opened 2018-11-27
[new-app] kelly/update_query_string

kellybinary:
- add `start_time` to query string and session storage when `contract_start_type` is forward
- add validations for `start_time`:
  - validate if time is in the past
  - validate if time is in the correct format `HH:MM`
  - validate if hour is between 0 and 23
  - validate if minute is between 0 and 59
- refactor timepicker and datepicker to use `toMoment` from `Utils/Date`


src/javascript/app_2/Stores/base_store.js
```diff
@@ -235,5 +235,15 @@ export default class BaseStore {
...
     }
-}
 
+    /**
+     * Update storage properties.
+     *
+     */
+    @action
+    updateStorageProperties(storage_name, properties) {
+        const storage_name_propertis = `${storage_name}_properties`;
```
negarn: typo, missing `e`

src/javascript/app_2/Stores/Modules/Trading/trade_store.js
```diff
@@ -355,13 +362,24 @@ export default class TradeStore extends BaseStore {
         this.processNewValuesAsync({ barrier_1, barrier_2 }, true);
     }
 
+    @action.bound
+    getQueryStringVariables(query_strings_list) {
+        return this.contract_start_type === 'forward' ? [...query_strings_list, 'start_time'] : query_strings_list;
```
negarn: can we make the start time pretty instead of `12%3A30`, `12:30`

morteza-binary:
@kellybinary 
IMO, we can handle it much simpler by below changes: 

__1. Add/Remove `start_time` from QueryString:__
By converting the `non_proposal_query_string_variable` in `query_string.js` to a function that gets the current state of the store and decides the return array should contain `start_time` or not; the function is something like the one you have implemented and called it `getQueryStringVariables` in `trade_store.js`.

__2. Add/Remove `start_time` from session storage:__
As you know, `BaseStore` just saves the latest state of the store in session/local storage based on the properties which are defined in its constructor for each particular storage so it is better to change the state of the store properly and assign `null` to `start_time` when `start_date` value is `0` (or `start_contract_tyep` value is `spot`).
In this way, we have a correct state of the store without nonsense values.

Please let me know your opinion about this solution.

src/javascript/app_2/Stores/Modules/Trading/trade_store.js
```diff
@@ -120,6 +121,15 @@ export default class TradeStore extends BaseStore {
                 this.changeDurationValidationRules();
             },
         );
+        // Adds intercept to update query strings and session storage properties 'start_time with '00:00'
+        reaction(
```
morteza-binary: @kellybinary 
we can set `start_time` to `null` when `contract_start_type` is `spot` in `processNewValuesAsync` function instead of adding reaction.
If `start_time` is changed it in `processNewValuesAsync`, it'll prevent react from doing extra render when `contract_start_type` is changed from `forward` to `spot` because both of changes are done in one transaction. Also, we keep all the logic in one place.

src/javascript/app_2/Stores/Modules/Trading/Constants/query_string.js
```diff
@@ -1,24 +1,34 @@
...
-    'contract_type',
-    'duration',
-    'duration_unit',
-    'expiry_date',
-    'expiry_type',
-    'last_digit',
-    'start_date',
-    'symbol',
-];
+export const getAllowedQueryStringVariables = (start_date) => {
```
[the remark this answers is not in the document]
kellybinary: sure, we can pass store

src/javascript/app_2/Stores/Modules/Trading/trade_store.js
```diff
@@ -71,7 +71,7 @@ export default class TradeStore extends BaseStore {
     // Start Time
     @observable start_date       = Number(0); // Number(0) refers to 'now'
     @observable start_dates_list = [];
-    @observable start_time       = '12:30';
```
cakasuma: how about expiry_time still `9:40` , should it be changed to null as well ?

kellybinary: @cakasuma let's work on `expiry_time` in another PR, i think it'll require more work than just changing it to `null`  🙂

src/javascript/app_2/Stores/Modules/Trading/Constants/validation_rules.js
```diff
@@ -31,6 +35,19 @@ const getValidationRules = () => ({
...
+            ['custom' , { func: (value, options, store) => store.contract_start_type === 'spot' || isTimeValid(value), message: localize('Please enter the start time in the format "HH:mm".') }],
+            ['custom' , { func: (value, options, store) => {
+                if (store.contract_start_type === 'spot') return true;
+                if (!isTimeValid(value)) return false;
+                const start_moment       = toMoment(store.start_date);
+                const start_moment_clone = start_moment.clone();
+                const [h, m] = value.split(':');
+                return isSessionAvailable(store.sessions, start_moment_clone.hour(h).minute(m), start_moment);
+            }, message: localize('Start time cannot be in the past.') }],
+        ],
```
morteza-binary: @kellybinary It should have a custom rule to check that the value of hour and minute are respectively between 0-23 and 0-59.

kellybinary: @morteza-binary done :)

src/javascript/app_2/Utils/Date/date_time.js
```diff
@@ -67,3 +80,27 @@ export const formatDuration = (duration) => {
...
+ * return true if the time_str is in "HH:MM" format, else return false
+ * @param {String} time_str time
+ */
+export const isTimeValid = time_str => /^(\d{1,2}):(\d{2})(:00)?$/.test(time_str);
+
+/**
+ * return true if the time_str's hour is between 0 and 23, else return false
+ * @param {String} time_str time
+ */
+export const isHourValid = time_str => /^(0[0-9]|1[0-9]|2[0-3])$/.test(time_str.split(':')[0]);
```
morteza-binary:
```suggestion
export const isHourValid = time_str => /^([01][0-9]|2[0-3])$/.test(time_str.split(':')[0]);
```

PR #4701, opened 2018-11-28
Nazanin/app_2_modules_utils_tests

nazaninreihani:


src/javascript/app_2/Stores/Modules/Trading/Helpers/__tests__/barrier.js
```diff
@@ -0,0 +1,131 @@
+import { expect }              from 'chai';
+import React                   from 'react';
+import { buildBarriersConfig } from '../barrier';
+
+describe('buildBarriersConfig', () => {
+    it('Returns Undefined it contract has no barriers', () => {
```
ashkanx:
```suggestion
    it('Returns Undefined if contract has no barriers', () => {
```

src/javascript/app_2/Stores/Modules/SmartChart/Helpers/__tests__/barriers.js
```diff
@@ -0,0 +1,29 @@
...
+    describe('isBarrierSupported', () => {
+        it('should return false when barrier is not in CONTRACT_SHADES', () => {
+            expect(isBarrierSupported('SomeThinMadeUp')).to.eql(false);
+        });
+        it('should return true when barrier is in CONTRACT_SHADES', () => {
+            expect(isBarrierSupported('CALL')).to.eql(true);
+        });
+    });
+
+    describe('barriersToString', () => {
```
nazaninreihani: @negarn These tests fail

src/javascript/app_2/Stores/Modules/SmartChart/Helpers/__tests__/barriers.js
```diff
@@ -0,0 +1,29 @@
...
+        it('should convert non-zero barriers which do not have +/- to string consisting of them without +/- while is_relative is false', () => {
+            expect(barriersToString(false, [10, 15])).to.deep.eql(['10','15']);
```
binary-hamedanchi: @nazaninreihani Seems `barriersToString()` doesn't expect to receive an array as the second argument, since it is a rest parameter:
https://github.com/binary-com/binary-static/blob/a0437f82d150fc44135659fcc54b9b68cc61f236/src/javascript/app_2/Stores/Modules/SmartChart/Helpers/barriers.js#L7

So, calling it like this: `barriersToString(false, 15, 10)` should work correctly.

Another sample call:
https://github.com/binary-com/binary-static/blob/a0437f82d150fc44135659fcc54b9b68cc61f236/src/javascript/app_2/Stores/Modules/SmartChart/chart_barrier_store.js#L67

src/javascript/app_2/Stores/Modules/Contract/Helpers/__tests__/details.js
```diff
@@ -0,0 +1,99 @@
...
+            });
+        });
+    });
+
+    describe('getDetailsExpiry', () => {
+        it('Should return an empty object if store is passed to it and property "is_ended" is false', () => {
+            expect(getDetailsExpiry({is_ended: false})).to.deep.include({});
+        });
+
+        it('Should return an empty object if store is passed to it and property "is_ended" is undefined', () => {
```
kellybinary: i think description should be: "..and invalid property `something_else` is passed"?

src/javascript/app_2/Stores/Modules/Trading/Helpers/__tests__/barrier.js
```diff
@@ -0,0 +1,131 @@
...
+            "contract_type":"CALL",
+            "exchange_name":"FOREX",
+            "expiry_type":"daily",
+            "market":"forex",
+            "max_contract_duration":"365d",
+            "min_contract_duration":"1d",
+            "sentiment":"up",
+            "start_type":"spot",
+            "submarket":"major_pairs",
+            "underlying_symbol":"frxAUDJPY"
```
kellybinary: would it be possible to define a contract object (without barriers) before running the tests, and only extends barriers when it is needed?
e.g: 
```
// global object
const contract_obj = {
    barrier_category: "euro_atm",
    contract_category: "callput",
    ...etc
};

it('...', () => {
    const contract = { 
           ...contract_obj, 
           barriers: 0, 
    }
});
```

src/javascript/app_2/Stores/Modules/SmartChart/Helpers/__tests__/barriers.js
```diff
@@ -0,0 +1,57 @@
...
+                main: {
+                    color: "green",
+                    draggable: false
+                },
+                somethingEmpty: {},
+            };
+            expect(Barriers.barriersObjectToArray(barriers)).to.deep.eql([{
+                color: "green",
+                draggable: false
+            }]);
```
kellybinary: same here, can we define the object `main` beforehand and call it where it's needed ? 😄

src/javascript/app_2/Stores/Modules/Contract/Helpers/__tests__/logic.js
```diff
@@ -0,0 +1,362 @@
...
+        it('should return true when contract does not have is_settleable, is_expired and status', () => {
+            const contract_info = {};
+            expect(Logic.isEnded(contract_info)).to.eql(false);
+        });
+    });
+
+    expect('getDisplayStatus', () => {
+        it('should return won if contract is ended and profit is more than zero', () => {
+            const contract_info = {
+                "status": "closed",
```
kellybinary: if not mistaken, valid contract statuses from the API are `open` `sold` `won` `lost`. maybe use `won` or `sold` here?

src/javascript/app_2/Stores/Modules/Contract/Helpers/__tests__/logic.js
```diff
@@ -0,0 +1,362 @@
...
+            const contract_info = {
+                "is_path_dependent": false,
+                "sell_spot_time": 123456,
+                "exit_tick_time": 987654321,
+            };
+            expect(Logic.getEndSpotTime(contract_info)).to.eql(987654321);
+        });
+        it('should return contract\'s exit tick time if is_path_dependent is undefined', () => {
+            const contract_info = {
+                "is_path_dependent": false,
```
kellybinary: since it's testing when `is_path_dependent` is `undefined`, could you remove this line?

src/javascript/app_2/Stores/Modules/Contract/Helpers/__tests__/logic.js
```diff
@@ -0,0 +1,362 @@
...
+    describe('getIndicativePrice', () => {
+        it('should return getFinalPrice if it has final price and contract is ended', () => {
+            const contract_info = {
+                "sell_price": 12345,
+                "status": "closed"
+            };
+            expect(Logic.getIndicativePrice(contract_info)).to.eql(12345);
+        });
+        it('should return null if it doesn\'t final price, bid_price and contract is not ended', () => {
+            const contract_info = {
```
kellybinary: typo: doesn't have :)

src/javascript/app_2/Stores/Modules/Contract/Helpers/__tests__/logic.js
```diff
@@ -0,0 +1,362 @@
...
+                "status": "open",
+            };
+            expect(Logic.getIndicativePrice(contract_info)).to.eql(null);
```
kellybinary: can we add another test case where sell_price is empty, bid_price not empty and status is open?

src/javascript/app_2/Stores/Modules/Contract/Helpers/__tests__/logic.js
```diff
@@ -0,0 +1,362 @@
...
+            const contract_info = {
+                status: 'open',
+                is_valid_to_sell: true,
+            };
+            expect(Logic.isValidToSell(contract_info)).to.eql(true);
+        });
+        it('should return false if contract is ended and is sold and contract is valid to sell', () => {
+            const contract_info = {
+                status: 'sold',
+                is_valid_to_sell: true,
```
kellybinary: `is_valid_to_sell` should be boolean value 0 or 1

src/javascript/app_2/Stores/Modules/Trading/Helpers/__tests__/proposal.js
```diff
@@ -0,0 +1,228 @@
...
+                has_error: true
+            });
+        });
+        it('should return profit and return calculated if proposal has no error', () => {
+            const store = {
+                currency: 'EUR'
+            };
+            const response = {
+                proposal: {
+                    proposal: 100,
```
kellybinary: `proposal: 1`?

src/javascript/app_2/Stores/Modules/Contract/Helpers/__tests__/logic.js
nazaninreihani: All done, thanks a lot

src/javascript/app_2/Stores/Modules/Contract/Helpers/__tests__/details.js
```diff
@@ -0,0 +1,99 @@
...
+                'Start Time'    : '2018-12-05 16:58:55 GMT',
+                'Entry Spot'    : '8,121.081',
+                'Purchase Price': <Money amount='10.00' currency='USD' />,
+            });
+        });
+    });
+
+    describe('getDetailsExpiry', () => {
+        it('Should return an empty object if store is passed to it and property "is_ended" is false', () => {
+            expect(getDetailsExpiry({is_ended: false})).to.deep.include({});
```
easteregg: If the desired return value SHOULD BE an empty object, we may use `to.be.equal` instead of non-strict `include`, Or better yet, we can check directly for empty objects using `expect({}).to.be.empty;`.

src/javascript/app_2/Stores/Modules/Contract/Helpers/__tests__/logic.js
kellybinary: thank you 😄

src/javascript/app_2/Stores/Modules/Trading/Helpers/__tests__/symbols.js
```diff
@@ -0,0 +1,108 @@
...
+                "symbol": "R_100",
+                "symbol_type": "stockindex"
+            }
+        ];
+
+        expect(pickDefaultSymbol(active_symbols)).to.eql('R_10');
+    });
+
+    it('It Returns the default symbol if active symbols has no value', () => {
+        expect(pickDefaultSymbol()).to.eql('');
```
kellybinary: to test empty strings or array, it's better to check with `to.be.empty` 🙏

PR #4706, opened 2018-11-29
aaron/optimisations

aaron-binary:


src/javascript/app/base/binary_pages.js
```diff
@@ -112,7 +112,7 @@ const pages_config = {
     trading                  : { module: TradePage,                  needs_currency: true },
     transferws               : { module: PaymentAgentTransfer,       is_authenticated: true, only_real: true },
     two_factor_authentication: { module: TwoFactorAuthentication,    is_authenticated: true },
-    virtualws                : { module: VirtualAccOpening,          not_authenticated: true },
+    virtualws                : { module: VirtualAccOpening,          only_non_authenticated: true },
```
negarn: why are we not using `not_authenticated` as before? this way it handles it for other pages too, such as `new-account`, which is what we wanted

src/javascript/app/base/footer.js
```diff
@@ -28,16 +29,26 @@ const Footer = (() => {
...
     
     const displayNotification = (message) => {
-        const $status_notification = $('#status_notification');
-        const $status_message_text = $('#status_notification_text');
-        const $close_icon = $('#status_notification_close');
-
-        $status_notification.css('display', 'flex');
-        $status_message_text.html(message);
+        const notification_storage = LocalStore.getObject('status_notification');
+        const time_difference = (window.time.unix() - notification_storage.close_time || 0);
```
negarn: maybe worth waiting for `time` response, might get called before we have the response of time

negarn: could you put `notification_storage.close_time || 0` in a bracket? also might be worth turning `notification_storage.close_time` into an integer

src/javascript/app/base/binary_loader.js
```diff
@@ -101,7 +102,7 @@ const BinaryLoader = (() => {
                     });
             }
         } else if (config.not_authenticated && Client.isLoggedIn()) {
-            BinaryPjax.load(Client.defaultRedirectUrl(), true);
+            handleNotAuthenticated(error_messages.not_authenticated());
```
negarn: if it's a global object, do we need to pass the error here? why not call it inside `handleNotAuthenticated` itself?

aaron-binary: We can keep the error messages grouped together so it'll be easier to update them in the future.

src/javascript/app/pages/cashier/cashier.js
```diff
@@ -12,7 +12,9 @@ const Cashier = (() => {
 
     const showContent = () => {
         Client.activateByClientType();
-        Header.upgradeMessageVisibility(); // To handle the upgrade buttons visibility
+        if (Client.getUpgradeInfo().type !== 'financial') {
```
negarn: I think we should act similarly to cashier page and remove all upgrade buttons from this page.

negarn: I mean, from payment methods page.

PR #4707, opened 2018-11-29
Nazanin/add_crypto_names_virtual

nazaninreihani:


src/javascript/app/common/currency.js
```diff
@@ -16,6 +16,17 @@ const getCurrencyList = (currencies) => {
     return $currencies.append($fiat_currencies.children().length ? $fiat_currencies : '').append($cryptocurrencies.children().length ? $cryptocurrencies : '');
 };
 
+const getCurrencyNameList = (currencies) => {
+    const currencies_name_list = [];
+    currencies.forEach((currency) => {
+        const is_crypto_currency = CurrencyBase.isCryptocurrency(currency);
+        const currency_name      = is_crypto_currency ? `${CurrencyBase.getCurrencyName(currency)} (${currency})` : currency;
+        currencies_name_list.push(currency_name);
```
negarn: can we combine this somehow with `getCurrencyList`? seems they have some parts in common

src/javascript/app/common/currency.js
```diff
@@ -1,21 +1,35 @@
 const CurrencyBase = require('../../_common/base/currency_base');
 const localize     = require('../../_common/localize').localize;
 
+const is_crypto_currency = (currency) => CurrencyBase.isCryptocurrency(currency);
```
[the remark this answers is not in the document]
nazaninreihani: I'll remove it

src/javascript/app/pages/user/accounts.js
```diff
@@ -1,15 +1,17 @@
...
-const urlFor             = require('../../../_common/url').urlFor;
+const moment              = require('moment');
+const setIsForNewAccount  = require('./account/settings/personal_details').setIsForNewAccount;
+const getCurrencies       = require('./get_currency').getCurrencies;
+const BinaryPjax          = require('../../base/binary_pjax');
+const Client              = require('../../base/client');
+const BinarySocket        = require('../../base/socket');
+const getCurrencyFullName = require('../../common/currency').getCurrencyFullName;
+const getCurrencyNameList = require('../../common/currency').getCurrencyNameList;
+const getCurrencyList     = require('../../common/currency').getCurrencyList;
```
negarn: wondering if it's better to import the whole currency module since we are using three of its functions

src/sass/static/pages.scss
```diff
@@ -694,7 +694,7 @@ body #not_authenticated_financial ul.checked > li {
 
 .accounts-content {
     .table-container {
-        overflow-x: hidden;
+        overflow-x: visible;
```
negarn: so do we need this style at all?

nazaninreihani: @negarn Due to having the responsive table we no longer need horizontal scroll in these tables, and when I set it to hidden it causes the currency dropdown to be hidden, but visible works fine

PR #4709, opened 2018-11-29
[new-app] kelly/outdated_browser_fix

kellybinary:
-  add link to update browser 
- prevent mounting Trade components when there's error


src/javascript/app_2/Modules/Trading/Containers/trade.jsx
```diff
@@ -23,77 +23,83 @@ class Trade extends React.Component {
...
-                        timeout={400}
-                        classNames='contract-wrapper'
-                        unmountOnExit
-                    >
-                        <div className='contract-wrapper'>
-                            <ContractDetails
-                                contract_id={contract_id}
-                                onClickNewTrade={this.props.onClickNewTrade}
+            <React.Fragment>
+                { this.props.is_trade_component_mounted &&
```
morteza-binary: @kellybinary it causes the `Trade.render` invoked two times, one time when it is mounting and one more time when it is mounted completely.

IMHO, it's better to add a route for the error page.

src/javascript/app_2/Services/outdated_browser.js
```diff
@@ -28,7 +29,14 @@ const OutdatedBrowser = (() => {
 
     const updateStore = action('showError', () => {
         if (common_store) {
-            common_store.showError(localize('Your web browser is out of date and may affect your trading experience. Proceed at your own risk.'));
+            common_store.showError(
```
morteza-binary: @kellybinary it's better to not use react and JSX tags in '.js' files.

src/javascript/app_2/Stores/base_store.js
```diff
@@ -270,7 +270,7 @@ export default class BaseStore {
 
     @action.bound
     disposeSwitchAccount() {
-        this.switch_account_disposer();
+        this.onSwitchAccount();
```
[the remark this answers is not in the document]
kellybinary: `this.switch_account_disposer` is not a function, it was throwing an error 🤔

src/javascript/app_2/App/Components/Elements/Errors/error_component.jsx
```diff
@@ -2,17 +2,32 @@ import PropTypes    from 'prop-types';
...
+            <p>{msg || localize('Sorry, an error occured while processing your request.')}</p>
+        </div>
+    );
+};
 
 ErrorComponent.propTypes = {
-    message: PropTypes.node,
-    type   : PropTypes.string,
+    message: PropTypes.oneOfType([
+        PropTypes.object,
```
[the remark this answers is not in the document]
kellybinary: okay, sure

src/javascript/app_2/Stores/base_store.js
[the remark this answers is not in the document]
easteregg: Check this out: 

https://github.com/binary-com/binary-static/blob/2f97972da7a992dc1e8bff191290dade6cab3026/src/javascript/app_2/Stores/base_store.js#L260-L264

It is just a fail-safe to let the developer know that there is no listener is set for switching accounts. That is not an actual error.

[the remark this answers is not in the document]
kellybinary: ahh I see.. thanks, that makes sense 👍

src/javascript/app_2/App/Components/Elements/Errors/error_component.jsx
```diff
@@ -2,17 +2,32 @@ import PropTypes    from 'prop-types';
...
-);
+const ErrorComponent = ({ type, message }) => {
+    let msg = '';
+    if (typeof message === 'object') {
+        msg = <Localize
+            str={message.str}
+            replacers={message.replacers}
+        />;
+    } else {
+        msg = message;
```
easteregg: Question, `msg` in the else block is not `localized`. Are we counting on the user to send the `localized` string?

src/javascript/app_2/Stores/base_store.js
```diff
@@ -270,7 +270,7 @@ export default class BaseStore {
 
     @action.bound
     disposeSwitchAccount() {
-        this.switch_account_disposer();
+        this.switch_account_disposer = null;
```
easteregg: `disposeSwitchAccount` is only called when we want to intercept `client switching mechanism` and stop client switch. So only after we call it (to stop client switch), we set it to null. I believe you can disregard this file entirely. if you have any question regarding this piece, drop me a slack message.

morteza-binary: @kellybinary 
Regarding the [Mobx document](https://mobx.js.org/refguide/when.html), `when` returns a disposer function to cancel the autorunner permanently so the `this.switch_account_disposer` should be a function. 
If you faced a case that `this.switch_account_disposer` was not a function, probably, it was a race condition and we should fix it.
Would you plz let @easteregg  know how to produce it?

@easteregg regarding the name convention, `switch_account_disposer` should be in camelcase instead of snake case.

kellybinary:
https://github.com/binary-com/binary-static/pull/4828

src/javascript/app_2/App/Components/Elements/Errors/error_component.jsx
kellybinary: those messages are `localized` before it is being passed to this component

PR #4577, opened 2018-11-05
add svg inliner

easteregg:

PR #4714, opened 2018-11-30
zaki/add_404_page

zakibinary:
Changelog
- Add 404 page assets and templates for `new-ui`.
- Route all undefined paths to 404 page.


src/javascript/app_2/Modules/Page404/Components/Page404.jsx
```diff
@@ -0,0 +1,27 @@
...
+import ErrorBox       from '../../../App/Components/Elements/ErrorBox';
+
+const Page404 = () => (
+    <div className='page-error-container'>
+        <ErrorBox
+            header={localize('Page not found')}
+            icon={<Icon404 />}
+            message={localize('Sorry, we couldn\'t find the page you are looking for.')}
+        >
+            <Link to={routes.trade}>
```
negarn: didn't we decide to suggest the closest page to them? e.g. if there's `statement` in the path, suggest `statement` instead of trade, and fallback to trade by default? @morteza-binary @ashkanx

zakibinary: It was suggested, but I think we might need an additional design for this. 

For me the Trade button should always be there in `404` to encourage clients back to Trade, but we can add a suggestion message below the button, something along the lines of `Did you mean: so and so`, similar to how Search Engines do it.

ashkanx: let's add the `trade` only. looks like New new-design will bring statement and portfolio into the trade page

src/javascript/app_2/Modules/Page404/Components/Icon404.jsx
```diff
@@ -0,0 +1,24 @@
+import classNames from 'classnames';
```
morteza-binary: @zakibinary 
According to PR #4577, we can import SVG files directly and use like other react component so we just need to save SVG files in `images` and import them like [the logo](https://github.com/binary-com/binary-static/pull/4577/files#diff-692734e21cd9894d9bcff2010d0d73a2R3)

P.S it's not a good idea to import all SVG files in JSX files because it makes the bundle file massive.

src/javascript/app_2/Modules/Page404/Components/Page404.jsx
morteza-binary: @zakibinary 
Since the `<Link>` tag is transformed to `<a>` and according to the [HTML5 spec document form W3C](http://w3c.github.io/html/textlevel-semantics.html#the-a-element), a `<button>` tag is not allowed to nest inside an `<a>` tag in HTML5. It's better to customize the `<Link>` component and add it the appropriates CSS classes.

Also, we have to fix it in [empty_statement_message](https://github.com/binary-com/binary-static/blob/master/src/javascript/app_2/Modules/Statement/Components/empty_statement_message.jsx#L23)

morteza-binary:
@zakibinary Would you pls resolve the conflict?

src/javascript/app_2/Modules/Page404/Components/Icon404.jsx
zakibinary: @morteza-binary , sure will switch to adding SVG via import instead.

zakibinary: @morteza-binary , after consideration of time and effort, I think for now we can keep the JSX implementation for the icon. 

Reason for this is that we apply a couple of global classes for modifying the `fill` and `stroke` to the paths in the SVG icon to suit `dark` and `light` theme. 

When using SVG Import, the `svgo` utility will remove and trim all of the classes inside the svg file which will remove the theme classes appended in the SVG icon. 

I have already optimized the SVG before appending it to the JSX file so in this case `svgo` isn't necessary.
 
I agree that it's not a good idea to import all SVG files into JSX, but for use-cases like this, where the icons are simple vector shapes and have maximum of two colors, I think we can keep the current implementation for now until we find a better way to manage/refactor theming.

morteza-binary: @morteza-binary we can postpone it after receiving the new design because we have to refactor the SCSS files too.

src/javascript/app_2/Modules/Page404/Components/Page404.jsx
```diff
@@ -0,0 +1,27 @@
+import React          from 'react';
+import { routes }     from 'Constants/index';
+import { localize }   from '_common/localize';
+import { Icon404 }    from './Icon404.jsx';
+import { ButtonLink } from '../../../App/Components/Routes';
```
morteza-binary: @zakibinary 
using alias path makes the maintenance more easier
```suggestion
import { ButtonLink } from 'App/Components/Routes';
```

src/javascript/app_2/Modules/Page404/Components/Page404.jsx
```diff
@@ -0,0 +1,27 @@
...
+import ErrorBox       from '../../../App/Components/Elements/ErrorBox';
```
morteza-binary:
```suggestion
import ErrorBox       from 'App/Components/Elements/ErrorBox';
```

src/javascript/app_2/Modules/Statement/Components/empty_statement_message.jsx
```diff
@@ -1,10 +1,9 @@
 import PropTypes         from 'prop-types';
 import React             from 'react';
-import { Link }          from 'react-router-dom';
 import { localize }      from '_common/localize';
-import Button            from 'App/Components/Form/button.jsx';
 import { IconStatement } from 'Assets/Header/NavBar';
 import { routes }        from 'Constants/index';
+import { ButtonLink } from '../../../App/Components/Routes';
```
morteza-binary:
```suggestion
 import { ButtonLink } from 'App/Components/Routes';
```

src/javascript/app_2/Modules/Page404/Components/Page404.jsx
zakibinary: Thanks!

PR #4716, opened 2018-11-30
ruskakimov/track_only_successful_email_submits

ruskakimov:


src/javascript/static/pages/home.js
```diff
@@ -60,9 +62,21 @@ const Home = (() => {
...
+            });
         } else {
             $('.signup-box div').replaceWith($('<p/>', { text: localize('Thank you for signing up! Please check your email to complete the registration process.'), class: 'gr-10 gr-centered center-text' }));
             $('#social-signup').setVisibility(0);
+            GTM.pushDataLayer({
+                event                   : 'email_submit',
+                email_submit_input      : response.echo_req.verify_email,
+                email_submit_days_passed: moment().diff(moment(localStorage.getItem('date_first_contact')), 'days'),
+                email_submit_source     : 'binary.com',
+            });
```
negarn: why not just declare the common variable and push the email submit source to it later then send it once?

ruskakimov: if we never get an error then sure 👍

negarn: it can be in a bigger else wrapper

src/javascript/static/pages/home.js
```diff
@@ -55,15 +57,22 @@ const Home = (() => {
...
+        if (isBinaryApp()) {
             BinaryPjax.load(urlFor('new_account/virtualws'));
         } else {
             $('.signup-box div').replaceWith($('<p/>', { text: localize('Thank you for signing up! Please check your email to complete the registration process.'), class: 'gr-10 gr-centered center-text' }));
             $('#social-signup').setVisibility(0);
         }
+        GTM.pushDataLayer({
+            event                   : 'email_submit',
+            email_submit_input      : response.echo_req.verify_email,
+            email_submit_days_passed: moment().diff(moment(localStorage.getItem('date_first_contact')), 'days'),
```
negarn: do we need to convert to utc?

ruskakimov: good point, `date_first_contact` is in UTC

PR #4717, opened 2018-11-30
amam/change_buttons_sentence_case

cakasuma:


src/sass/_common/base/mixin.scss
```diff
@@ -61,7 +61,7 @@
     text-align: center;
     background: $COLOR_GREEN;
     border-radius: $BORDER_RADIUS;
-    text-transform: capitalize;
+    text-transform: initial;
```
negarn: same as Kelly's comment in the other PR, can we simply remove this line?

PR #4719, opened 2018-12-02
easteregg/add_remarks

easteregg:


src/sass/app/paymentagent_transfer.scss
```diff
@@ -0,0 +1,4 @@
+.paymentagent_description {
+    height: 4em;
+    width: 206px;
+}
```
negarn: if it's a small amount of css, we generally put it in `pages.scss` for static and `websocket_pages.scss` for websocket pages

src/sass/app/statement.scss
```diff
@@ -0,0 +1,9 @@
+#statement-table {
```
negarn: we are using this identifier in `table.scss`, we could move this to the same place

src/templates/_common/components/forms.jsx
```diff
@@ -44,14 +44,38 @@ export const FormRow = ({
...
+                                    name={id}
+                                    id={option_id}
+                                    value={value}
+                                    defaultChecked={value === default_option}
+                                />
+                                <label htmlFor={option_id}>
+                                    <span
+                                        data-balloon={data_balloon_text}
+                                        data-balloon-length='medium'
+                                    >{label_text}
```
negarn: as far as I know we don't have a convention for putting the closing tag and text of the element on the same line.

src/templates/_common/components/forms.jsx
```diff
@@ -44,14 +44,38 @@ export const FormRow = ({
...
+                                    type='radio'
+                                    name={id}
+                                    id={option_id}
+                                    value={value}
+                                    defaultChecked={value === default_option}
+                                />
+                                <label htmlFor={option_id}>
+                                    <span
+                                        data-balloon={data_balloon_text}
+                                        data-balloon-length='medium'
```
negarn: generally two attributes should be possible to put on the same line, not sure why line breaks are needed here.

src/templates/_common/components/forms.jsx
```diff
@@ -44,14 +44,38 @@ export const FormRow = ({
...
                 </div>
             );
         }
+        if (type === 'textarea') {
+            return (
+                <React.Fragment>
+                    <textarea
+                        className={className}
+                        id={id}
+                        {...attributes}
```
negarn: same here, as far as I remember our convention is to break when there are four or more attributes on the same line, here there are only three.

src/sass/static/pages.scss
```diff
@@ -1460,6 +1460,11 @@ body #not_authenticated_financial ul.checked > li {
     }
 }
 
+.paymentagent_description {
+    height: 4em;
```
OskarAhl: What is the reason for having the height in `em`?

src/sass/app/websocket_pages.scss
```diff
@@ -116,6 +116,12 @@
     }
 }
 
+.statementws-content {
+    .container {
+        max-width: 1200px !important;
+    }
+}
+
 #statement-container {
```
negarn: not sure if we should move this inside line 123, perhaps the diff becomes too large

PR #4720, opened 2018-12-02
[new-app] easteregg/check_duplicate

easteregg:
- **Update**: This description needs update
- **Update 2**: Updated the description to match command signatures. 
- **Update 3**: Added `check_directory_naming_convention` command. 

This PR adds 4 commands to `grunt`:

```
grunt check_duplicated_images [--section=all|app_2]

This command checks for duplicate images inside `src/images/app_2` by default. 

OPTIONS:
--section [default=all]:  

if section app_2 is provided, this command checks `src/images/app_2` directory.
```
```
grunt check_image_naming_convention [--auto-fix] [--dry-run] [--section=all|app_2]

OPTIONS: 
--section [default=all]: 
This option accepts 'all' and 'app_2' values. 

By default, this command only checks 'src/images' directory for 
image file names that are not according to file names convention. 

--auto-fix [default=false]: 
If option provided, the command will attempt to rename the files. 
Also, it will try to update the references it has found in the source directories: 
  - 'src/javascipt/[app_2]'
  - 'src/sass/[app_2]'

NOTE: This command DOES NOT commit the new files. that should 
be done manually. 

--dry-run 
If provided with '--auto-fix' option, instead of renaming the files and 
and editing source files, it just prints a verbose output of the rename action. 
It is SAFER and RECOMMENDED to run with '--dry-run' option first. 
```

```
grunt check_orphaned_images [--section=all|app_2] [--delete=false] [--dry-run=false]

This commands checks for images' reference inside javascript and sass directory, 
and prepares a list for a user to consider removing. 

OPTIONS: 
--section [default=all]
If option provided, will try to check 'src/images' instead of 'src/images/app_2'. 
Also will look for 'src/sass' and 'src/javascript' directories. If 'app_2' is selected,
then only the following directories will be checked: 
    - src/images/app_2/
  Inside: 
    - src/javascript/app_2/
    - src/sass/app_2/

--delete [default=false]
If this option is provided, this command will delete the files marked as orphaned. 
```

```
grunt check_missing_images [--section=all|app_2] 
this command attempts to find the images referenced inside javascript or sass folder. If it cannot find the file mentioned in the source files, it'll list them as a warning and fails. 

OPTIONS: 
--section [default=all]
If option is provided, it will try to check everywhere. If nothing is provided, it will only look for the 'app_2' references. 
```

There is also a new grunt alias defined that runs all the above commands in `app_2` mode. 

Command `check_image_naming_convention` and `check_directory_naming_convention` has the same signature and behavior. But I implemented another command just to be on the safe side.


build/aliases.yaml
```diff
@@ -54,3 +54,6 @@ release:
   - 'shell:make_cname'
   - 'gh-pages:main'
   - 'shell:trigger_tests'
+
+check_duplicates:
+  - 'duplicate_check'
```
binary-hamedanchi: Since it's only about images, better to mention that in the name. e.g. `duplicated_images`.

easteregg:
Updated PR description with the new command. (check_orphaned_images)

build/checkImagesNamingConvention.js
```diff
@@ -0,0 +1,61 @@
+const path      = require('path');
+const common    = require('../scripts/common');
+const {
+          prompt, flatten, fetchFiles, renameFiles, validateFilename, convertToSnakeCase,
```
negarn: preferably each variable should be on a new line

build/checkImagesNamingConvention.js
```diff
@@ -0,0 +1,61 @@
...
+    grunt.initConfig({
+        checkImagesNamingConvention: {},
+    });
+
+    grunt.registerTask('image_naming_convention', 'Check naming convention for images', () => {
+        const default_path  = grunt.option('all') ? '' : 'app_2';
+        const where_to_look = path.resolve(common.root_path, `src/images/${default_path}/`);
+
+        // Actual test is running from here. We can check for other directories as well, but it is highly unlikely that
+        // someone put a duplicate in other folders. This grunt task will accept a --all option. --all option will try
```
negarn: perhaps this comments need to be updated?

build/checkMissingImages.js
```diff
@@ -0,0 +1,66 @@
...
+    grunt.registerTask('check_missing_images', 'check missing images', () => {
+        const default_path                = grunt.option('all') ? '' : 'app_2';
+        const images_directory            = path.resolve(common.root_path, `src/images/${default_path}/`);
+        const javascript_source_directory = [
+            path.resolve(common.root_path, `src/javascript/${default_path}`),
+            path.resolve(common.root_path, `src/sass/${default_path}`),
+        ];
+
+        // Actual test is running from here. We can check for other directories as well, but it is highly unlikely that
+        // someone put a duplicate in other folders. This grunt task will accept a --all option. --all option will try
```
negarn: also this

build/checkOrphanedImages.js
```diff
@@ -0,0 +1,89 @@
...
+                        return `${matches[1]}/${item.filename}`;
+                    }
+                    return false;
+                })
+                .filter(item => item)
+                .filter(item => !ignored_language_directories.some(ignored => item.startsWith(ignored)));
+        };
+
+        // Actual test is running from here. We can check for other directories as well, but it is highly unlikely that
+        // someone put a duplicate in other folders. This grunt task will accept a --all option. --all option will try
```
negarn: and here

build/checkMissingImages.js
```diff
@@ -0,0 +1,61 @@
...
+            })
+            .filter(item => !!item)
+            .map(item => item.map(line => line.match(/(images)([\/\w\-_d.]+)/g)));
+
+        // Check for existence of the images relative to images directory
+        const not_found = [
+            ...new Set(flatten(references)
+                .filter(item => !!item)
+                .filter(item => {
+                    return !fs.existsSync(path.resolve(common.root_path, `src/${item}`));
```
binary-hamedanchi: Please note that `app_2` images will be deployed to `/app/images/...` https://github.com/binary-com/binary-static/blob/44b227ac570a1d24db7ce567a7b6bb2b7d6194f8/build/copy.js#L43

e.g. https://github.com/binary-com/binary-static/blob/44b227ac570a1d24db7ce567a7b6bb2b7d6194f8/src/sass/app_2/_common/form/date_picker.scss#L278 is actually referring to an image located here: `src/images/app_2/common/close_circle.svg`

build/checkMissingImages.js
```diff
@@ -0,0 +1,61 @@
...
+                })),
+        ].map(item => {
+            grunt.log.ok(`Could not find image: ${item}`);
```
binary-hamedanchi: Would be useful to also report the location(s) which the image is being referenced. (filename & line number)

build/checkOrphanedImages.js
```diff
@@ -0,0 +1,91 @@
...
+                // Search for the occurrence of images, if found, remove the entry from that array.
+                const file                   = fs.readFileSync(path.resolve(item.path, item.filename), {
+                    encoding: 'utf8',
+                });
+                images_and_closest_directory = images_and_closest_directory.filter(dir => file.indexOf(dir) === -1);
+            });
+        if (images_and_closest_directory.length > 0) {
+            grunt.log.ok(`Couldn't find references for ${images_and_closest_directory.length} files.`);
+            grunt.log.ok(`Consider finding usage for the following files: `);
+            images_and_closest_directory.forEach(orphan => grunt.log.writeln(`File: ${orphan}`));
```
binary-hamedanchi: - `closest_directory` won't be helpful for the developer who gets the report. Please report relative path to the project's root `./src/images/...`
- Just a bullet like would be better than keep repeating `File: ` at every line.

build/utils/cli.js
```diff
@@ -0,0 +1,84 @@
...
+const convertToSnakeCase = (item) => {
+    let correct_name = snakeCase(item.split('.')[0]);
+    correct_name += `.${item.split('.')[1]}`;
+    return correct_name;
+};
+
+const renameFiles = (bad_file_names, grunt) => {
+    Object.keys(bad_file_names).forEach(item => {
+        const converted = convertToSnakeCase(item, grunt);
+        if (!grunt.option('dry-run')) {
```
binary-hamedanchi: Not a good practice to check for external options inside a helper function. Better to pass a boolean instead. Otherwise, this function doesn't belong to this file.

build/checkOrphanedImages.js
```diff
@@ -0,0 +1,91 @@
+const path     = require('path');
+const fs       = require('fs');
+const common   = require('../scripts/common');
+const {
+  prompt,
+  flatten,
+  fetchFiles } = require('./utils/cli');
+
+const ignored_files            = [];
```
morteza-binary: @easteregg Also here it should be better to use `.gitignore` file as default value.

build/checkDuplicatedImages.js
```diff
@@ -0,0 +1,63 @@
...
+        const duplicated = Object.keys(hashes).filter(hash => {
+            return Object.values(hashes[hash]).length > 1;
+        });
+        if (!duplicated.length) {
+            grunt.log.subhead(`Found no duplicate among ${flattened.length} files.`.green);
+            return true;
+        } else {
+            grunt.log.subhead(`Duplicates are found. refer below: `);
+            duplicated.forEach((item, index) => {
+                const errorMessage = `Hash: ${index + 1}: ${item}`.bold;
```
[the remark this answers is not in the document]
easteregg: :facepalm: will do.

build/checkOrphanedImages.js
```diff
@@ -0,0 +1,91 @@
...
+];
+
+module.exports = function(grunt) {
+    prompt('Checking for images with bad naming...');
+    grunt.initConfig({
+        checkOrphanedImages: {},
+    });
+
+    grunt.registerTask('check_orphaned_images', 'Check orphaned images', () => {
+        const default_path                = grunt.option('all') ? '' : 'app_2';
```
morteza-binary: @easteregg it would be use full if there were a switch (e.g. --delete) to remove orphaned images too.

build/checkImagesNamingConvention.js
```diff
@@ -0,0 +1,64 @@
...
+            Object.keys(bad_file_names).forEach(item => {
+                const error_message = `file ${bad_file_names[item]}/${item} should be changed to 
+                ${convertToSnakeCase(item, grunt)}`.bold;
+                grunt.log.ok(error_message);
+            });
+
+            grunt.log.writeln('\n'.repeat(2));
+            if (grunt.option('auto-fix')) {
+                grunt.log.ok('Attempting to rename the files...');
+                renameFiles(bad_file_names, grunt);
```
morteza-binary: @easteregg It would be better to rename the images' names in SCSS,  JS(X), ... files which use or print the lists of their names.

build/checkOrphanedImages.js
```diff
@@ -0,0 +1,91 @@
...
+  prompt,
+  flatten,
+  fetchFiles } = require('./utils/cli');
+
+const ignored_files            = [];
+const ignored_image_extensions = [
+    'DS_Store', 'zip',
+];
+
+const ignored_language_directories = [
```
morteza-binary: @easteregg I would be better to import and use https://github.com/binary-com/binary-static/blob/44b227ac570a1d24db7ce567a7b6bb2b7d6194f8/scripts/common.js#L36

build/checkImagesNamingConvention.js
```diff
@@ -0,0 +1,64 @@
...
+const ignored_files = [];
+
+module.exports = function(grunt) {
+    prompt('Checking for images with bad naming...');
+    grunt.initConfig({
+        checkImagesNamingConvention: {},
+    });
+
+    grunt.registerTask('image_naming_convention', 'Check naming convention for images', () => {
+        const default_path  = grunt.option('all') ? '' : 'app_2';
```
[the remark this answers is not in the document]
easteregg: I partially implemented this. since selecting app and all does the same in the code. but the switch has changed to 'section'.

build/image_checks/checkImagesNamingConvention.js
```diff
@@ -0,0 +1,110 @@
...
+    getDefaultPath,
+    getImageReferences,
+    gitIgnoreToArray,
+    flatten,
+    fetchFiles,
+    renameFiles,
+    validateFilename,
+    convertToSnakeCase }      = require('../utils/cli');
+
+const ignored_files = gitIgnoreToArray();
```
morteza-binary: @easteregg 
It seems that the `ignore-files` doesn't affect the result.
image

build/image_checks/checkOrphanedImages.js
```diff
@@ -0,0 +1,96 @@
...
+                    fs.unlinkSync(orphan.relative_path);
+                }
+            });
+        } else if (found.length > 0) {
+            grunt.log.error(`Couldn't find references for ${found.length} files.`);
+            grunt.log.writeln('>> Consider find a usage for the following files: ');
+            found.forEach(orphan => grunt.log.writeln(`${orphan.relative_path}`));
+            grunt.log.writeln('='.repeat(80));
+            grunt.log.error(
+                'Some of the files above could be used dynamically.\nIf you are going to delete some, proceed with caution.\n' +
```
morteza-binary: @easteregg It seems that the result of the task is not reliable, e.g all the images in [this directory](https://github.com/binary-com/binary-static/tree/master/src/images/pages/trade-explanation) are marked as orphaned (at least 40% of the result) but all of them are using.

morteza-binary:
the task `check_image_naming_convention` just checks and fixes the files' names, how about the directories' names in `./src/images/`?  should the task handle directories' name convention too?

build/image_checks/checkDuplicatedImages.js
```diff
@@ -0,0 +1,60 @@
...
+                !ignored_files.some(ignored => path.resolve(common.root_path, ignored) === item.filename) &&
+                !ignored_extensions.some(ignored => item.filename.split('.')[1] === ignored));
+
+        const hashes    = {};
+
+        flattened.forEach(item => {
+            hashes[item.hash] = [...(hashes[item.hash] || []), { ...item }];
+        });
+        // loop over values of the hashes, report those with more than 1 value
+        const duplicated = Object.keys(hashes).filter(hash => Object.values(hashes[hash]).length > 1);
```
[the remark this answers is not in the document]
easteregg: They are the same. bad naming.

build/image_checks/checkOrphanedImages.js
easteregg: I ignored language images in the past commits, since we are using them dynamically, we cannot detect detect their uses. I try to push a workaround/not-a-fix for it.

build/image_checks/checkImagesNamingConvention.js
easteregg: I think I didn't have any .DS_Store locally. I'll push a fix.

build/image_checks/checkDuplicatedImages.js
```diff
@@ -0,0 +1,62 @@
...
+        const default_path = getDefaultPath(grunt.option('section'));
+        const where_to_look = path.resolve(common.root_path, `src/images/${default_path}/`);
+        grunt.log.writeln(`Checks for duplicate images based on their content.\nSearching: src/images/${default_path}`);
+
+        // Actual test is running from here. We can check for other directories as well, but it is highly unlikely that
+        // someone put a duplicate in other folders. This grunt task will accept a --all option. --all option will try
+        // to check every single file inside images directory. if no option is provided, only the images inside
+        // app_2 will be processed.
+        const flattened = flatten(fetchFiles(where_to_look))
+            .filter(item => !!item &&
```
OskarAhl: Would it be possible to add filter utils and structure the filter to something like this?
Pros: can reuse the filtering logic here and in `checkImagesNamingConvention.js` and `checkOrphanedImages.js`

```javascript
.filter(Boolean) // same as filter(item => !!item)
.filter(isNotIgnoredFileCommonRoot)
.filter(isNotIgnoredExtension)

// filter utils:
const isNotIgnoredFileCommonRoot = (item) => !ignored_files.some(ignored => path.resolve(common.root_path, ignored) === item.filename);

const isNotIgnoredFile = (item) => !ignored_files.some(ignored => ignored === item.filename);

const isNotIgnoredExtension = (item) => {
  return !ignored_extensions.some(ignored => {
        const matchingGroups = item.filename.match(/\.[\w\d]{1,5}$/i);
        return matchingGroups && matchingGroups[0].substring(1) === ignored;
  }));
}
```

easteregg: It was like this at the beginning, but decided to change the structure to make it more fast and run the filters only once.

morteza-binary:
@easteregg Would you pls resolve the conflict?

build/utils/cli.js
```diff
@@ -39,6 +39,23 @@ const fetchFilesFromPath = (location) => {
         });
 };
 
+const fetchDirectoriesFromPath = (location) => {
+    const directory = fs.readdirSync(location);
+    const directories = [];
+    directory.map(file => {
+        if (fs.statSync(path.join(location, file)).isDirectory()) {
+            fetchDirectoriesFromPath(path.join(location, file));
```
morteza-binary: @easteregg What is the purpose of calling the function recursively without using its result?

build/utils/cli.js
```diff
@@ -121,10 +146,11 @@ const getImageReferences = (source_directories, ignored_files = []) => flatten(f
             if (!line) {
                 return;
             }
-            const matched = line.match(/Images\/app_2\/([^';]*)|<img([^>]*)>|background: url([^-color][^;]*)/g);
+            const matched = line.match(/Images\/app_2\/([^';]*)|<img([^>]*)>|background[-image]+: url([^-color][^;]*)/g);
```
morteza-binary: @easteregg Would you plz consider the regex again? e.g should `[-image]+` be `(-image)?` or it'd be better to add more groups to make clear the `or`s

build/utils/cli.js
```diff
@@ -121,10 +146,11 @@ const getImageReferences = (source_directories, ignored_files = []) => flatten(f
...
             if (matched && matched.length > 0) {
                 matches.push({
-                    path   : item.relative_path, line   : index + 1, matched: matched[0],
+                    path   : item.relative_path,
+                    line   : index + 1, matched: matched[0],
```
morteza-binary: should `matched: matched[0],` be in a new line?

build/image_checks/checkDirectoryNamingConvention.js
```diff
@@ -0,0 +1,114 @@
...
+
+            item.map(found => {
+                if (found.reference != null) {
+                    let filename  = found.reference.split('/');
+                    filename.pop();
+                    filename = filename.join('/');
+                    const converted = convertToSnakeCase(filename);
+                    if (filename !== converted) {
+                        if (!is_dry_run) {
+                            // we should update
```
morteza-binary: what should be updated?

package.json
```diff
@@ -86,6 +86,7 @@
     "jsdom-global": "2.1.1",
     "load-grunt-config": "0.19.2",
     "load-grunt-tasks": "3.5.2",
+    "md5": "^2.2.1",
```
morteza-binary: would you plz use the exact version?

build/image_checks/checkOrphanedImages.js
```diff
@@ -60,6 +61,14 @@ module.exports = function(grunt) {
                     return false;
                 }
 
+                // Since language different images may be used dynamically, we remove that check from the list.
```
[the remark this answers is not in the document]
easteregg: What is your suggestion? just a warning or you have some other thing in mind?

build/image_checks/checkDirectoryNamingConvention.js
```diff
@@ -0,0 +1,114 @@
...
+            Object.assign(bad_file_names, validateFilename(item.filename) ? null : {
+                [item.filename]: {
+                    relative_path: item.relative_path,
+                    path         : item.path,
+                },
+            });
+        });
+
+        Object.keys(bad_file_names).forEach((key) => (bad_file_names[key] == null) && delete bad_file_names[key]);
+        if (!bad_file_names || (bad_file_names && Object.keys(bad_file_names).length === 0)) {
```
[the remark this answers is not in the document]
easteregg: Importing the library throws: 

```ReferenceError: Element is not defined
>>   at Object.<anonymous> (/Users/easteregg/www/new-app/src/javascript/_common/lib/polyfills/element.matches.js:1:63)
>>   at Module._compile (internal/modules/cjs/loader.js:721:30)
>>   at Object.Module._extensions..js (internal/modules/cjs/loader.js:732:10)```

is that one meant to be run in node env? reverting the changes.

build/utils/cli.js
morteza-binary: `(-image)?` means zero or one of `-image` so it doesn't miss `background: url(`.

morteza-binary: the `?` is belong to the suggestion and not for the question :-P

build/image_checks/checkOrphanedImages.js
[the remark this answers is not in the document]
morteza-binary: at least we should show a warning to inform that the task doesn't check those particular directories. the best practice is to find a way to check the images which are used dynamically. Maybe we can handle it by parsing the js(x) files by Babel or some other libraries like something @binary-hamedanchi has done for extracting texts.

PR #4723, opened 2018-12-03
oskar/change_release_procedure

OskarAhl:
* Add `app_2` to valid sections in release config for production and staging
--> Should be able to run `grunt --production/staging=1 section=app_2/app/all`
* Add common 404 and update redirect (remove app and app_2 specific 404)
* Generate new sitemaps for app_2 and remove old sitemap for binary.nex
* Prevent cleanup if only releasing one section to staging
* Update app_2 `index.html` from binarynex to binary.com


src/root_files/_common/404.html
```diff
@@ -0,0 +1,128 @@
...
+                ws.close();
+                return cb(lang);
+            }
+        };
+    }
+
+    var url_lang_segment
+    var redirectFn;
+    if (routeHandler.isApp()) {
+        url_lang_segment = 3;
```
morteza-binary: ditto

src/root_files/_common/404.html
```diff
@@ -0,0 +1,128 @@
...
+        redirectFn = routeHandler.redirect;
+    } else { // app_2
+        url_lang_segment = 4
```
morteza-binary: ditto

src/root_files/app_2/sitemap.app_2.id.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
morteza-binary: @OskarAhl do we server the app on `.me` domain just for `id`?

src/root_files/_common/404.html
```diff
@@ -0,0 +1,128 @@
...
+        }
+        return conLang;
+    }
+
+    var routeHandler = (function routeHandler() {
+        var location = window.location;
+
+        return {
+            isApp: function() {
+                return !location.href.includes('/app/');
```
binary-hamedanchi: This path is only for `app_2`. So, if we add other sections in the future, this won't return a correct value. Preferably, change it to `isApp2()` to check for that section only, as well as the following `if..else`:
https://github.com/binary-com/binary-static/blob/793ab5161e11f5dceb68de6ac15df2a266116820/src/root_files/_common/404.html#L101-L107

src/root_files/app_2/sitemap.app_2.id.xml
OskarAhl: Hmm it's implemented this way for `binary.com`  🤔

build/config/init.js
```diff
@@ -23,17 +27,16 @@ const initGlobals = (grunt) => {
         global.branch_prefix = '';
         global.branch        = global.release_info.target_folder;
 
-        if (global.release_target === 'staging') {
+        if (global.release_target === 'staging' && global.section === 'all') {
             grunt.option('cleanup', true); // always cleanup when releasing to staging
+        } else if (global.release_target === 'staging' && grunt.option('cleanup')) {
+            grunt.fail.fatal('can\'t release only one section to staging with --cleanup');
```
morteza-binary:
```suggestion
             grunt.fail.fatal(`can\'t release only one section to ${global.release_target} with --cleanup`);
```

build/config/init.js
```diff
@@ -23,17 +27,16 @@ const initGlobals = (grunt) => {
         global.branch_prefix = '';
         global.branch        = global.release_info.target_folder;
 
-        if (global.release_target === 'staging') {
+        if (global.release_target === 'staging' && global.section === 'all') {
             grunt.option('cleanup', true); // always cleanup when releasing to staging
+        } else if (global.release_target === 'staging' && grunt.option('cleanup')) {
```
morteza-binary:
```suggestion
         } else if (global.release_info.valid_sections.length > 1 && global.section !== 'all' && grunt.option('cleanup')) {
```

PR #4729, opened 2018-12-03
kelly/data_show_default

kellybinary:
- add `data-show='default'` to handle logged-out clients

### DIFF
--- a/src/javascript/app/pages/new_account.js
+++ b/src/javascript/app/pages/new_account.js
@@ -83,8 +83,8 @@ const NewAccount = (() => {
 
     const onUnload = () => {
         getElementById('footer').setVisibility(1);
-        $google_btn.off('click');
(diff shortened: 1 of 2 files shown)


src/javascript/app/pages/new_account.js
```diff
@@ -83,8 +83,8 @@ const NewAccount = (() => {
 
     const onUnload = () => {
         getElementById('footer').setVisibility(1);
-        $google_btn.off('click');
-        $login_btn.off('click');
+        if ($google_btn) $google_btn.off('click');
+        if ($login_btn) $login_btn.off('click');
```
negarn: should we just move it to `off()` before it does `on()` on line 45? we don't really follow any strict convention for these

kellybinary: oh yes, that's better

PR #4732, opened 2018-12-04
Nazanin/date_of_birth

nazaninreihani:


src/javascript/app/pages/user/new_account/financial_acc_opening.js
```diff
@@ -42,16 +42,16 @@ const FinancialAccOpening = (() => {
             Object.keys(get_settings).forEach((key) => {
                 $element = $(`#${key}`);
                 value    = get_settings[key];
-                if (key === 'date_of_birth') {
+                if (key === 'date_of_birth' && value) {
                     const moment_val = moment.utc(value * 1000);
                     get_settings[key] = moment_val.format('DD MMM, YYYY');
                     $element.attr({
                         'data-value': toISOFormat(moment_val),
+                        'value'     : toISOFormat(moment_val),
```
kellybinary: should this be converted to readable format instead?

nazaninreihani: @negarn @kellybinary What do you suggest?

negarn: in desktop yes, but I believe in mobile there's no way for that.

kellybinary: it seems that the helper file `date_picker.js` will handle the value to be displayed, i think it'll be fine if you remove it

nazaninreihani: I added it to fix the issue when you load the page in mobile view the date of birth is empty although it has value

PR #4733, opened 2018-12-04
Nazanin/fix_currency_list

nazaninreihani:


src/javascript/app/pages/user/get_currency.js
```diff
@@ -43,9 +43,13 @@ const GetCurrency = (() => {
...
         const currency_values = getCurrencyValues();
 
+        const allowed_currencies =
+              Client.getLandingCompanyValue({ real: 1 }, landing_company, 'legal_allowed_currencies');
+
         const available_crypto =
               currency_values.cryptocurrencies.filter(c =>
-                  currency_values.other_currencies.concat(is_crypto ? client_currency : []).indexOf(c) < 0);
+                  currency_values.other_currencies.concat(is_crypto ? client_currency : []).indexOf(c) < 0 &&
+                  allowed_currencies.indexOf(c) > -1);
```
negarn: can't we simply check the `is_suspended` flag in website_status for that currency? that's why they've added it

negarn: maybe your approach is better, we can discuss it

nazaninreihani: I think this way we can show only currencies which are available for this landing company instead of what is supported by us or not

PR #4738, opened 2018-12-05
aaron/profile-details-update

aaron-binary:


src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -238,11 +238,13 @@ const PersonalDetails = (() => {
                     return;
                 }
                 getDetailsResponse(get_settings);
+                if (additionalCheck(get_settings)) {
+                    showFormMessage(localize('Your settings have been updated successfully.'), !is_error);
```
negarn: this is already inside the `!is_error` if condition, so you can simply send `1` or `true` instead

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -238,11 +238,13 @@ const PersonalDetails = (() => {
...
+                }
             });
+        } else { // is_error
+            showFormMessage((getPropertyValue(response, ['error', 'message']) || localize('Sorry, an error occurred while processing your account.')), is_error);
```
negarn: also here, we know it's an error so we can send `0` or `false` instead

PR #4740, opened 2018-12-05
aaron/profile-details-message

aaron-binary:


src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -111,7 +111,7 @@ const PersonalDetails = (() => {
             el_id               = `${should_show_label ? 'lbl_' : ''}${key}`;
             el_key              = CommonFunctions.getElementById(el_id);
             if (el_key) {
-                editable_fields[key] = get_settings[key];
+                editable_fields[key] = (get_settings[key] !== null ? get_settings[key] : '');
```
kellybinary: i think this can be simplified to `get_settings[key] || ''`

negarn: in that case it will change any values that are non-null but equal to false, e.g. `0`, to `''`. We should see if that causes any issues.

kellybinary: I see, some values can be boolean.. let's keep it as it is then :)

PR #4744, opened 2018-12-06
zaki/mt_landing_update

zakibinary:
Changelog
- Update FE due to upcoming changes in API for metatrader landing companies.


src/templates/app/user/metatrader.jsx
```diff
@@ -216,6 +216,17 @@ const Metatrader = () => (
...
+                            description={it.L('Our MetaTrader 5 Advanced account provides you with tight spreads, higher ticket size and offers more products.')}
+                            items={[
+                                it.L('Leverage up to [_1]', '1:100'),
+                                it.L('Variable spreads'),
+                                it.L('Market execution'),
+                                it.L('No commission'),
+                            ]}
+                        />
                         <AccountDesc
                             account_type={['vanuatu_advanced']}
```
binary-hamedanchi: Do we still have `vanuatu_advanced` or it's changed to `labuan_advanced`?

zakibinary: It's going to be changed to `labuan_advanced` but I left it there for the time being until we get better confirmation to how we handle the transition from `vanuatu` to `labuan` for advanced account creation. I'll comment out `vanuatu_advanced` for now. Thanks

src/javascript/app/common/content_visibility.js
```diff
@@ -49,10 +52,17 @@ const ContentVisibility = (() => {
     const init = () => {
         BinarySocket.wait('authorize', 'landing_company', 'website_status').then(() => {
             const current_landing_company_shortcode = State.getResponse('authorize.landing_company_name') || 'default';
+            const mt_financial_company              = State.getResponse('landing_company.mt_financial_company');
+
+            // Check mt_financial_company by account type, since we are offering different landing companies for standard and advanced
+            const mt5fin_shortcodes = Object.keys(mt_financial_company)
```
negarn: could we name it beginning with `arr_`?

src/javascript/app/common/content_visibility.js
```diff
@@ -111,7 +121,7 @@ const ContentVisibility = (() => {
         attr_str,
         current_landing_company_shortcode,
         client_has_mt_company,
-        mt5fin_company_shortcode
+        mt5fin_company_shortcodes
```
negarn: same here

src/javascript/app/common/content_visibility.js
```diff
@@ -131,7 +141,10 @@ const ContentVisibility = (() => {
         else if (is_exclude !== rule_set_has_current) show_element = true;
         if (rule_set_has_eu_country && is_eu_country) show_element = !is_exclude;
 
-        if (mt5fin_rules.includes(mt5fin_company_shortcode)) show_element = !is_exclude;
+        // Check if list of mt5fin_company_shortcodes is array type and filter with defined mt5fin rules
+        if (Array.isArray(mt5fin_company_shortcodes)) {
+            if (mt5fin_company_shortcodes.some(el => mt5fin_rules.includes(el))) show_element = !is_exclude;
+        }
```
negarn: should we handle else as before?

src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -90,6 +90,9 @@ const MetaTraderConfig = (() => {
         };
     })();
 
+    // we need to check if the account type is standard or advanced account before returning landing_company shortcode
+    const checkIfAdvanced = acc_type => `${/_advanced$/.test(acc_type) ? 'advanced' : 'standard'}`;
```
negarn: I would suggest naming it to something like `getMTFinancialAccountType`

src/javascript/app/common/content_visibility.js
zakibinary: Don't think we need `else` in this case as if there's no `array` found, the rules won't be applied and content visibility will ignore the `data-show` rules set in the templates.

PR #4749, opened 2018-12-06
Amam/optimize_tracker

cakasuma:


src/templates/_common/includes/tracker.jsx
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable */
+import React from 'react';
+
+const Tracker = () => (
```
ashkanx: maybe call it `GoogleOptimizer` and comment what it does: https://support.google.com/optimize/answer/7100284

cakasuma: yes, thats sounds better

src/templates/_common/includes/google_optimizer.jsx
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable */
+import React from 'react';
+// Optimize loading with page hiding snippet by temporarily hide the page while the optimize container loads
```
negarn: please fix the grammar issues, perhaps something like:
```
Optimize loading with the following page-hiding snippet by temporarily hiding the page while the optimizer container loads
```

src/templates/app_2/head.jsx
```diff
@@ -9,7 +9,7 @@ import AntiClickjack from '../_common/includes/anti_clickjack.jsx';
 const Head = () => (
     <head>
         {
-            // Prompt a message in the browser if users disabled
+            // Prompt a message in the browser if user's disabled JS
```
negarn: same here, perhaps something like:
```
Prompt a message in the browser if the user has disabled JS
```

src/templates/_common/includes/google_optimizer.jsx
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable */
```
negarn: why are we disabling eslint? please only disable the specific rule, on the specific line/smaller section that is effected

ashkanx: I think we do the same in gtm.jsx. should be because of using `dangerouslySetInnerHTML` ?

negarn: then we should only disable that specific rule that breaks

PR #4756, opened 2018-12-10
kelly/remove_hint

kellybinary:
- remove Asset open hours hint

### DIFF
--- a/src/javascript/app/pages/trade/common_independent.js
+++ b/src/javascript/app/pages/trade/common_independent.js
@@ -133,12 +133,6 @@ const getSelectedOption = ($selector) => {
     return $selected_option;
 };
 
-const showAssetOpenHours = ($selector) => {
-    const $selected_option = $selector ? getSelectedOption($selector) : '';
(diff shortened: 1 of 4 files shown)


src/javascript/app/pages/trade/common_independent.js
```diff
@@ -133,12 +133,6 @@ const getSelectedOption = ($selector) => {
     return $selected_option;
 };
 
-const showAssetOpenHours = ($selector) => {
-    const $selected_option = $selector ? getSelectedOption($selector) : '';
-    const sessions         = $selected_option ? $selected_option.attr('data-sessions') : '';
```
negarn: seems like we can also remove `data-sessions`?

kellybinary: we can't because `data-sessions` is also used to create forward starting dates in the dropdown

negarn: where is that? I only see its usage in starttime.js where it's being set, not checked

PR #4761, opened 2018-12-10
removed charity page from site

brandon-binary:


brandon-binary:
@negarn @ashkanx Thanks for the tip

PR #4770, opened 2018-12-11
aaron/update-allow-equals-tooltip

aaron-binary:


src/templates/app/trade/trading.jsx
```diff
@@ -175,7 +175,7 @@ const Trading = () => (
                                     <div className='row' id='reset_time' />
                                     <div className='row invisible' id='equals_row'>
                                         <input id='callputequal' type='checkbox' />
-                                        <label htmlFor='callputequal'><span data-balloon-length='xlarge' data-balloon={it.L('Win payout if exit spot is also equal to entry spot.')}>{it.L('Allow equals')}</span></label>
+                                        <label htmlFor='callputequal'><span data-balloon-length='large' data-balloon={it.L('Win payout if exit spot is also equal to entry spot.')}>{it.L('Allow equals')}</span></label>
```
negarn: but with large it turns into two lines :(

dialloaliou: What if you totally remove the `data-balloon-length="xlarge"`, it should fit well.

PR #4780, opened 2018-12-12
aaron/authentication-issue

aaimio:


src/javascript/app/base/binary_loader.js
```diff
@@ -142,28 +142,27 @@ const BinaryLoader = (() => {
...
+            <div class="center-text">
+                <div class="gr-padding-10">
+                    <div class="center-text notice-msg">${ error_messages.not_authenticated() }</div>
+                </div>
+                <div class="gr-padding-10">
+                    <button id="sign-out-button">
+                        <span>${ localize('Sign out') }</span>
+                    </button>
+                </div>
+            </div>`
```
negarn: we don't generally set html like this, we create the divs. feel free to instead add the element as a layout component in pages that need it and make it visible when required instead.

PR #4782, opened 2018-12-13
amam/add_unit_tests

cakasuma:
Modules being tested

```Stores/Modules/Trading/Helpers/duration.js```
```Utils/Date/date_time.js```
```Services/Helpers/switch_account.js```
```Services/Helpers/process_data.js```

### CODE REVIEW

File: src/javascript/app_2/Services/Helpers/__tests__/process_data.js [hunk 1]
```diff
@@ -0,0 +1,22 @@
+import { expect } from 'chai';
+import { getOauthAppsObject } from '../process_data.js';
+import BinarySocket from '_common/base/socket_base';
+
+describe('getOauthAppsObject', () => {
+    it('Expects default oauth object when there is no arguments', () => {
+        expect(getOauthAppsObject()).to.eql({ 2: 'Binary.com Autoexpiry' });
+    })
```
kellybinary: missing semicolon?

src/javascript/app_2/Services/Helpers/__tests__/process_data.js
```diff
@@ -0,0 +1,22 @@
+import { expect } from 'chai';
+import { getOauthAppsObject } from '../process_data.js';
+import BinarySocket from '_common/base/socket_base';
+
+describe('getOauthAppsObject', () => {
+    it('Expects default oauth object when there is no arguments', () => {
```
zakibinary: Nitpick: Maybe `are` instead of `is`? 😄

src/javascript/app_2/Utils/Date/__tests__/date_time.spec.js
```diff
@@ -1 +1,87 @@
...
+        expect(DateTime.daysFromTodayTo()).to.eql('');
+    });
+
+    it('return empty string if the user selected previous day', () => {
+        const date = "2018-12-13";
+        expect(DateTime.daysFromTodayTo(date)).to.eql('');
+    });
+
+    it('return difference value between selected date and today', () => {
+        const date = "2018-12-20";
```
kellybinary: this test will break for dates after `2018-12-20`, maybe use a future date some x years from today?

src/javascript/app_2/Utils/Date/__tests__/date_time.spec.js
```diff
@@ -1 +1,89 @@
...
+        //get previous day
+        const date = moment().subtract(1, 'days').format("YYYY-MM-DD");
+        expect(DateTime.daysFromTodayTo(date)).to.eql('');
+    });
+
+    it('return difference value between selected date and today', () => {
+        //get date three days from now
+        const date = moment().add('3', 'days').format("YYYY-MM-DD");
+        const diff = moment(date).utc().diff(moment().utc(), 'days');
+        expect(DateTime.daysFromTodayTo(date)).to.deep.equal(diff + 1);
```
kellybinary: i think it's better to check with 3 instead of `diff + 1`  because we expected the days would be exactly 3 from today. there's no need to add the logic in line 67 (that should be taken care by `daysFromTodayTo()` itself)

src/javascript/app_2/Utils/Date/__tests__/date_time.spec.js
```diff
@@ -1 +1,89 @@
...
+
+describe('convertDuration', () => {
+    const start_time = moment() / 1000; //time epoch now
+    const end_time = moment().add(3, 'minutes') / 1000;
+
+    describe('getDiffDuration', () => {
+        it('return correct value when argument passed', () => {
+            //expecting 180000 == three minutes
+            expect(DateTime.getDiffDuration(start_time, end_time)).to.eql(moment.duration(moment.unix(end_time).diff(moment.unix(start_time))));
+        });
```
kellybinary: here too: you can put `..to.eql(180000)` directly, without adding the logic 😄

cakasuma: indeed, thank you for pointing that out

src/javascript/app_2/Stores/Modules/Trading/Helpers/__tests__/durations.js
```diff
@@ -0,0 +1,135 @@
...
+        expect(Duration.convertDurationUnit("180", "s", "m")).to.eql(3);
+    });
+
+    it('Returns correct value convert minute to day', () => {
+        expect(Duration.convertDurationUnit("2880", "m", "d")).to.eql(2);
+    });
+
+    it('Returns correct value convert second to day', () => {
+        expect(Duration.convertDurationUnit("86400", "s", "d")).to.eql(1);
+    });
```
easteregg: I think we should add a couple of test cases where the argument is there but it is not a valid one. e.g. what should happen when we pass `convertDurationUnit('alpha', 'm', 'd')`? Do we expect null or an error?

src/javascript/app_2/Stores/Modules/Trading/Helpers/__tests__/durations.js
```diff
@@ -0,0 +1,135 @@
...
+    it('Returns correct value for minutes unit', () => {
+        expect(Duration.convertDurationLimit(180, "m")).to.eql(3);
+    });
+
+    it('Returns correct value for hour unit', () => {
+        expect(Duration.convertDurationLimit(7200, "h")).to.eql(2);
+    });
+
+    it('Returns correct value for day unit', () => {
+        expect(Duration.convertDurationLimit(86400, "d")).to.eql(1);
```
easteregg: Same as here, we should test against the failing scenario(s).

src/javascript/app_2/Utils/Date/__tests__/date_time.spec.js
```diff
@@ -1 +1,88 @@
...
+
+describe('daysFromTodayTo', () => {
+    it('return empty string when there is no argument passed', () => {
+        expect(DateTime.daysFromTodayTo()).to.eql('');
+    });
+
+    it('return empty string if the user selected previous day', () => {
+        //get previous day
+        const date = moment().subtract(1, 'days').format("YYYY-MM-DD");
+        expect(DateTime.daysFromTodayTo(date)).to.eql('');
```
easteregg: This is more a question: shouldn't `DateTime.daysFromTodayTo`  return `-1` when yesterday is passed?

cakasuma: i think its because we dont want to get yesterday as arguments ?

src/javascript/app_2/Stores/Modules/Trading/Helpers/__tests__/durations.js
```diff
@@ -0,0 +1,147 @@
...
+    });
+
+    it('Returns correct value convert second to day', () => {
+        expect(Duration.convertDurationUnit('86400', 's', 'd')).to.eql(1);
+    });
+});
+
+describe('getExpiryType', () => {
+    const getServerTime = async () => {
+        const { time: server_time } = await BinarySocket.send({ time: 1 });
```
morteza-binary: @cakasuma 
As I mentioned before `BinarySocket.send` returns an unresolvable promise because there isn't any initialization for WebSocket before it.

BTW, it's better to use mock data instead of getting `server time` from the API. The purpose of the test is to check the correctness of performance of `getExpiryType` function when it receives valid data from the API.

src/javascript/app_2/Stores/Modules/Trading/Helpers/__tests__/durations.js
```diff
@@ -0,0 +1,147 @@
...
+        return server_time;
+    }
+
+    const store = {
+        expiry_date: '2018-12-13',
+        expiry_type: 'duration',
+        root_store: {
+            common: {
+                server_time: getServerTime(),
```
morteza-binary: @cakasuma 
Since there isn't any test, when the `expiry_type` is `endtime`, there isn't need to set `server_time`.

Indeed, it would be better to test all aspect of the function.

PR #4783, opened 2018-12-14
aaron/handle-homepage-redirect

aaron-binary:


src/javascript/app/base/binary_loader.js
```diff
@@ -66,9 +66,9 @@ const BinaryLoader = (() => {
 
         const this_page = e.detail.getAttribute('data-page');
         if (this_page in pages_config) {
```
OskarAhl: I think `hasOwnProperty` is safer here, or is there any reason for looking up in the prototype chain of pages_config?

src/javascript/app/base/binary_loader.js
```diff
@@ -82,7 +82,8 @@ const BinaryLoader = (() => {
         not_authenticated: () => localize('This page is only available to logged out clients.'),
     };
 
-    const loadHandler = (config) => {
+    const loadHandler = (page) => {
+        const config = pages_config[page];
```
OskarAhl: Would it be better to create a copy here instead of referencing the pages_config object? to prevent possible side effects.

PR #4810, opened 2018-12-21
aaron/gtm-deposit-trigger

aaron-binary:


src/javascript/app/pages/user/account/portfolio/portfolio.init.js
```diff
@@ -126,6 +127,11 @@ const PortfolioInit = (() => {
             });
         } else if (response.transaction.action === 'sell') {
             removeContract(response.transaction.contract_id);
+        } else if (response.transaction.action === 'deposit') {
```
ashkanx: better to name the event to something general. for example: `transaction` 
Then pass `transaction_type`, `transaction_reference`, etc.. together with event

PR #4812, opened 2018-12-21
kelly/update_pa_error_note

kellybinary:
- add note to advise user to not press Enter in payment agent withdrawal's description


kellybinary:
https://github.com/binary-com/binary-static/pull/4719

PR #4815, opened 2018-12-24
Amam/set _endtime_err

cakasuma:
fix end time on forward start type contract. based on https://trello.com/c/DCivcW7J card


src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
```diff
@@ -40,10 +40,16 @@ const createProposalRequestForContract = (store, type_of_contract) => {
     const obj_expiry = {};
     if (store.expiry_type === 'endtime') {
         const expiry_date = moment.utc(store.expiry_date);
-        const start_date  = moment.unix(store.start_date || (store.root_store.common.server_time / 1000)).utc();
-        const is_same_day = expiry_date.isSame(start_date, 'day');
-        const expiry_time = is_same_day ? store.expiry_time : '23:59:59';
-        obj_expiry.date_expiry = convertToUnix(expiry_date.unix(), expiry_time);
+        const start_time = store.start_time.split(':');
```
easteregg: I think we can refactor this to define `start_time` only in the else block (a little more efficient).

src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
```diff
@@ -40,10 +40,16 @@ const createProposalRequestForContract = (store, type_of_contract) => {
...
+        if (store.contract_start_type === 'spot') {
+            const expiry_time = store.expiry_time;
+            obj_expiry.date_expiry = convertToUnix(expiry_date.unix(), expiry_time);
+        } else {
+            const start_date  = moment.unix(store.start_date || (store.root_store.common.server_time / 1000)).utc();
+            const is_same_day = expiry_date.isSame(start_date, 'day');
```
easteregg: Since `is_same_day` is not relevant to the `if` statement, we can promote it to upper scope.

src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
```diff
@@ -40,10 +40,16 @@ const createProposalRequestForContract = (store, type_of_contract) => {
...
+            const expiry_time = is_same_day ? store.expiry_time : moment(window.time).hour(start_time[0]).minute(start_time[1]).add(-10, 'minutes').format('HH:mm');
```
easteregg: This is a question: why we are hardcoding `-10` minutes from the `start_time`? (btw you can use `subtract` method and use a positive integer)

src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
```diff
@@ -38,12 +38,19 @@ export const createProposalRequests = (store) => {
...
         const start_date  = moment.unix(store.start_date || (store.root_store.common.server_time / 1000)).utc();
         const is_same_day = expiry_date.isSame(start_date, 'day');
-        const expiry_time = is_same_day ? store.expiry_time : '23:59:59';
-        obj_expiry.date_expiry = convertToUnix(expiry_date.unix(), expiry_time);
+
+        if (store.contract_start_type === 'spot') {
+            const expiry_time = store.expiry_time;
+            obj_expiry.date_expiry = convertToUnix(expiry_date.unix(), expiry_time);
+        } else {
+            const expiry_time = is_same_day ? store.expiry_time : moment().utc().format('HH:mm');
```
negarn: better use server time rather than local time (`moment().utc()`)

src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
```diff
@@ -38,12 +38,19 @@ export const createProposalRequests = (store) => {
...
+            obj_expiry.date_expiry = convertToUnix(expiry_date.unix(), expiry_time);
```
negarn: seems this line is just repeated in both if and else? if so it can be moved outside and set only once

src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
```diff
@@ -36,11 +36,20 @@ export const createProposalRequests = (store) => {
...
     if (store.expiry_type === 'endtime') {
+        const server_time = store.root_store.common.server_time;
+        const start_date  = moment.unix(store.start_date || (server_time / 1000)).utc();
         const expiry_date = moment.utc(store.expiry_date);
-        const start_date  = moment.unix(store.start_date || (store.root_store.common.server_time / 1000)).utc();
         const is_same_day = expiry_date.isSame(start_date, 'day');
-        const expiry_time = is_same_day ? store.expiry_time : '23:59:59';
+        let expiry_time;
+
+        if (store.contract_start_type === 'spot') {
```
negarn: why not  `store.contract_start_type === 'spot' || is_same_day`? seems they both result in `expiry_time = store.expiry_time`

src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
```diff
@@ -36,11 +36,18 @@ export const createProposalRequests = (store) => {
...
+        const start_date  = moment.unix(store.start_date || (server_time / 1000)).utc();
         const expiry_date = moment.utc(store.expiry_date);
-        const start_date  = moment.unix(store.start_date || (store.root_store.common.server_time / 1000)).utc();
         const is_same_day = expiry_date.isSame(start_date, 'day');
-        const expiry_time = is_same_day ? store.expiry_time : '23:59:59';
+        let expiry_time;
+
+        if (store.contract_start_type === 'spot' || is_same_day) {
+            expiry_time = is_same_day ? store.expiry_time : moment(server_time).utc().format('HH:mm');
+        }
```
negarn: it
it would be better to simply remove the if and use 
```
expiry_time = (store.contract_start_type === 'spot' ||  is_same_day) ? store.expiry_time : moment(server_time).utc().format('HH:mm');
 }
```

src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
```diff
@@ -36,11 +36,14 @@ export const createProposalRequests = (store) => {
 
 const createProposalRequestForContract = (store, type_of_contract) => {
     const obj_expiry = {};
+
     if (store.expiry_type === 'endtime') {
+        const server_time = store.root_store.common.server_time;
+        const start_date  = moment.unix(store.start_date || (server_time / 1000)).utc();
```
negarn: does server_time/1000 need to be .toFixed(0)?

morteza-binary: since `store.root_store.common.server_time` is a moment object it doesn't need to divide it by 1000.

src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
```diff
@@ -39,7 +39,7 @@ const createProposalRequestForContract = (store, type_of_contract) => {
 
     if (store.expiry_type === 'endtime') {
         const server_time = store.root_store.common.server_time;
-        const start_date  = moment.unix(store.start_date || (server_time / 1000)).utc();
+        const start_date  = moment.unix(store.start_date || (server_time / 1000).toFixed(0)).utc();
```
morteza-binary: since `store.root_store.common.server_time` is a moment object it doesn't need to divide it by 1000.

src/javascript/app_2/Stores/Modules/Trading/Helpers/proposal.js
```diff
@@ -36,11 +36,20 @@ export const createProposalRequests = (store) => {
 
 const createProposalRequestForContract = (store, type_of_contract) => {
     const obj_expiry = {};
+
     if (store.expiry_type === 'endtime') {
+        const server_time = store.root_store.common.server_time;
+        const start_date  = moment.unix(store.start_date || server_time).utc();
```
morteza-binary: @cakasuma `moment.unix()` return an invalid object. it should be
```suggestion
        const start_date  = moment.utc(store.start_date || server_time).unix();
```

morteza-binary: @cakasuma ignore my last comment. Your implementation is also correct. since `moment.unix(xyz)` returns a moment object.

PR #4816, opened 2018-12-24
aaron/gtm-trigger-on-transaction

aaron-binary:


src/javascript/app/pages/user/account/portfolio/portfolio.init.js
```diff
@@ -32,8 +32,8 @@ const PortfolioInit = (() => {
         BinarySocket.send({ portfolio: 1 }).then((response) => {
             updatePortfolio(response);
         });
-        // Subscribe to transactions to auto update new purchases
-        BinarySocket.send({ transaction: 1, subscribe: 1 }, { callback: transactionResponseHandler });
+        // Wait for transactions to auto update on new purchases
+        BinarySocket.wait('transaction').then(transactionResponseHandler);
```
binary-hamedanchi: The previous version (send) was passing every response to its callback, however by this change it passes only the last response once.

src/javascript/app_2/Stores/Modules/Portfolio/portfolio_store.js
```diff
@@ -62,6 +63,7 @@ export default class PortfolioStore extends BaseStore {
         } else if (act === 'sell') {
             this.removePositionById(contract_id);
         }
+        GTM.pushTransactionData(response, { bom_ui: 'new' });
```
binary-hamedanchi: Would be more reliable if calling this at a higher level like [`socket_general`](https://github.com/binary-com/binary-static/blob/master/src/javascript/app_2/Services/socket_general.js).

src/javascript/_common/base/gtm.js
```diff
@@ -162,10 +162,69 @@ const GTM = (() => {
...
+            const data = {
+                event             : 'transaction',
+                bom_account_type  : ClientBase.getAccountType(),
+                bom_age           : parseInt((moment().utc().unix() - State.getResponse('get_settings.date_of_birth')) / 31557600),
+                bom_email_consent : State.getResponse('get_settings.email_consent'),
+                bom_country       : State.getResponse('get_settings.country'),
+                bom_country_abbrev: State.getResponse('get_settings.country_code'),
+                bom_salutation    : State.getResponse('get_settings.salutation'),
+                bom_firstname     : State.getResponse('get_settings.first_name'),
+                bom_lastname      : State.getResponse('get_settings.last_name'),
```
binary-hamedanchi: Do we need this data there?

src/javascript/_common/base/gtm.js
```diff
@@ -162,10 +162,69 @@ const GTM = (() => {
         pushDataLayer(gtm_data);
     };
 
+    const pushTransactionData = (response, extra_data = {}) => {
+        if (!isGtmApplicable() || ClientBase.get('is_virtual')) return;
+        if (!response.transaction || !response.transaction.action) return;
+
+        const today = moment();
```
binary-hamedanchi: Please, never rely on client's time. It should be always our server time.

src/javascript/_common/base/gtm.js
```diff
@@ -162,10 +162,69 @@ const GTM = (() => {
...
+                bom_email_consent : State.getResponse('get_settings.email_consent'),
+                bom_country       : State.getResponse('get_settings.country'),
+                bom_country_abbrev: State.getResponse('get_settings.country_code'),
+                bom_salutation    : State.getResponse('get_settings.salutation'),
+                bom_firstname     : State.getResponse('get_settings.first_name'),
+                bom_lastname      : State.getResponse('get_settings.last_name'),
+                bom_today         : moment().utc().unix(),
+                transaction       : {
+                    id      : response.transaction.transaction_id,
+                    type    : response.transaction.action,
```
binary-hamedanchi: Are we going to push this for all actions including buy?
If yes, then we need to remove the redundancy with `pushPurchaseData`.

negarn:
failing travis

src/javascript/_common/base/gtm.js
```diff
@@ -60,7 +60,7 @@ const GTM = (() => {
             bom_country_abbrev: get_settings.country_code,
             bom_email         : get_settings.email,
             url               : window.location.href,
-            bom_today         : Math.floor(Date.now() / 1000),
+            bom_today         : window.time,
```
easteregg: `window.time` is`undefined` for me in chrome and firefox. Did we create the value ourself?

src/javascript/_common/base/gtm.js
```diff
@@ -162,10 +162,55 @@ const GTM = (() => {
         pushDataLayer(gtm_data);
     };
 
+    const pushTransactionData = (response, extra_data = {}) => {
+        if (!isGtmApplicable() || ClientBase.get('is_virtual')) return;
```
easteregg: @morteza-binary What do you suggest we do to not use the `ClientBase` and use the store value instead?

src/javascript/app/base/subscription_manager.js
```diff
@@ -0,0 +1,204 @@
...
+                subscribers: [fncCallback],
+            };
+
+            BinarySocket.send(request_obj, {
+                callback: (response) => dispatch(response, sub_id),
+            });
+        } else if (!hasCallbackFunction(sub_id, fncCallback)) {
+            // there is already an active subscription for the very same request which fncCallback is not subscribed to it yet
+            subscriptions[sub_id].subscribers.push(fncCallback);
+        }
```
easteregg: **Q:** What happens if `sub_id` is false and `else if` is not applicable? should we throw?

src/javascript/_common/base/gtm.js
negarn: you would have to wait for the response of time before expecting the value.

src/javascript/_common/base/gtm.js
negarn: why? it's already imported here, is there any harm?

easteregg: Since the change of no-reloading client switch, it will create inconsistency with our source of truth, which is`client_store`. The value should be handled by that object, see: 

https://github.com/binary-com/binary-static/blob/master/src/javascript/app_2/Stores/client_store.js#L120-L123

morteza-binary: @easteregg you're right we will face some issues in the future when we want to split the new_ui and `binary_static` and keep the data of `client_store` like other stores by `base_store` but unfortunately, this file is used directly from both projects.
We can make a card to refactor it in the future.

src/javascript/app/base/subscription_manager.js
morteza-binary: when there isn't value for `sub_id` (is `false`) the `if` condition becomes `true` so it's not important that the `else if` is applicable or not.

src/javascript/app/base/subscription_manager.js
```diff
@@ -0,0 +1,204 @@
+import BinarySocket     from './socket';
```
morteza-binary: @aaron-binary 
It'd be better to move the `subscribtion_manager.js` from `src/javascript/app_2/Services/` to `src/javascript/_common/` and use it in both projects instead of duplicating it.

src/javascript/_common/base/gtm.js
```diff
@@ -52,6 +52,7 @@ const GTM = (() => {
             pushDataLayer({ bom_affiliate_token: affiliate_token.t });
         }
 
+        const now = moment(window.time || Date.now()).unix().utc();
```
negarn: I suggest `moment(window.time || undefined)` instead of using a combination of Date and moment

negarn: I don't think we can turn unix value to utc, it should be first utc and then unix if you want a unix value in the end. if so, perhaps it's better to rename the variable to make it clear what value now is supposed to be, something like `moment_now` or `unix_now`

src/javascript/_common/base/gtm.js
```diff
@@ -162,10 +163,55 @@ const GTM = (() => {
         pushDataLayer(gtm_data);
     };
 
+    const pushTransactionData = (response, extra_data = {}) => {
```
negarn: seems we're only pushing deposit and withdrawal data, not other transactions, perhaps renaming this function or adding a comment to explain that will help

src/javascript/_common/base/gtm.js
```diff
@@ -69,7 +70,7 @@ const GTM = (() => {
         }
 
         if (!ClientBase.get('is_virtual')) {
-            data.bom_age       = parseInt((moment().unix() - get_settings.date_of_birth) / 31557600);
+            data.bom_age       = data.bom_today.diff(get_settings.date_of_birth, 'year');
```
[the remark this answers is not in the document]
aaron-binary: I incorrectly used the `diff` function, the first argument should be a `moment`, so I believe I should update it to:
- `moment_now.diff(moment.unix(get_settings.date_of_birth).utc(), 'year');`

`moment.unix()`: Create a `moment` from a unix timestamp (in seconds)
`utc()`: Use UTC standard and not the client's timezone

src/javascript/_common/base/gtm.js
```diff
@@ -162,10 +163,55 @@ const GTM = (() => {
         pushDataLayer(gtm_data);
     };
 
+    const pushTransactionData = (response, extra_data = {}) => {
+        if (!isGtmApplicable() || ClientBase.get('is_virtual')) return;
+        if (!response.transaction || !response.transaction.action) return;
+        if (!['deposit', 'withdrawal'].includes(response.transaction.action)) return;
+ 
+        const now         = moment.unix(window.time || Date.now()).utc();
```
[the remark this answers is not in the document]
aaron-binary: > same here about using `window.time || undefined`

Instead of `undefined` should we use `moment().utc()` as this will be more clear

src/javascript/_common/base/gtm.js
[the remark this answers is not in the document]
negarn: this should work, but diff works with epoch, year, or moment. it just doesn't work with a unix value properly.

src/javascript/_common/base/gtm.js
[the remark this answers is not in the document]
negarn: sure, then you can change it to `window.time || moment.utc()`

src/javascript/app/base/socket_general.js
```diff
@@ -79,6 +80,7 @@ const BinarySocketGeneral = (() => {
                         BinarySocket.send({ get_account_status: 1 });
                         BinarySocket.send({ payout_currencies: 1 });
                         BinarySocket.send({ mt5_login_list: 1 });
+                        SubscriptionManager.subscribe('transaction', { transaction: 1, subscribe: 1 }, () => false);
```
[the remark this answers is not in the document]
aaron-binary: We require the subscription manager because in order to push as many `deposit` and `withdrawal` transactions we need to subscribe to the `transaction`-stream right after a client has authorised, but we also need to maintain the functionality in the **Portfolio** page—where the client _again_ tries to subscribe to the `transactions`-stream to update the page in real-time. Without the subscription manager the API would give us a `AlreadySubscribed` error.

src/javascript/_common/base/gtm.js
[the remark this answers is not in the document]
aaron-binary: You're correct, I checked the wrong (duration-)part of the docs. 👍

src/javascript/_common/base/gtm.js
```diff
@@ -162,10 +162,55 @@ const GTM = (() => {
...
+                localStorage.removeItem(storage_key);
+                gtm_transactions = { timestamp: now.format('X') };
+            }
+        }
+        const transactions_arr = gtm_transactions.transactions || [];
+        if (!transactions_arr.includes(response.transaction.transaction_id)) {
+            const data = {
+                event            : 'transaction',
+                bom_account_type : ClientBase.getAccountType(),
+                bom_email_consent: State.getResponse('get_settings.email_consent'),
```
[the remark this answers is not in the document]
ashkanx: no, we don't need the `email consent`

src/javascript/_common/base/gtm.js
```diff
@@ -12,7 +12,7 @@ const getPropertyValue = require('../utility').getPropertyValue;
 const getAppId         = require('../../config').getAppId;
 
 const GTM = (() => {
-    const isGtmApplicable = () => (/^(1|1098|14473|15284)$/.test(getAppId()));
+    const isGtmApplicable = () => (/^(1|1098|14473|15284|15760)$/.test(getAppId()));
```
ashkanx: Do not commit this

ashkanx:
@aaron-binary tests are failing

src/javascript/_common/base/gtm.js
```diff
@@ -162,10 +164,53 @@ const GTM = (() => {
...
+    const pushTransactionData = (response, extra_data = {}) => {
+        if (!isGtmApplicable() || ClientBase.get('is_virtual')) return;
+        if (!response.transaction || !response.transaction.action) return;
+        if (!['deposit', 'withdrawal'].includes(response.transaction.action)) return;
+ 
+        const moment_now  = window.time || moment().utc();
+        const storage_key = 'GTM_transactions';
+        
+        // Remove values from prev days so localStorage doesn't grow to infinity
+        let gtm_transactions  = JSON.parse(localStorage.getItem(storage_key)) || {};
```
negarn: extra whitespace before `=`

src/javascript/_common/base/gtm.js
```diff
@@ -162,10 +164,53 @@ const GTM = (() => {
...
+        if (Object.prototype.hasOwnProperty.call(gtm_transactions, 'timestamp')) {
+            if (moment_now.isAfter(moment.utc().unix(gtm_transactions.timestamp), 'day')) {
```
negarn: probably should be `moment.unix(gtm_transactions.timestamp).utc()` instead

PR #4817, opened 2018-12-25
[new-app] morteza/app2_logos

morteza-binary:
Refactored `build/copy.js` to copy appropriate logos and manifest file for each section


kellybinary:
@morteza-binary LGTM

PR #4818, opened 2018-12-25
[new-app] trading page (or any pages) won't load

easteregg:
Since the new loading indicator, react won't load and an exception is being thrown in the console. Apparently it attempts to initiate UI store and mount react on a non-existent `document.body`.


easteregg:
@cakasuma, yeah, basically waits for the entire DOM load before mounting react. Although store and network are initiated before that.

morteza-binary:
@easteregg 
It's better to uncomment this part of code in `app2/index.html`
https://github.com/binary-com/binary-static/blob/4aa5eea56a2d7f44579cfd0d65ebfadd8650ef2d/src/javascript/app_2/index.js#L14-L26

I commented out them because I couldn't reproduce the issue after changing the process of loading and rendering JS files.

PR #4820, opened 2018-12-26
mohammad/cashier_session_expired

binary-hamedanchi:
Add cashier session expired page.


binary-hamedanchi:
@cakasuma The redirection happens in the iframe.

PR #4847, opened 2019-01-02
easteregg/change_digit_chart_to_table

easteregg:


src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,101 @@
+const DIGIT_BLOCK_SIZE        = 36;
+const DIGIT_BLOCK_LEFT_OFFSET = 66;
+const DIGIT_BLOCK_CORRECTION  = 2;
```
negarn: I don't think we do this, all constants for this should be inside DigitTicker, same as the digits array, and should be in lowercase

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,101 @@
...
+
+const DigitTicker = (() => {
+    let $container, source;
+    const digits = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9];
```
negarn: probably good to start the variable name with `arr_` prefix

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,101 @@
+const DIGIT_BLOCK_SIZE        = 36;
+const DIGIT_BLOCK_LEFT_OFFSET = 66;
+const DIGIT_BLOCK_CORRECTION  = 2;
+
+const DigitTicker = (() => {
+    let $container, source;
```
negarn: generally should be in two separate lines,
```
let $container,
      source;
```

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,101 @@
...
+
+        $container = $(`#${container_id}`);
+        $container
+            .append($('<div />', { class: 'peek-box' })
+                .append($('<div />', { class: 'mask', text: '0/0' }))
+                .append($('<div />', { class: 'peek' })))
+            .append($('<div />', {
+                class: 'digits', html : digits.map(digit => `<div class='digit digit-${digit}'>${digit}</div>`).join(''),
+            }));
+        colorWinningBoxes(winningNumbers(contract));
```
negarn: if we already have source as global, why do we need to pass contract around?

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,101 @@
...
+            case 'DIGITMATCH':
+                return digits.filter(digit => +digit === +contract.barrier);
+            case 'DIGITDIFF':
+                return digits.filter(digit => +digit !== +contract.barrier);
+            case 'DIGITODD':
+                return digits.filter(digit => +digit % 2 !== 0);
+            case 'DIGITEVEN':
+                return digits.filter(digit => +digit % 2 === 0);
+            default:
+                throw new Error('Cannot Determine Winning numbers.');
```
negarn: I think for default we shouldn't make any of them highlighted as winning numbers, instead of exiting with an error? this might cause BE to not be able to release new digit contract types without breaking FE

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,101 @@
...
+        }
+    };
+
+    const colorWinningBoxes = (winning_numbers) => {
+        winning_numbers.forEach(digit => {
+            $(`.digit-${digit}`).css({
+                backgroundColor: '#9ddfcb',
+                color          : '#67b9a0',
+            });
```
negarn: it's better to add a digit-winning or similar class to it instead, and keep the styling for that class in the css file. we don't want css styling in javascript files.

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,101 @@
...
+        });
+    };
+
+    const isRunning = (digit) => /[\d]/g.test(digit);
+
+    const calculateOffset = (digit) => isRunning(digit) ?
+        (+digit * DIGIT_BLOCK_SIZE) + DIGIT_BLOCK_LEFT_OFFSET - DIGIT_BLOCK_CORRECTION : 0;
+
+    const getTotal = () => source.tick_count;
```
negarn: not sure if this just adds ambiguity to have an extra function, rather than calling source.tick_count directly

src/javascript/app/pages/trade/digit_ticker.js
negarn: not sure if it's more consistent to rename source to contract, and pass in the variable as proposal_open_contract, source seems a bit vague to me

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,101 @@
...
+
+    const isWon = () => source.status === 'won';
+
+    const update = (tick_count, current) => {
```
negarn: can we rename current to something more meaningful? current what?

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,101 @@
...
+        $container.removeClass('invisible');
+        const peekBox = $('.peek-box');
```
negarn: should be renamed to $peek_box

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,101 @@
...
+        const peek    = $('.peek');
```
negarn: same here, should be renamed to $peek

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,101 @@
...
+        const mask    = $('.peek-box > .mask');
```
negarn: and $mask

negarn: if we already have $peek_box, this can be changed to
```
const mask = $peek_box.find(' > .mask');
```

src/javascript/app/pages/trade/digit_ticker.js
negarn: and tick_count is unclear too, perhaps current_tick_count? I'm not sure what this is indicating here

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,101 @@
...
+        peekBox.addClass('running');
+
+        mask.text(`${tick_count}/${getTotal()}`);
+
+        peek.text(current);
+
+        peekBox.css({
+            left           : `${calculateOffset(current)}px`,
+            color          : '#ffffff', backgroundColor: '#454545',
+        });
```
negarn: same here, better add a class that indicates its position, and keep the css in the css file.

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,101 @@
...
+                color: '#077453',
+            });
+        } else {
+            peekBox.css({
+                backgroundColor: '#cb0433',
+                color          : '#fffff',
+            });
+            peek.css({
+                color: '#cb0433',
+            });
```
negarn: in these cases as well.

src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -0,0 +1,94 @@
...
+const ViewPopupUI          = require('../user/view_popup/view_popup.ui');
+const showLocalTimeOnHover = require('../../base/clock').showLocalTimeOnHover;
+const BinarySocket         = require('../../base/socket');
+const localize             = require('../../../_common/localize').localize;
+const getPropertyValue     = require('../../../_common/utility').getPropertyValue;
+
+const DigitDisplay = (() => {
+    let $container,
+        contract,
+        tick_count;
```
negarn: perhaps can be renamed to current_tick_count as well.

src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -0,0 +1,94 @@
...
+        $container = $(`#${id_render}`);
+        $container
+            .addClass('normal-font')
+            .html($('<h5 />', { text: contract.display_name, class: 'center-text' }))
+            .append($('<div />', { class: 'gr-6 gr-centered' })
+                .append($('<div />', { class: 'gr-row', id: 'table_digits' })
+                    .append($('<strong />', { class: 'gr-3', text: localize('Tick') }))
+                    .append($('<strong />', { class: 'gr-3', text: localize('Spot') }))
+                    .append($('<strong />', { class: 'gr-6', text: localize('Spot Time (GMT)') }))))
+            .append($('<div />', { class: 'digit-ticker invisible', id: 'digit-ticker-container' }));
```
negarn: id is generally separated by `_`

src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -0,0 +1,94 @@
...
+
+        const request     = {
```
negarn: this is my bad, extra empty space before the `=`

src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -0,0 +1,94 @@
...
+            ticks_history: contract.underlying,
+            start        : contract.date_start,
+        };
+        if (contract.current_spot_time < contract.date_expiry) {
```
negarn: perhaps a comment above this line to explain this means that the contract is still ongoing/streaming

src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -0,0 +1,94 @@
...
+            request.subscribe = 1;
+            request.end       = 'latest';
+        } else {
+            request.end = contract.date_expiry;
+        }
+        DigitTicker.init('digit-ticker-container', contract);
```
negarn: seems like a weird place to call this, can't it be called before we initialize our request variable?

src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -0,0 +1,94 @@
...
+    const updateTable = (spot, time) => {
+        $container
+            .find('#table_digits')
+            .append($('<p />', { class: 'gr-3', text: tick_count }))
+            .append($('<p />', { class: 'gr-3 gray', html: tick_count === contract.tick_count ? `${spot.slice(0, spot.length - 1)}<strong>${spot.slice(-1)}</strong>` : spot }))
+            .append($('<p />', { class: 'gr-6 gray digit-spot-time no-underline', text: moment(+time * 1000).utc().format('YYYY-MM-DD HH:mm:ss') }));
+
+        DigitTicker.update(
+            tick_count,
+            spot.slice(-1),
```
negarn: since this is repeated here and on line 48, perhaps worth saving it in a last_digit variable

src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -0,0 +1,94 @@
...
+            contract
```
negarn: seems we are only using contract.tick_count and contract.status in DigitTicker, not sure if it's better to only send those two or the whole contract object

src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -0,0 +1,94 @@
...
+        } else if (response.tick) {
+            updateTable(response.tick.quote, response.tick.epoch);
+            tick_count += 1;
+        }
+        showLocalTimeOnHover('.digit-spot-time');
+    };
+
+    const end = (proposal_open_contract) => {
+        contract = proposal_open_contract;
+        // here we need to highlight won/lost
```
negarn: I left this so we could pass the final status to DigitTicker, assuming that view popup is opened while contract is streaming and we don't know if won/lost, since contract gets past only on init, so it needs to be passed again once it ends.

src/javascript/app/pages/trade/digit_ticker.js
easteregg: This is according to the zeplin design. Maybe we should pull them in in this conversation.

src/javascript/app/pages/trade/digit_ticker.js
easteregg: Position is dynamic, we can't use that in css.

src/javascript/app/pages/trade/digit_trade.js
easteregg: This is total tick count. I updated variable names to be more clear.

src/javascript/app/pages/trade/digit_trade.js
easteregg: Calling the end is not required. I removed it.

src/javascript/app/pages/trade/digit_ticker.js
easteregg: The need for global contract is resolved.

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,124 @@
...
+            <div class='digits'>
+                ${array_of_digits.map(digit => `<div class='digit digit-${digit}'>${digit}</div>`).join('')}
+            </div>
+        `;
+        paintWinningNumbers(winningNumbers(contract_type, barrier));
+        observeResize();
+    };
+
+    const observeResize = () => {
+        window.onresize = () => {
```
negarn: do we need to off resize somewhere?

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,124 @@
...
+            calculateOffset();
+        };
+    };
+
+    const paintWinningNumbers = (winning_numbers) => {
+        winning_numbers.forEach(digit => {
+            const element = document.querySelector(`.digit-${digit}`);
```
negarn: is this something that can be searched within container instead of the whole document?

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,124 @@
...
+    };
+
+    const success = () => {
+        $peek_box.classList.remove('digit-losing', 'digit-running');
+        $peek.classList.remove('digit-losing', 'digit-running');
+        $peek_box.classList.add('digit-winning');
+        $peek.classList.add('digit-winning');
+    };
+
+    const queryDom = () => {
```
negarn: I think this should be renamed to something like `setElements` or similar

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,124 @@
...
+    // Calculate left margin of the peek-box against the "width" of the container
+    const calculateLeftMargin = (width) => (width - (digit_block_size * 10)) / 2;
+
+    // Calculate peek-box left offset.
+    const calculateOffset = () => {
+        const left_offset = calculateLeftMargin($container.offsetWidth);
+        return (+current_spot * digit_block_size) + left_offset - style_offset_correction;
+    };
+
+    const fail = () => {
```
negarn: I think this should be moved closer to the success function.

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,124 @@
...
+        $mask.innerText = `${current_tick_count}/${total_tick_count}`;
+        $peek.innerText = current_spot;
+
+        $peek_box.classList.add('digit-running');
+        $peek.classList.add('digit-running');
+
+        $peek_box.setAttribute('style', `transform: translateX(${calculateOffset()}px)`);
+
+        if (contract_status === 'won') {
+            success();
```
negarn: I would also suggest renaming these, perhaps to markSuccess and markFailure, just to make it clear what the function is doing somehow

src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -0,0 +1,99 @@
...
+            .find('#table_digits')
+            .append($('<p />', { class: 'gr-3', text: tick_count }))
+            .append($('<p />', { class: 'gr-3 gray', html: tick_count === contract.tick_count ? `${spot.slice(0, spot.length - 1)}<strong>${last_digit}</strong>` : spot }))
+            .append($('<p />', { class: 'gr-6 gray digit-spot-time no-underline', text: moment(+time * 1000).utc().format('YYYY-MM-DD HH:mm:ss') }));
+
+        DigitTicker.update(
+            tick_count,
+            {
+                quote: last_digit,
+            }
```
negarn: we don't need to send the last status to mark won/lost?

src/javascript/app/pages/trade/digit_ticker.js
easteregg: all methods are ordered alphabetically. Should I not do that?

negarn: not really. generally we order them with init being the first function, any onunload sort of function at the bottom, and then the most relevant functions together in between.

src/javascript/app/pages/trade/digit_trade.js
easteregg: We send the last status. But calling additional method is not required.

src/sass/app/trade.scss
```diff
@@ -1011,3 +1011,141 @@
 .show-mb {
     display: none;
 }
+
+.digit-ticker {
```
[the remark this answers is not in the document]
easteregg: Hmm, looks like test link doesn't have CSS changes.

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,146 @@
...
+        if (contract_status === 'won') {
+            markAsWon();
+        }
+        if (contract_status === 'lost') {
+            markAsLost();
+        }
+    };
+
+    const remove = () => {
+        window.onresize = null;
```
negarn: do we want to only remove that specific function? what if there are other resize functions being set from others that we want to keep?

easteregg: there is always 1 resize function. they don't stack.

src/javascript/app/pages/trade/digit_ticker.js
ruskakimov: @negarn @easteregg emm we don't really have a design for this edge-case 😅, not sure what you are referring to

src/sass/app/components/peek_box.scss
```diff
@@ -0,0 +1,292 @@
...
+        @media screen and (max-width: 480px) {
+            margin-top: 6px;
+            min-width: initial;
+            width: 35px;
+            grid-template-rows: 4px 19px 30px 4px;
+        }
+    }
+    .mask {
+        grid-area: mask;
+        font-size: 10px;
```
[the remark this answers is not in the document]
easteregg: On mobile devices the number `10 / 10` will break the line, that was why 10px was used.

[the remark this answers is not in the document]
easteregg: I made the size 11 px for desktop and 10px for mobile.

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,242 @@
...
+            }
+        };
+    };
+
+    // Calculate left margin of the peek-box against the "width" of the container
+    const calculateLeftMargin = (width) => (width - (digit_block_size * 10)) / 2;
+
+    // Calculate peek-box left offset.
+    const calculateOffset = () => {
+        const left_offset = calculateLeftMargin(el_container.offsetWidth);
```
kellybinary: instead of calculating left margin, how about getting `offsetLeft` of the element `.digit-won`? 
this way, we know the exact position of `.digit-won` in the container with respect to the screen sizes :)

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,242 @@
...
+            i++;
+            current += increment;
+
+            element.innerHTML = render(current.toFixed(decimal_points));
+
+            if (i <= 15) {
+                requestAnimationFrame(renderTick);
+            } else {
+                current = end;
+                element.innerHTML = render(current.toFixed(decimal_points));
```
kellybinary: do we need to call cancelAnimationFrame here?

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,242 @@
...
+        el_peek.classList.add('digit-running');
+
+        el_peek_box.setAttribute('style', `transform: translateX(${calculateOffset()}px)`);
+    };
+
+    const remove = () => {
+        window.onresize = null;
+    };
+
+    const countDecimals = function(value) {
```
kellybinary: arrow function () => here ?

ruskakimov:
also when contract ends `#trading_socket_container #contract_prices_container #contract_purchase_profit_list>div>p { margin: auto 0 5px; }` is no longer applied, resulting in additional margin that moves everything once the contract ends

src/sass/app/components/peek_box.scss
```diff
@@ -0,0 +1,306 @@
...
+        & .digit-tick-epoch {
+            display: inline-block;
+            vertical-align: top;
+            margin-right: 2px;
+            top: 14px;
+            font-size: 14px;
+            line-height: 14px;
+            color: $COLOR_PEEK_RUNNING_GRAY;
+            padding-left: 14px;
+            padding-right: 2px;
```
ruskakimov: you can remove `padding-right: 2px` as well, the point was to right align timestamp with the `.quote`

src/sass/app/components/peek_box.scss
```diff
@@ -0,0 +1,306 @@
+$PEEKBOX_GRID_ITEMS: mask, topright, topleft, top, left, right, bottom, bottomleft, bottomright;
+/* stylelint-disable color-no-hex */
```
zakibinary: Maybe the put colors in constants? And  `/* stylelint-disable color-no-hex */` should be re-enabled after this block.

src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -0,0 +1,242 @@
...
+        observeResize();
+    };
+
+    const populateContainer = (container_element) => {
+        // remove previous elements and start fresh.
+        while (container_element.firstChild) {
+            container_element.removeChild(container_element.firstChild);
+        }
+
+        const temp_epoch_el = document.createElement('div');
```
zakibinary: Just a nitpick perhaps, but would it be better to have the `DOM` elements created in a static template for this?

src/javascript/app/pages/trade/digit_ticker.js
easteregg: Good point.

src/javascript/app/pages/trade/digit_ticker.js
easteregg: I don't think we need to call it, since we are calling this function 16 times and then bail.

src/sass/app/components/peek_box.scss
```diff
@@ -0,0 +1,306 @@
...
+        & .quote {
+            width: auto;
+            color: $COLOR_PEEK_LIGHT_GRAY;
+            font-weight: 100;
+            margin-top: 2px;
+        }
+        & .digit-tick-epoch {
+            display: inline-block;
+            vertical-align: top;
+            margin-right: 2px;
```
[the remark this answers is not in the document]
easteregg: :joy:

src/sass/app/components/peek_box.scss
easteregg: Since they are not going to be used anywhere else other than the peek_box, I believe this a better way to put everything beside each other. Enabling stylelint after that tho :)

src/javascript/app/pages/trade/digit_ticker.js
easteregg: Since we render these dynamically, the option of having them embedded in the `view-popup` or `smart trader` will make additional DOM elements that are not required by all purchase types. In `app_2`, I think creating the template is a sane choice, tho 👍

src/sass/app/components/peek_box.scss
easteregg: all done.

src/sass/app/components/peek_box.scss
```diff
@@ -0,0 +1,306 @@
...
+        background-color: transparent;
+        display: grid;
+        grid-template-areas: 'topleft top topright' 'left mask right' 'left peek right' 'bottomleft bottom bottomright';
+        grid-template-rows: 4px 19px 36px 4px;
+        grid-template-columns: 4px 1fr 4px;
+        align-content: center;
+        justify-items: center;
+        transition: all 0.9s ease;
+        position: absolute;
+        top: 3px;
```
ruskakimov: `top: 2px`

src/sass/app/components/peek_box.scss
```diff
@@ -0,0 +1,306 @@
...
+            grid-area: bottom;
+            height: 4px;
+            width: 100%;
+            background-color: $COLOR_PEEK_RED;
+        }
+        .bottomleft {
+            grid-area: bottomleft;
+            border-bottom-left-radius: 4px;
+            width: 4px;
+            height: 4px;
```
ruskakimov: `height: 5px`

src/sass/app/components/peek_box.scss
```diff
@@ -0,0 +1,306 @@
...
+        .bottomright {
+            grid-area: bottomright;
+            border-bottom-right-radius: 4px;
+            width: 4px;
+            height: 4px;
+            background-color: $COLOR_PEEK_RED;
+        }
+        .bottom {
+            grid-area: bottom;
+            height: 4px;
```
ruskakimov: `height: 5px`

src/sass/app/components/peek_box.scss
```diff
@@ -0,0 +1,306 @@
...
+        position: absolute;
+        top: 3px;
+        left: 0;
+        z-index: 5;
+
+        .bottomright {
+            grid-area: bottomright;
+            border-bottom-right-radius: 4px;
+            width: 4px;
+            height: 4px;
```
ruskakimov: `height: 5px`

src/sass/app/trade.scss
```diff
@@ -528,8 +530,7 @@
                 padding: 0 10px;
 
                 > p {
-                    margin: auto 0 5px;
-                    padding-top: 10px;
+                    margin: auto 0 0;
```
ruskakimov: any reason for no margin here? if we remove all styles from this paragraph then the jump disappears :)

ruskakimov: @easteregg please check this one out, once you are free :)

easteregg: @ruskakimov Will do sir.

easteregg: Done. you are right, I missed the jumping results already 😂

PR #4859, opened 2019-01-07
kelly/update_tick_chart_label

kellybinary:


src/javascript/app/common/chart_settings.js
```diff
@@ -17,7 +17,7 @@ const ChartSettings = (() => {
         labels = labels || { // needs to be inside setLabels function so localize works
             barrier_line  : `<div class='nowrap gr-padding-10 gr-parent'><span style="${common_horizontal_line_style} border-color: green; border-style: solid;"></span>${localize('Barrier')}&nbsp;</div>`,
             barrier_spot  : `<div class='nowrap gr-padding-10 gr-parent'><span style="${common_horizontal_line_style} border-color: green; border-style: dotted;"></span>${localize('Barrier')}&nbsp;</div>`,
-            end_time      : `<div class='nowrap gr-padding-10 gr-parent'><span style="${common_vertical_line_style} border-color: #e98024; border-style: dashed;"></span>${localize('End Time')}&nbsp;</div>`,
+            end_time      : `<div class='nowrap gr-padding-10 gr-parent'><span style="${common_vertical_line_style} border-color: #e98024; border-style: dashed;"></span>${params.is_tick_trade ? localize('Exit Spot') : localize('End Time')}&nbsp;</div>`,
```
zakibinary: Should we store the colors as `const`? 
It'll be easier to update in the future instead of reading through the text.

We should probably should use hex codes for all colors, like in `L19` for example, to make things more consistent.

PR #4862, opened 2019-01-08
amam/help_another_test

cakasuma:
## base_name.js

this is to assist  testing base_name.js. can refer as solution


src/javascript/app_2/Utils/URL/__tests__/base_name.js
```diff
@@ -0,0 +1,19 @@
+import { expect } from 'chai';
+import getBaseName from '../base_name';
+const Url                     = require('../../../../_common/url');
+const { setURL } = require('../../../../_common/__tests__/tests_common');
```
OskarAhl: Could you format the import location to start at the same line?

cakasuma: no worries, its already copied to the other pr, hence i closed this pr

PR #4869, opened 2019-01-09
ahkuma/Prof check message

ahkuma:


src/javascript/app/pages/user/account/settings/professional_client.js
```diff
@@ -28,10 +28,20 @@ const professionalClient = (() => {
         }
 
         const status = State.getResponse('get_account_status.status') || [];
-        if (is_in_page && /professional/.test(status)) {
+        if (is_in_page && status === 'professional') {
```
negarn: this doesn't work, status is an array.

ahkuma: I'll revert it back to regex. I think that works.

src/javascript/app/pages/user/account/settings/professional_client.js
```diff
@@ -28,10 +28,20 @@ const professionalClient = (() => {
         }
 
         const status = State.getResponse('get_account_status.status') || [];
-        if (is_in_page && /professional/.test(status)) {
+        if (is_in_page && /^professional$/.test(status)) {
```
negarn: also doesn't work, if status is an array with multiple items. you can fake it in console and just try it. use an array like ['age_verified', 'professional'] and try your regex.

negarn: because you are using ^ and $, it won't match. use something else to test if an item is exactly inside an array. there are multiple ways.

src/javascript/app/pages/user/account/settings/professional_client.js
```diff
@@ -28,10 +28,20 @@ const professionalClient = (() => {
...
             $('#loading').remove();
             $('#frm_professional').setVisibility(0);
-            $(`#${/professional_requested/.test(status) ? 'processing' : 'professional'}`).setVisibility(1);
+            $('#professional').setVisibility(1);
+            return;
+        } else if (is_in_page && /^professional_requested$/.test(status)) {
```
tazeek: Why not try using `status.includes(...)`?

src/javascript/app/pages/user/account/settings/professional_client.js
ahkuma: @negarn why was regex used in the old code if that was the case?

src/javascript/app/pages/user/account/settings/professional_client.js
```diff
@@ -28,10 +28,20 @@ const professionalClient = (() => {
...
+            return;
+        } else if (is_in_page && status.includes('professional_requested')) {
+            $('#loading').remove();
+            $('#frm_professional').setVisibility(0);
+            $('#processing').setVisibility(1);
+            return;
+        } else if (is_in_page && status.includes('professional_rejected')) {
+            $('#loading').remove();
+            $('#frm_professional').setVisibility(0);
+            $('#rejected').setVisibility(1);
```
negarn: since these three lines are always repeated and the only difference is the id to make visible, why not write a function and pass the id of the element to make visible to it? this will stop us having to repeat code and make changing it in the future easier.

src/javascript/app/pages/user/account/settings/professional_client.js
negarn: without ^ and $ it works, but it catches anything that has the word professional in it, such as professional_rejected, which is what caused the issue in the first place.

src/javascript/app/pages/user/account/settings/professional_client.js
ahkuma: @negarn  Where should we organize that function?

negarn: it can be outside this function, after init function or so

src/javascript/app/pages/user/account/settings/professional_client.js
ahkuma: I see. That makes sense.

src/javascript/app/pages/user/account/settings/professional_client.js
```diff
@@ -18,6 +18,12 @@ const professionalClient = (() => {
         populateProfessionalClient(is_financial);
     };
 
+    const setVisible = (set_visible) => {
```
negarn: perhaps renaming to `selector` would be better, set_visible sounds like a boolean, not a selector

ahkuma: You are right

PR #4872, opened 2019-01-09
ahkuma/New acc status

ahkuma:


src/javascript/app/base/header.js
```diff
@@ -275,43 +275,47 @@ const Header = (() => {
...
-                tnc                  : () => Client.shouldAcceptTnc(),
-                unwelcome            : () => hasStatus('unwelcome'),
-                withdrawal_locked    : () => hasStatus('withdrawal_locked'),
+                authenticate          : () => +get_account_status.prompt_client_to_authenticate,
+                cashier_locked        : () => hasStatus('cashier_locked'),
+                currency              : () => !Client.get('currency'),
+                document_needs_action : () => hasStatus('document_needs_action'),
+                document_review       : () => hasStatus('document_under_review'),
+                document_expired      : () => hasStatus('document_expired'),
+                document_expiring_soon: () => hasStatus('document_expiring_soon'),
```
negarn: this is sorted alphabetically, these two should be two lines above

src/javascript/app/base/header.js
```diff
@@ -275,43 +275,47 @@ const Header = (() => {
...
+                currency              : () => buildMessage(localizeKeepPlaceholders('Please set the [_1]currency[_2] of your account.'),                                                                                                'user/set-currency'),
+                document_needs_action : () => buildMessage(localizeKeepPlaceholders('[_1]Your Proof of Identity or Proof of Address[_2] did not meet our requirements. Please check your email for further instructions.'),             'user/authenticate'),
+                document_review       : () => buildMessage(localizeKeepPlaceholders('We are reviewing your documents. For more details [_1]contact us[_2].'),                                                                           'contact'),
+                document_expired      : () => buildMessage(localize('Your proof-of-identity documents have expired and your account has been limited. Please send updated documents to customer support.')),
+                document_expiring_soon: () => buildMessage(localize('Your proof-of-identity documents will expire within the next month. Please send updated documents to customer support to prevent your account being limited.')),
```
negarn: same here, it should be alphabetically sorted

src/javascript/app/base/header.js
```diff
@@ -275,43 +275,47 @@ const Header = (() => {
...
+                risk                  : () => buildMessage(localizeKeepPlaceholders('Please complete the [_1]financial assessment form[_2] to lift your withdrawal and trading limits.'),                                               'user/settings/assessmentws'),
+                tax                   : () => buildMessage(localizeKeepPlaceholders('Please [_1]complete your account profile[_2] to lift your withdrawal and trading limits.'),                                                        'user/settings/detailsws'),
+                unwelcome             : () => buildMessage(localizeKeepPlaceholders('Trading and deposits have been disabled on your account. Kindly [_1]contact customer support[_2] for assistance.'),                                'contact'),
+                withdrawal_locked     : () => localize('Withdrawals have been disabled on your account. Please check your email for more details.'),
+                tnc                   : () => buildMessage(has_no_tnc_limit
```
negarn: the reason why tnc is not alphabetically sorted is that it has multiple lines, so we kept it at the bottom separately

PR #4880, opened 2019-01-10
Nazanin/purchase_button

nazaninreihani:
includes changes for longcode


src/javascript/app_2/Modules/Trading/Containers/purchase.jsx
```diff
@@ -35,15 +40,32 @@ const Purchase = ({
...
-                className='primary green'
+                className='primary green btn-purchase'
                 has_effect
-                text={localize('Purchase')}
                 onClick={() => { onClickPurchase(info.id, info.stake, type); }}
                 wrapperClassName='submit-section'
-            />
+            >
+                <React.Fragment>
+                    <div className='btn-purchase-rec' />
```
kellybinary: could you use **BEM** for styling?

src/javascript/app_2/Modules/Trading/Components/Form/Purchase/contract_info.jsx
```diff
@@ -3,60 +3,63 @@ import React             from 'react';
...
+        this.previous_price = null;
+    }
+
+    getIconType = () => {
+        let icon_type = '';
+        const contract_basis = this.props.contract_basis.value;
+        const price = this.props.proposal_info[contract_basis];
+        if (price !== this.previous_price) {
+            icon_type = price > this.previous_price ? 'profit' : 'loss';
+            this.previous_price = price;
```
kellybinary: i think it's better to use `state` and `setState` for `previous_price` and `icon_type`. you can define it before constructor as such: 
```
state = {
   previous_price: null,
   icon_type: null,
};
```

src/javascript/app_2/Modules/Trading/Components/Form/Purchase/contract_info.jsx
```diff
@@ -3,60 +3,63 @@ import React             from 'react';
...
+                        <div className='info-wrapper'>
+                            <div>{localize('[_1]', this.props.contract_basis.text)}</div>
+                            <div>
+                                <Money
+                                    amount={this.props.proposal_info[this.props.contract_basis.value]}
+                                    currency={this.props.currency}
+                                />
+                            </div>
+                            <div className='icon_price_move_container'>
+                                { iconType && <IconPriceMove type={iconType} /> }
```
kellybinary: here you can pass `type={this.state.icon_type}` in `<IconPriceMove />`

src/javascript/app_2/Modules/Trading/Components/Form/Purchase/contract_info.jsx
```diff
@@ -3,60 +3,63 @@ import React             from 'react';
...
-                        <div><Money amount={proposal_info.payout} currency={currency} /></div>
-                    </div>
-                    <div className='info-wrapper'>
-                        <div>{localize('Net Profit')}:</div>
-                        <div><Money amount={proposal_info.profit} currency={currency} /></div>
+        this.previous_price = null;
+    }
+
+    getIconType = () => {
+        let icon_type = '';
```
kellybinary: it's better to check changes in props using React's `getDerivedStateFromProps` method, then compare the state of `previous_price` with the new price and do update on the states using `setState` accordingly

nazaninreihani: I'm not sure if it's a good idea, as it's mentioned to better not use it by react,
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html
``A common misconception is that getDerivedStateFromProps and componentWillReceiveProps are only called when props “change”. These lifecycles are called any time a parent component rerenders, regardless of whether the props are “different” from before. Because of this, it has always been unsafe to unconditionally override state using either of these lifecycles. Doing so will cause state updates to be lost.``.
why is that we need to use it?

nazaninreihani: we want to show green icon for rise and red for fall of price and when the price is same we should have no icon, there are cases that getDerivedStateFromProps fires and price is not changed yet, and it goes in the condition of having same prev and current price, and the icon will be removed although really we didn't have change in proposal_info

kellybinary: thanks, i was thinking since we are getting price whenever the `proposal_info` prop changes, i thought it's better to check for a change in props, and hence suggesting the use of `getDerivedStateFromProps` method

kellybinary: with the latest changes you made, it seems to work fine 😄

src/javascript/app_2/Modules/Trading/Containers/purchase.jsx
```diff
@@ -28,22 +30,38 @@ const Purchase = ({
     trade_types,
 }) => (
     Object.keys(trade_types).map((type, idx) => {
-        const info        = proposal_info[type] || {};
-        const is_disabled = !is_purchase_enabled || !is_trade_enabled || !info.id || !is_client_allowed_to_visit;
+        const info               = proposal_info[type] || {};
+        const is_disabled        = !is_purchase_enabled || !is_trade_enabled || !info.id || !is_client_allowed_to_visit;
```
kellybinary: please remove the extra spaces before `=`

src/javascript/app_2/Stores/Modules/Trading/Helpers/__tests__/proposal.js
```diff
@@ -1,31 +1,51 @@
-import { expect }    from 'chai';
-import React         from 'react';
-import * as Proposal from '../proposal';
+import { expect }                from 'chai';
+import React                     from 'react';
+import * as Proposal             from '../proposal';
```
kellybinary: here too: please remove the extra spaces 😄

PR #4886, opened 2019-01-14
negar/el_last_digit_prediction

negarn:
add last digit prediction element


src/javascript/app_2/App/Components/Form/digit_selector.jsx
```diff
@@ -0,0 +1,40 @@
+import PropTypes from 'prop-types';
+import React     from 'react';
+
+class DigitSelector extends React.Component {
```
OskarAhl: I think this doesn't have to be a class? can be a pure component?

src/sass/app_2/modules/trading.scss
```diff
@@ -156,6 +160,27 @@
             padding: 0;
             background: transparent;
         }
+        .digit-selector {
```
OskarAhl: Should we use BEM naming for the digit-selector component?

src/javascript/app_2/App/Components/Form/digit_selector.jsx
```diff
@@ -0,0 +1,51 @@
...
+    const handleSelect = (item) => {
+        if (+item.target.getAttribute('data-value') !== selected_digit) {
+            onChange({ target: { name, value: +item.target.getAttribute('data-value') } });
+        }
+    };
+
+    return (
+        <div className='digit-selector center-text'>
+            <div>
+                {[...Array(5).keys()].map(i =>
```
OskarAhl: to prevent repetition perhaps we can make a function for this and reuse at line 30? `makeRow(rowItems, rowIndexStart)...` ?

src/javascript/app_2/App/Components/Form/digit_selector.jsx
```diff
@@ -0,0 +1,51 @@
...
+    selected_digit,
+}) => {
+    const handleSelect = (item) => {
+        if (+item.target.getAttribute('data-value') !== selected_digit) {
+            onChange({ target: { name, value: +item.target.getAttribute('data-value') } });
+        }
+    };
+
+    return (
+        <div className='digit-selector center-text'>
```
OskarAhl: Why not add center text as a part of the `digit-selector` block?

negarn: I think it's a common class, we use it in app, it's quite useful. do you think it's better to give center alignment to everything that needs it?

src/javascript/app_2/App/Components/Form/digit_selector.jsx
negarn: why not just store it as a variable `const arr_five = [...Array(5).keys()];`? mapping doesn't change the original value of it and it can be reused

src/javascript/app_2/App/Components/Form/digit_selector.jsx
OskarAhl: Since the `center-text` belongs to the `digit-selector` block I think it's easier to read if it's within the actual `digit-selector` css block

negarn: so we want to repeat styles instead of using common classes from now on?

negarn: haha. that sounds like more refactoring than BEM, probably :) for now I've removed the center-text from here.

PR #4892, opened 2019-01-14
ahkuma/refactor_timepicker

ahkuma:


src/javascript/app_2/App/Components/Form/TimePicker/time_picker_dropdown.jsx
```diff
@@ -0,0 +1,139 @@
...
+            <div className={`${preClass}-dropdown ${this.props.className}`}>
+                <div
+                    className={`${preClass}-panel`}
+                    onClick={toggle}
+                >
+                    <span className={value ? '' : 'placeholder'}>{value || localize('Select time')}</span>
+                    {(!('is_clearable' in this.props) || this.props.is_clearable) &&
+                        <span
+                            className={`${preClass}-clear`}
+                            onClick={this.clear}
```
kellybinary: i think need to add `IconClear` as well? it isn't clear where to click to clear the timepicker 🤔

src/javascript/app_2/App/Components/Form/TimePicker/time_picker.jsx
```diff
@@ -0,0 +1,123 @@
...
+
+    componentDidMount() {
+        document.addEventListener('mousedown', this.handleClickOutside);
+    }
+
+    componentWillUnmount() {
+        document.removeEventListener('mousedown', this.handleClickOutside);
+    }
+
+    toggleDropDown = () => {
```
zakibinary: I suggest we rename this to `toggleTimePicker` as we already have a `DropDown` component.

src/javascript/app_2/App/Components/Form/TimePicker/time_picker.jsx
```diff
@@ -0,0 +1,123 @@
+import { observer }       from 'mobx-react';
+import PropTypes          from 'prop-types';
+import React              from 'react';
+import TimePickerDropdown from './time_picker_dropdown.jsx';
+import InputField         from '../input_field.jsx';
+
+class TimePicker extends React.PureComponent {
```
zakibinary: @morteza-binary , shouldn't this be `React.Component` instead? Correct me if I'm wrong.

ahkuma: @zakibinary Morteza was the one who suggest me to use PureComponent instead.

src/javascript/app_2/App/Components/Form/TimePicker/time_picker.jsx
```diff
@@ -0,0 +1,123 @@
...
+    constructor (props) {
```
zakibinary: We can remove the `constructor` using something like this:
```state = {
    is_open: false, 
value: '',
  };```

File: src/javascript/app_2/App/Components/Form/TimePicker/time_picker.jsx [hunk 5]
```diff
@@ -0,0 +1,123 @@
...
+                            max={end_time}
+                        />
+                        : (
+                            <React.Fragment>
+                                <InputField
+                                    error_messages={validation_errors}
+                                    type='text'
+                                    is_read_only
+                                    id={`${prefix_class}-input`}
+                                    className={`${prefix_class}-input ${this.state.is_open ? 'active' : ''}`}
```
zakibinary: Nitpick: We can use `classNames` library to make this cleaner and more readable instead of string template.

src/javascript/app_2/App/Components/Form/TimePicker/time_picker.jsx
```diff
@@ -0,0 +1,123 @@
...
+                                    value={value}
+                                    onClick={this.toggleDropDown}
+                                    name={name}
+                                    placeholder={placeholder}
+                                />
+                                <TimePickerDropdown
+                                    className={`${this.state.is_open ? 'active' : ''} from-left`}
```
zakibinary: Ditto here for `classNames` library.

src/javascript/app_2/App/Components/Form/TimePicker/time_picker.jsx
ahkuma: Agree. I was planning to use that too but I hold it off until I can work the popover class, so that timepicker and datepicker don't need 2 different implementation to do the same thing.

src/sass/app_2/_common/form/time_picker.scss
```diff
@@ -23,72 +22,46 @@
...
     .time-picker-dropdown {
         font-size: 16px;
         position: absolute;
         top: 0;
-        display: flex;
-        flex-direction: column;
-        height: 0;
+        z-index: 10;
+        height: 100%;
+        max-height: 261px;
```
zakibinary: Do we need `max-height`?

src/javascript/app_2/App/Components/Form/TimePicker/time_picker.jsx
zakibinary: Not sure what do you mean by two different implementations? I don't see the relevance in that with how we should refactor the `classNames`.

src/sass/app_2/_common/form/time_picker.scss
ahkuma: Actually, I am not sure. It's something that I copied when I checked how DatePicker's animation is implemented.

src/javascript/app_2/App/Components/Form/TimePicker/time_picker.jsx
ahkuma: I suppose I should treat this as a one off thing just in case I don't get the chance to work on what I wanted to.

zakibinary: The `Popover` component just clones the child elements/component and wraps itself around as a parent `component`, you will still need to refactor the `classNames` for the child components used. 
Either way, it's up to you whether you refactor it now or later on a supplementary card. 

I just assumed since the PR is called `Refactor/Timepicker`, it is more relevant here.

src/sass/app_2/_common/form/time_picker.scss
kellybinary: @ahkuma i'm actually changing the datepicker animation to follow dropdown animation using CSSTransition 😄, could you do the update for timepicker as well?

ahkuma: @kellybinary Instead of doing it seperately, could we write a common animation and maybe put it somewhere like `animation.scss`? so we don't need 2 people to do 2 different thing that does the same thing?

src/javascript/app_2/App/Components/Form/TimePicker/dialog.jsx
```diff
@@ -0,0 +1,66 @@
+import React                          from 'react';
+import { localize }                   from '_common/localize';
+import { toMoment }                   from 'Utils/Date';
+import classNames                     from 'classnames';
```
zakibinary: Whitespaces here, and import order is incorrect. `classNames` should be at top.

zakibinary: Same here, import order is wrong for `classNames`.

src/javascript/app_2/App/Components/Form/TimePicker/dialog.jsx
```diff
@@ -0,0 +1,66 @@
...
+            <div className={`${preClass}-selector`}>
+                <div className={`${preClass}-hours`}>
+                    <div className='list-title center-text'><strong>{localize('Hour')}</strong></div>
+                    <div className='list-container'>
+                        {hours.map((h, key) => {
+                            to_compare_moment.hour(h);
+                            const is_enabled = to_compare_moment.isBetween(start_time_moment, end_time_moment);
+                            return (
+                                <div
+                                    className={classNames(`list-item`, `${hour === h ? 'selected' : ''}`, `${is_enabled ? '' : 'disabled'}`)}
```
zakibinary: You don't need to use back-ticks inside `classNames` unless it's a string template.
For `boolean` based classes you can do like this:
e.g - `'select': (hour === h)`

src/javascript/app_2/App/Components/Form/TimePicker/time_picker.jsx
```diff
@@ -0,0 +1,134 @@
...
+    componentDidMount() {
+        document.addEventListener('mousedown', this.handleClickOutside);
+    }
+
+    componentWillUnmount() {
+        document.removeEventListener('mousedown', this.handleClickOutside);
+    }
+
+    toggleDropDown = () => {
+        this.setState({ is_open: !this.state.is_open });
```
kellybinary: better to pass function that returns new state when calling setState here: 
`this.setState((state) => ({ is_open : !state.is_open }));`

src/javascript/app_2/App/Components/Form/TimePicker/time_picker.jsx
```diff
@@ -0,0 +1,134 @@
...
+import { observer } from 'mobx-react';
+import PropTypes    from 'prop-types';
+import React        from 'react';
+import Dialog       from './dialog.jsx';
+import InputField   from '../input_field.jsx';
+
+class TimePicker extends React.PureComponent {
+    state = {
+        is_open: false,
+        value  : '',
```
kellybinary: `this.state.value` is not used anywhere?

src/javascript/app_2/App/Components/Form/TimePicker/dialog.jsx
```diff
@@ -0,0 +1,71 @@
...
+function selectOption(type, value, props, is_enabled = true) {
+    if (is_enabled && props.value) {
+        const [ prev_hour, prev_minute ] = props.value.split(':');
+        if ((type === 'h' && value !== prev_hour) || (type === 'm' && value !== prev_minute)) {
+            props.onChange(`${type === 'h' ? value : prev_hour}:${type === 'm' ? value : prev_minute}`);
+        }
+    }
+}
+
+function Dialog(props) {
```
OskarAhl: Why not use the same style as our other functional stateless components? I Think it's easier for consistency if we write them the same way 😃 
```
const Dialog = ({
 destructure props here
}) => { ... }
```

src/javascript/app_2/App/Components/Form/TimePicker/dialog.jsx
```diff
@@ -0,0 +1,71 @@
+import classNames   from 'classnames';
+import React        from 'react';
+import { localize } from '_common/localize';
+import { toMoment } from 'Utils/Date';
+
+function selectOption(type, value, props, is_enabled = true) {
```
OskarAhl: Think it's better if you move this to the `Duration` component? The function is only related to that component and that would follow the same style as our other components

src/javascript/app_2/App/Components/Form/TimePicker/dialog.jsx
```diff
@@ -0,0 +1,78 @@
+import classNames   from 'classnames';
+import React        from 'react';
+import { localize } from '_common/localize';
+import { toMoment } from 'Utils/Date';
+
+const Dialog = ({
```
OskarAhl: I think you need to add proptypes for this component as well `Dialog.PropTypes = { ... }` (runtime type checking of the props)

src/javascript/app_2/App/Components/Form/TimePicker/time_picker.jsx
zakibinary: @ahkuma , any updates on this? `PureComponent` will give an error/warning since it calls `shouldComponentUpdate()`.

kellybinary: i think better to use React.Component if we're going to attach an observer to this component

ahkuma: @zakibinary @kellybinary I am just going to use `Components` as it seems to be the best way. If we were to use `observer` with `PureComponent` it's going to apply a `shouldComponentUpdate()` anyway.

negarn:
@ahkuma conflicts

ahkuma:
> @ahkuma conflicts

merged

PR #4904, opened 2019-01-17
kelly/datepicker

kellybinary:
- update logic so calendar week to start on **Monday** instead of **Sunday**
- add `IconCalendarToday` at calendar footer
- replace current DatePickerInput to use `InputField`:
  - separate `InputField` for duration picker and date picker
- show `IconClear` onHover - should clears the datepicker onClick
- change datepicker display format to **DD MMM YYYY**
- refactor style using BEM
- localize month and weeks headers


src/javascript/app_2/App/Components/Elements/Calendar/panels/calendar_days.jsx
```diff
@@ -37,13 +53,17 @@ const getDays = ({ calendar_date, date_format, max_date, min_date, start_date, o
...
+        const is_other_month = moment_date.month() !== moment_cur_date.month();
         days.push(
             <span
                 key={date}
-                className={classNames('calendar-date', {
-                    active  : is_active && !is_disabled,
-                    today   : is_today,
-                    disabled: is_disabled,
+                className={classNames('calendar__cell', {
+                    'calendar__cell--is-active'     : is_active && !is_disabled,
```
easteregg: I think it is better to use the modifier name, instead of its binary value. For example, `calendar__cell--active`, `calendar__cell--disabled`, etc. extra word is not needed, since `--mod` should be clear enough to understand the modifier.

src/javascript/app_2/App/Components/Elements/Calendar/calendar.jsx
```diff
@@ -141,7 +141,7 @@ class Calendar extends React.PureComponent {
         const { calendar_date, calendar_view, selected_date  } = this.state;
 
         return (
-            <div id={id} className='calendar' value={selected_date}>
+            <div id={id || undefined} className='calendar' data-value={selected_date}>
```
OskarAhl: won't `id` already be undefined from line 140 if it's not passed as a prop?

src/javascript/app_2/App/Components/Elements/Calendar/calendar_button.jsx
```diff
@@ -1,22 +1,25 @@
...
+const CalendarButton = ({
+    children,
+    className,
+    is_hidden,
+    label,
+    onClick,
+}) => (
+    <React.Fragment>
+        { !is_hidden &&
+            <span
```
OskarAhl: should this be a `<button />` instead of a span?

src/javascript/app_2/App/Components/Form/DatePicker/date_picker.jsx
```diff
@@ -102,14 +106,45 @@ class DatePicker extends React.PureComponent {
...
+            data-value={value || undefined}
+            error_messages={this.props.validation_errors}
+            is_read_only
+            name={this.props.name}
+            onClick={this.handleVisibility}
+            placeholder={this.props.placeholder || localize('Select date')}
+            type='text'
+            value={toMoment(value).format('DD MMM YYYY')}
+        />
+    );
```
OskarAhl: What about adding a `input_field_props` object shared between `renderDurationPickerInputField` and `renderDatePickerInputField `  to reduce the props duplication?

```javascript
const inputFieldProps = {
  className: `datepicker__input`,
  data-tip: false,
  etc....
}

// and then
renderDurationPickerInputField = (value) => (
<InputField 
 {...inputFieldProps}
/>


renderDatePickerInputField = (value) => (
<InputField 
 {...inputFieldProps}
/>
```

ashkanx:
@kellybinary conflict

PR #4908, opened 2019-01-18
Nazanin/app_2_components_tests_elements

nazaninreihani:


src/javascript/app_2/App/Components/Elements/AccountSwitcher/__tests__/upgrade_button.spec.js
```diff
@@ -0,0 +1,18 @@
...
+configure({ adapter: new Adapter() });
+
+describe('UpgradeButton', () => {
+    it('should render one <UpgradeButton /> component', () => {
+        const wrapper = shallow(<UpgradeButton />);
+        expect(wrapper).to.have.length(1);
+    });
+    it('should have onClick as an instance of function when passed', () => {
+        const wrapper = shallow(<UpgradeButton onClick={() => true} />);
+        expect(wrapper.find('Button').prop('onClick')).to.be.an.instanceof(Function);
```
easteregg: When we are testing against the prop-types, we need to make sure that proper errors are thrown upon getting error. Since you are passing onClick as a function, and then testing the same behavior, we can't really make sure if underlying prop validations are in place or not. So, what happens when we pass an object, for example, instead of a function? 

I think it is better to test out the PropTypes of the component, and either trust that propTypes are in place and we are good to go, or test the component and stub the console to see the error.

src/javascript/app_2/App/Components/Elements/ToastMessage/__tests__/close_button.spec.jsx
```diff
@@ -0,0 +1,18 @@
...
+configure({ adapter: new Adapter() });
+
+describe('CloseButton', () => {
+    it('should render one <CloseButton /> component', () => {
+        const wrapper = shallow(<CloseButton />);
+        expect(wrapper).to.have.length(1);
+    });
+    it('should have onClick as an instance of function when passed', () => {
+        const wrapper = shallow(<CloseButton onClick={() => true} />);
+        expect(wrapper.find('Button').prop('onClick')).to.be.an.instanceof(Function);
```
easteregg: same as here.

src/javascript/app_2/App/Components/Elements/ToastMessage/__tests__/toast.spec.jsx
```diff
@@ -0,0 +1,18 @@
...
+configure({ adapter: new Adapter() });
+
+describe('Toast', () => {
+    it('should render one <Toast /> component', () => {
+        const wrapper = shallow(<Toast />);
+        expect(wrapper).to.have.length(1);
+    });
+    it('should have onClick as an instance of function when passed', () => {
+        const wrapper = shallow(<Toast onClick={() => true} />);
+        expect(wrapper.find('.toast__body').prop('onClick')).to.be.an.instanceof(Function);
```
easteregg: ditto.

src/javascript/app_2/App/Components/Elements/__tests__/localize.spec.js
```diff
@@ -0,0 +1,23 @@
...
+        const wrapper = shallow(<Localize str={str} />);
+        expect(wrapper).to.have.length(1);
+    });
+    it('should render with replacers', () => {
+        const str = 'something with [_1] replacer [_2]';
+        const replacers = {
+            '1_2': <a className='a-cool-classname' />,
+        };
+        const wrapper = shallow(<Localize str={str} replacers={replacers} />);
+        expect(wrapper).to.have.length(1);
```
easteregg: Can we change the assertion to make sure that `<a className='a-cool-classname'> replacer </a>` is rendered? (e.g. `shallow.find('.a-cool-classname').length`)

src/javascript/app_2/App/Components/Elements/Notifications/__tests__/notifications.spec.js
```diff
@@ -0,0 +1,22 @@
...
+    it('should render one <Notifications /> component', () => {
+        const wrapper = shallow(<Notifications />);
+        expect(wrapper).to.have.length(1);
+    });
+    it('should render .no-notifications-container when the list is not passed', () => {
+        const wrapper = shallow(<Notifications />);
+        expect(wrapper.find('.no-notifications-container').exists()).to.be.true;
+    });
+    it('should not render .no-notifications-container when the list is passed', () => {
+        const wrapper = shallow(<Notifications list={['First', 'Second', 'Third']} />);
```
easteregg: I think one of the pros of writing tests is to find out code smells and improve the code as a result. This is one of those cases, I am not sure if that would fall under this very same PR or not, but while we are writing tests, it is better to refactor as well.

For example, here, instead of looking for 'attached classes' we can check if the component contains `<Notification />` or `<EmptyNotification />`. So we can refactor the code and change the following: 

```jsx
<div className='no-notifications-container'>
    <div className='notification-message'>
        <div className='bell-icon'>
            <IconBell />
        </div>
        <div>
             <h4>{localize('No Notifications')}</h4>
             <span className='no-notifications-message'>{localize('You have yet to receive any notifications')}</span>
        </div>
    </div>
</div>
```

to Something like this: 
```jsx
<EmptyNotification />
```

And then Test the `<EmptyNotification />` component by itself. 

> Note to self: Writing tests at this stage (when the development is considered as done) has always comes with dangerous areas. On of those areas that we should be all aware of while writing them is when we just try to pass the tests and adjust them with the current code, while we should be targeting the correct behaviour. If tests are failing and we get red (and we know that test is right). That is more than alright to just modify the code and not the test itself.

src/javascript/app_2/App/Components/Elements/SettingsDialog/__tests__/language_dialog.spec.js
```diff
@@ -0,0 +1,30 @@
...
+        expect(wrapper.find('.show').exists()).to.be.false;
+    });
+    it('should not have .show when is_visible is false and is_settings_on is true', () => {
+        const wrapper = shallow(<LanguageDialog is_visible={false} is_settings_on={true} />);
+        expect(wrapper.find('.show').exists()).to.be.false;
+    });
+    it('should not have .show when is_visible is false and is_settings_on is false', () => {
+        const wrapper = shallow(<LanguageDialog is_visible={false} is_settings_on={false} />);
+        expect(wrapper.find('.show').exists()).to.be.false;
+    });
```
easteregg: I think we should add one truth-y test as well, just to make sure the data received rendered correctly.

src/javascript/app_2/App/Components/Elements/ErrorBox/__tests__/error_box.spec.js
```diff
@@ -0,0 +1,19 @@
...
+
+describe('ErrorBox', () => {
+    it('should render one <ErrorBox /> component', () => {
+        const wrapper = shallow(<ErrorBox />);
+        expect(wrapper).to.have.length(1);
+    });
+    it('should render header as passed to it', () => {
+        const wrapper = shallow(<ErrorBox header='This is a header' />);
+        expect(wrapper.find('.page-error-header').text()).to.be.eql('This is a header');
+    });
```
easteregg: I think we should test the children rendering as well. Like what you did in `settings_control.spec.js`. Or even refactor that test into a general function. So any components that contains children can be tested with that function.

src/javascript/app_2/App/Components/Elements/ToastMessage/__tests__/close_button.spec.jsx
```diff
@@ -0,0 +1,14 @@
...
+import CloseButton            from '../close_button.jsx';
+
+configure({ adapter: new Adapter() });
+
+describe('CloseButton', () => {
+    it('should render one <CloseButton /> component', () => {
+        const wrapper = shallow(<CloseButton />);
+        expect(wrapper).to.have.length(1);
+    });
+});
```
easteregg: Should we test the required `onClick` propType? I mean, closeButton without close functionality, is just an icon, right? should we check that the component requires onClick prop to be passed? [See here](https://reactjs.org/docs/typechecking-with-proptypes.html)

src/javascript/app_2/App/Components/Elements/ToastMessage/__tests__/transition.spec.js
```diff
@@ -0,0 +1,23 @@
...
+describe('Transition', () => {
+    it('should render one <Transition /> component', () => {
+        const wrapper = shallow(<Transition />);
+        expect(wrapper).to.have.length(1);
+    });
+    it('should render children when passed in', () => {
+        const child_div = <div className='sweet-child-of-mine' />;
+        const wrapper = shallow(
+            <Transition>
+                { child_div }
```
easteregg: Regarding test functionality extraction.

src/javascript/app_2/App/Components/Elements/__tests__/popover.spec.js
```diff
@@ -0,0 +1,36 @@
...
+        const wrapper = shallow(<Popover subtitle='This is a subtitle'>
+            <div></div>
+        </Popover>);
+        expect(wrapper.find('.popover-subtitle').exists()).to.be.true;
+    });
+    it('should have the state is_open equal to true on mouseenter and false on mouseleave', () => {
+        const wrapper = shallow(<Popover subtitle='This is a subtitle'>
+            <div className='sweet-child-of-mine'></div>
+        </Popover>);
+        wrapper.find('.sweet-child-of-mine').simulate('mouseenter');
```
easteregg: Apparently we should avoid `.simulate` entirely, [see this](https://github.com/airbnb/enzyme/issues/1551).

src/javascript/app_2/App/Components/Elements/ToastMessage/__tests__/close_button.spec.jsx
nazaninreihani: we can add isRequired to the onclick prop but still it passes the test without passing onclick prop, shouldn't it fail?

src/javascript/app_2/App/Components/Elements/SettingsDialog/__tests__/language_dialog.spec.js
nazaninreihani: there are 4 conditions and just one of them is truthy, that's when both props is_visible and is_settings are true, this case is tested

src/javascript/app_2/App/Components/Elements/ToastMessage/__tests__/close_button.spec.jsx
easteregg: It drops a `console.warn`, we should check out the actual console output or mock it to capture that and assert that it works. Or, we can ignore that part of testing and just make sure that `isRequired` exists in the component's PropTypes.onclick property.

src/javascript/app_2/test_helper.js
```diff
@@ -0,0 +1,12 @@
+import React                  from 'react';
+import { expect }             from 'chai';
+import { configure, shallow } from 'enzyme';
+import Adapter                from 'enzyme-adapter-react-16';
+
+configure({ adapter: new Adapter() });
+
+export const TestChildren = (component) => {
```
easteregg: if it is possible, use camelCase for `TestChildren`.

src/javascript/app_2/App/Components/Elements/Notifications/notifications.jsx
```diff
@@ -14,24 +13,14 @@ const Notifications = ({ list }) => (
...
-                        <div className='bell-icon'>
-                            <IconBell />
-                        </div>
-                        <div>
-                            <h4>{localize('No Notifications')}</h4>
-                            <span className='no-notifications-message'>{localize('You have yet to receive any notifications')}</span>
-                        </div>
-                    </div>
-                </div>
+                <EmptyNotification />
```
easteregg: 💯

src/javascript/app_2/App/Components/Elements/SettingsDialog/__tests__/language_dialog.spec.js
easteregg: You are right, I was thinking about implementing test cases for `getAllowedLanguages` as well. But that should not be test here, we should write one for `getAllowedLanguages` on it's own specification file.

src/javascript/app_2/App/Components/Elements/ToastMessage/__tests__/close_button.spec.jsx
```diff
@@ -11,4 +11,8 @@ describe('CloseButton', () => {
         const wrapper = shallow(<CloseButton />);
         expect(wrapper).to.have.length(1);
     });
+    it('should render one <CloseButton /> component', () => {
+        const wrapper = shallow(<CloseButton />);
+        expect(wrapper.prop('onClick').isRequired).to.be.true();
```
easteregg: 👍

easteregg:
Would you please resolve the conflicts?

nazaninreihani:
> Would you please resolve the conflicts?

@easteregg Done:)

PR #4911, opened 2019-01-22
amam/added_display_dialog_message

cakasuma:
For GDPR cookie usage notice
- added style for dialog message

---

check whether the the client is logged in and check whether the client has cookieConsent set in the cookie -> check isEucountry -> display dialog message -> if click ok -> set cookieConsent.


src/javascript/app/base/footer.js
```diff
@@ -27,6 +28,42 @@ const Footer = (() => {
         const $status_notification = $('#status_notification');
         $status_notification.slideUp(200);
     };
+
+    const adjustElevioAndScrollup = (elevio_height, scrollup_height) => {
+        const $elevio_button = $('#_elev_io ._6byvm');
+        const $scrollup = $('#scrollup');
+        $elevio_button.attr('style', `bottom: ${elevio_height}px !important`);
+        $scrollup.attr('style', `bottom: ${$elevio_button.height() + scrollup_height}px`);
```
negarn: it's better if we can add classes to the elements that have these styles instead of changing css in javascript

src/javascript/app/base/footer.js
```diff
@@ -27,6 +28,42 @@ const Footer = (() => {
...
+
+        $dialog_notification.slideUp(200);
+        adjustElevioAndScrollup(8, 18);
+    };
+
+    const displayDialogMessage = () => {
+        BinarySocket.wait('website_status', 'authorize', 'landing_company').then(() => {
+            if (isEuCountry()) {
+                const $dialog_notification = $('#dialog_notification');
+                const $dialog_notification_accept = $('#dialog_notification_accept');
```
negarn: can we use javascript for this since it doesn't require any jquery animations?

src/javascript/app/base/footer.js
```diff
@@ -27,6 +28,42 @@ const Footer = (() => {
...
+                adjustElevioAndScrollup($dialog_notification.height() + 30, $dialog_notification.height() + 40);
+                $dialog_notification_accept
+                    .off('click')
+                    .on('click', () => {
+                        adjustElevioAndScrollup(8, 18);
+                        $dialog_notification.slideUp(200);
+                        Cookies.set('CookieConsent', 1);
+                    });
+                $(window).resize(() => {
+                    adjustElevioAndScrollup($dialog_notification.height() + 30, $dialog_notification.height() + 40);
```
negarn: perhaps we can store 30 and 40 in variables that make it clear what they are?

src/javascript/app/base/footer.js
cakasuma: you mean using document.geteElementById('###') ?

src/templates/app/_layout/footer.jsx
```diff
@@ -30,6 +30,13 @@ const StatusNotification = () => (
     </div>
 );
 
+const DialogNotification = () => (
+    <div id='dialog_notification' className='primary-bg-color'>
+        <p id='dialog_notification_text'>{it.L('Our website uses cookies to give you the best user experience. For more information, [_1].', `<a href="${it.url_for('terms-and-conditions')}?anchor=cookies-and-device-information#privacy" target="_blank" rel="noopener noreferrer">${it.L('view our policy')}</a>`)}</p>
```
ashkanx: @cakasuma the link `view our policy`  should get underlined.

src/templates/static/_layout/footer.jsx
```diff
@@ -30,6 +30,13 @@ const StatusNotification = () => (
     </div>
 );
 
+const DialogNotification = () => (
+    <div id='dialog_notification' className='primary-bg-color'>
+        <p id='dialog_notification_text'>{it.L('Our website uses cookies to give you the best user experience. For more information, [_1].', `<a href="${it.url_for('terms-and-conditions')}?anchor=cookies-and-device-information#privacy" target="_blank" rel="noopener noreferrer">${it.L('view our policy')}</a>`)}</p>
```
ashkanx: same here

PR #4914, opened 2019-01-23
[new-app] zaki/dropdown_update

zakibinary:
Changelog
- Refactored the dropdown component and stylesheet.
- Added `position` prop for left alignment of dropdown list.
- Disable toggling if only one option is available in dropdown.
- Layout updates by changing from `cssgrids` to `flexbox` to ensure trade parameter sidebar is fixed width, removed unused and redundant styles.


src/javascript/app_2/App/Components/Form/DropDown/dropdown.jsx
```diff
@@ -34,14 +30,33 @@ class Dropdown extends React.Component {
         document.removeEventListener('mousedown', this.handleClickOutside);
     }
 
-    handleSelect(item) {
+    isSingleOption = isArrayLike(this.props.list) ?
```
kellybinary: boolean should be underscore i think? `is_single_option`

zakibinary: good spot, missed this 👍

PR #4928, opened 2019-01-24
Update transfer between account disclaimer

msamprz:


src/templates/app/cashier/account_transfer.jsx
```diff
@@ -12,68 +12,72 @@ const Row = ({ id }) => (
...
-    </React.Fragment>
-);
+            <div className='hint invisible' id='transfer_info'>
+                <p>{it.L(`Note: ${account_transfer_note}`)}</p>
+            </div>
+
+            <div className='hint invisible' id='transfer_fee'>
+                {it.L('Notes:')}
+                <ul className='bullet'>
+                    <li>{it.L(account_transfer_note)}</li>
```
negarn: translation doesn't work this way, it needs to be wrapped around the string

msamprz: so just put the `account_transfer_note` in template strings? Like so? ```it.L(`${account_transfer_note}`)```

msamprz: Please note updated code.

negarn: looks fine now, thanks

src/templates/app/cashier/account_transfer.jsx
```diff
@@ -12,72 +12,64 @@ const Row = ({ id }) => (
...
+                <FormRow label={it.L('Transfer from')} type='label'  id='lbl_transfer_from' />
+                <FormRow label={it.L('Transfer to')}   type='select' id='transfer_to' />
+                <FormRow label={it.L('Amount')}        type='custom' id='transfer_amount'>
+                    <label id='currency' />
+                    <input id='amount' name='amount' type='text' maxLength='20' autoComplete='off' />
+                    <div className='hint' id='range_hint' />
+                </FormRow>
+            </Fieldset>
 
-            <div className='hint invisible' id='transfer_info'>
```
negarn: we can't remove old code, this needs to become visible in some cases, and the second longer list needs to become visible in other cases. you can have a look at the corresponding JS file and debug why in your branch it's showing both at the same time

PR #4929, opened 2019-01-24
Visibility filter for CFD, based on location (France)

msamprz:
+ Added visibility filter for CFD based on location (France)
+ Added `id` prop to `FillBox` element


src/javascript/app/common/country_base.js
```diff
@@ -18,9 +18,15 @@ const isEuCountry = () => {
     );
 };
 
+const isFrance    = () => {
```
negarn: better to make the naming more general in case we add more countries later, so something like hasCFD, and there's no need for the whitespace before `=`

src/javascript/app/common/country_base.js
```diff
@@ -18,9 +18,15 @@ const isEuCountry = () => {
...
+    const clients_country = Client.get('residence') || State.getResponse('website_status.clients_country');
+    return /^fr$/.test(clients_country);
+};
+
 const isIndonesia = () => State.getResponse('website_status.clients_country') === 'id';
 
 module.exports = {
     isEuCountry,
     isIndonesia,
+    isFrance,
```
negarn: since it's being place above isIndonesia, let's also maintain the same placement in the export

src/javascript/static/pages/regulation.js
```diff
@@ -21,6 +24,13 @@ const Regulation = (() => {
                     relocateLinks();
                 }
             });
+
+            BinarySocket.wait('website_status').then(() => {
+                if (isFrance()) {
+                    const $cfd_selector = $('#cfd_fillbox');
+                    $cfd_selector.remove();
```
negarn: better to use javascript if jquery is not necessary

src/javascript/app/common/country_base.js
msamprz: Addressed in updated code.

src/javascript/static/pages/regulation.js
msamprz: Got it. Fixed that and simplified in latest commit.

src/javascript/app/common/country_base.js
msamprz: Thanks. Addressed in latest commit.

src/javascript/static/pages/regulation.js
```diff
@@ -21,6 +24,12 @@ const Regulation = (() => {
                     relocateLinks();
                 }
             });
+
+            BinarySocket.wait('website_status').then(() => {
+                if (hasCFD()) {
+                    document.querySelectorAll('#cfd_fillbox').remove();
```
negarn: if it's id there should be only a single element, so just getElementByID works (note that we have a custom function for that)

msamprz: Addressed in [#611f3d3](https://github.com/binary-com/binary-static/pull/4929/commits/611f3d3cecf7c31483d6e0bdcddaf510a3729b3c)

msamprz: Thanks. Addressed in [#6341808](https://github.com/binary-com/binary-static/pull/4929/commits/634180883314fe4bcb83e7ed513053538fb2aaf3).

src/javascript/app_2/Stores/Modules/Trading/Helpers/duration.js
```diff
@@ -80,8 +80,8 @@ export const getExpiryType = (store) => {
     const server_time = store.root_store.common.server_time;
 
     const duration_is_day       = expiry_type === 'duration' && duration_unit === 'd';
-    const expiry_is_after_today = (expiry_type === 'endtime' && toMoment(expiry_date).isAfter(toMoment(server_time), 'day')) ||
-        !hasIntradayDurationUnit(duration_units_list);
+    const expiry_is_after_today = expiry_type === 'endtime' && (toMoment(expiry_date).isAfter(toMoment(server_time), 'day') ||
```
negarn: looks like an issue with resolving conflict, this shouldn't be a part of your PR

PR #4930, opened 2019-01-24
ahkuma/Only en

ahkuma:


src/javascript/app_2/Utils/Language/language.js
```diff
@@ -3,7 +3,22 @@ import { get, getAll, urlFor } from '_common/language';
...
+        'FR',
+        'ID',
+        'IT',
+        'PL',
+        'PT',
+        'RU',
+        'TH',
+        'VI',
+        'ZH_CN',
+        'ZH_TW',
```
negarn: instead of adding all here, I suggest just returning 'EN' in language_list, because if API starts sending a new language we have to update this list but if we only return 'EN' as the allowed language no such update is needed

src/templates/app_2/head.jsx
```diff
@@ -66,6 +66,15 @@ const Head = () => (
             <script key={inx} src={js_file.replace('{PLACEHOLDER_FOR_LANG}', it.language.toLowerCase())} defer />
         ))}
         <script type='text/javascript' src={`${it.root_url}pushwoosh-web-notifications.js`} defer />
+        { it.language.toLowerCase() !== 'en' &&
+            // TODO remove this script and the inline if wrapping this when the design is ready.
+            <script
+                type='text/javascript'
+                dangerouslySetInnerHTML={{ __html: `
+                    (window.location.href = window.location.href.replace(/\\/app\\/(\\w{2}|\\w{2}_\\w{2})\\//, '/app/en/'))
+                ` }}
```
negarn: I suggest we handle this in 404.html page instead, in `redirectToReactRoute` function, instead of getting the language we default it to english so when path is being built it builds it to en instead of having an extra redirect, @morteza-binary what do you think? would that work?

src/javascript/app_2/Utils/Language/language.js
ahkuma: much simpler, I agree.

PR #4931, opened 2019-01-24
easteregg/change_transfer_between_accounts_limit

easteregg:


src/javascript/_common/base/currency_base.js
```diff
@@ -109,6 +109,11 @@ const getCurrencyName = currency => getPropertyValue(CryptoConfig.get(), [curren
 
 const getMinPayout = currency => getPropertyValue(currencies_config, [currency, 'stake_default']);
 
+const getMaxTransfer = currency => {
+    const max_transfer = getPropertyValue(currencies_config, [currency, 'transfer_between_accounts', 'limits', 'max']);
```
negarn: We already have this function `getMinTransfer`, I suggest grouping them together in one function such as `getTransferLimits(value, currency)` and pass in either `'min'` or `'max'` to get that instead of repeating logic in two functions

negarn: then this would change to 
```
const transfer_limit = getPropertyValue(currencies_config, [currency, 'transfer_between_accounts', 'limits', value]);
```

src/javascript/app/pages/cashier/account_transfer.js
```diff
@@ -215,12 +219,30 @@ const AccountTransfer = (() => {
...
                         getElementById(messages.parent).setVisibility(1);
                         return;
                     }
-                    getElementById('range_hint').textContent = `${localize('Min')}: ${min_amount} ${localize('Max')}: ${client_balance <= withdrawal_limit ? localize('Current balance') : localize('Withdrawal limit')}`;
+                    max_amount = Currency.getMaxTransfer(
+                        Client.get('currency')
+                    );
+                    transferable_amount = max_amount ?
+                        Math.min(max_amount, withdrawal_limit, client_balance) :
+                        Math.min(withdrawal_limit, client_balance);
```
negarn: regarding having a fallback value, I had a discussion with Farzaneh and she said that if API does not return rates to us we should have the account transfer form disabled and should show a message that rates are not available at this time and to try again later. can you check with her on that?

src/javascript/app/pages/cashier/account_transfer.js
```diff
@@ -215,12 +219,30 @@ const AccountTransfer = (() => {
...
+
+                    getElementById('range_hint').textContent = `${localize('Min')}: ${min_amount} ${localize('Max')}: ${transferable_amount.toFixed(Currency.getDecimalPlaces(Client.get('currency')))}`;
+
+                    el_explain_dynamic_limits.innerHTML = localize(
```
negarn: why does this need to be added with javascript?

src/javascript/app/pages/cashier/account_transfer.js
```diff
@@ -215,12 +219,30 @@ const AccountTransfer = (() => {
...
                     }
-                    getElementById('range_hint').textContent = `${localize('Min')}: ${min_amount} ${localize('Max')}: ${client_balance <= withdrawal_limit ? localize('Current balance') : localize('Withdrawal limit')}`;
+                    max_amount = Currency.getMaxTransfer(
+                        Client.get('currency')
+                    );
+                    transferable_amount = max_amount ?
+                        Math.min(max_amount, withdrawal_limit, client_balance) :
+                        Math.min(withdrawal_limit, client_balance);
+
+                    getElementById('range_hint').textContent = `${localize('Min')}: ${min_amount} ${localize('Max')}: ${transferable_amount.toFixed(Currency.getDecimalPlaces(Client.get('currency')))}`;
```
negarn: same here, can min and max be set in the template and we just add their values via javascript?

src/javascript/app/pages/cashier/account_transfer.js
easteregg: We tested this together. right now, the only way to reproduce it is to stop exchange feed. in which it causes `get_limits` API to throw an `InternalServerError` and we don't show the form at all. So I thought, what if this doesn't happen in production? if `get_limits` API does not throw an internal server error, then the form will break. That was the reason I added the fallback value just to be on the safe side.

src/javascript/app/pages/cashier/account_transfer.js
easteregg: Since the text is created dynamically, I thought this is the better way to do it. is there a better alternative that I am not aware of?

src/javascript/app/pages/cashier/account_transfer.js
easteregg: Ah, I see what you mean, so something like: 
```
<div id=range_hint>
   {it.L('Min:'}<span id='account_transfer_min_value'></span>
   {it.L('Max:'}<span id='account_transfer_max_value'></span>
</div>
```
?

src/javascript/_common/base/currency_base.js
easteregg: I first created the function `getMinMaxTransferLimits`, and then I thought not to complicate the function API and use another function to minify the changes. 

I'll go with your suggestion. fine by me :)

src/javascript/app/pages/cashier/account_transfer.js
negarn: sure, let's go with this then

src/javascript/app/pages/cashier/account_transfer.js
negarn: that's right

src/javascript/_common/base/currency_base.js
```diff
@@ -79,11 +79,24 @@ const CryptoConfig = (() => {
...
+ * @returns {min: numeric, ?max: numeric}
+ */
+const getTransferLimits = (currency, which) => {
+    const transfer_limits = getPropertyValue(currencies_config, [currency, 'transfer_between_accounts', 'limits']) || getMinWithdrawal(currency);
     const decimals     = getDecimalPlaces(currency);
-    return min_transfer.toFixed(decimals); // we need toFixed() so that it doesn't display in scientific notation, e.g. 1e-8 for currencies with 8 decimal places
+
+    switch (which) {
+        case 'max':
+            return transfer_limits.max.toFixed(decimals) || undefined;
```
negarn: in case that we fall back to `getMinWithdrawal(currency)` for `transfer_limits` variable, we won't have `transfer_limits.max` so will `transfer_limits.max.toFixed(decimals)` throw an error?

negarn: same for below cases, getMinWithdrawal doesn't return an object with `min` key, so it's better to handle that

src/javascript/app/pages/cashier/account_transfer.js
```diff
@@ -225,30 +223,52 @@ const AccountTransfer = (() => {
...
                     );
+                    populateHints();
                     populateAccounts(accounts);
                 });
             }
         });
     };
 
+    const populateHints = () => {
+        getElementById('limit_current_balance').innerHTML  = Currency.formatMoney(
```
negarn: `textContent` isn't enough? we need `innerHTML`?

src/javascript/_common/base/currency_base.js
easteregg: Good point. will do.

src/javascript/app/pages/cashier/account_transfer.js
easteregg: formatMoney produces HTML, so it'll get escaped.

src/javascript/_common/base/currency_base.js
```diff
@@ -78,11 +78,24 @@ const CryptoConfig = (() => {
...
+const getTransferLimits = (currency, which) => {
+    const transfer_limits = getPropertyValue(currencies_config, [currency, 'transfer_between_accounts', 'limits']) || getMinWithdrawal(currency);
     const decimals     = getDecimalPlaces(currency);
-    return min_transfer.toFixed(decimals); // we need toFixed() so that it doesn't display in scientific notation, e.g. 1e-8 for currencies with 8 decimal places
+
+    switch (which) {
+        case 'max':
+            return transfer_limits.max ? transfer_limits.max.toFixed(decimals) : undefined;
+        case 'all':
+            return transfer_limits;
```
negarn: nitpicking about the fallback option again, if we fallback to getMinWithdrawal for `transfer_limits` we won't have and object with min and max, but otherwise we will. I suggest removing the `all` case altogether since it's not used anywhere, or handling the fallback better if you want to keep it.

easteregg: I want to save the behaviour of giving out both of the limits for future. So I have added a better failsafe for it.

src/sass/app/components/account_transfer.scss
```diff
@@ -0,0 +1,93 @@
+/* stylelint-disable color-no-hex */
+$TRANSFER_BACKGROUND_LIGHT_GRAY: #f2f2f2;
+$TRANSFER_BACKGROUND_LIGHT_GRAY_BORDER: #dedede;
+$TRANSFER_COLOR_BRAND_ORANGE: #e98024;
```
kellybinary: i think these has been defined as `$COLOR_LIGHT_GRAY`, `$COLOR_GRAY`,` $COLOR_ORANGE` respectively

src/javascript/_common/base/currency_base.js
```diff
@@ -79,11 +79,20 @@ const CryptoConfig = (() => {
...
 
-// returns in a string format, e.g. '0.00000001'
-const getMinTransfer = currency => {
-    const min_transfer = getPropertyValue(currencies_config, [currency, 'transfer_between_accounts', 'limits', 'min']) || getMinWithdrawal(currency);
-    const decimals     = getDecimalPlaces(currency);
-    return min_transfer.toFixed(decimals); // we need toFixed() so that it doesn't display in scientific notation, e.g. 1e-8 for currencies with 8 decimal places
+/**
+ * Returns the transfer limits for the account.
+ * @param currency
+ * @param which
```
negarn: perhaps explain the possible values to pass to which?

src/javascript/_common/base/currency_base.js
```diff
@@ -79,11 +79,20 @@ const CryptoConfig = (() => {
...
+ * @returns numeric
```
negarn: seems like sometimes object is returned (e.g. when which is missing) and sometimes undefined is returned. can we really say it returns numeric then?

src/javascript/_common/base/currency_base.js
```diff
@@ -79,11 +79,20 @@ const CryptoConfig = (() => {
...
+ */
+const getTransferLimits = (currency, which) => {
+    const transfer_limits = getPropertyValue(currencies_config, [currency, 'transfer_between_accounts', 'limits']) || getMinWithdrawal(currency);
+    const decimals        = getDecimalPlaces(currency);
+    if (which === 'max') {
+        return transfer_limits.max ? transfer_limits.max.toFixed(decimals) : undefined;
+    }
+    
+    return transfer_limits.min ? transfer_limits.min.toFixed(decimals) : transfer_limits;
```
negarn: can you add a test for this function?

src/javascript/_common/base/currency_base.js
easteregg: Good point. I'll add a check to see if no `which` is passed and `transfer_limits.min` is also undefined,  throw error.

src/javascript/_common/base/currency_base.js
easteregg: Yes. will do.

PR #4934, opened 2019-01-25
amam/fix_tooltip_equals

cakasuma:
alter design tooltip for allow equals following new design

-- altering the color to fix same variables `$COLOR_GRAY`


src/sass/app_2/_common/base/constants.scss
```diff
@@ -26,7 +26,7 @@ $COLOR_LIGHT_BLACK  : #333;
 $COLOR_BLACK        : #000;
 
 $COLOR_GREEN        : #4bbfab;
-$COLOR_GRAY         : #7f8397;
```
OskarAhl: Isn't `$COLOR_GRAY` used in a lot of places throughout app2?

cakasuma: there are two $COLOR_GRAY's actually, the duplicated added from one of the commit with master

PR #4935, opened 2019-01-25
ahkuma/password_reset page update

ahkuma:


src/templates/app/user/reset_password.jsx
```diff
@@ -15,10 +15,10 @@ const ResetPassword = () => (
             <div className='gr-parent gr-padding-10' id='container_reset_password'>
                 <form id='frm_reset_password'>
                     <Fieldset legend={it.L('Details')}>
+                        <FormRow type='radio' id='have_real_account' options={ [{ value: 0, label: 'Yes' }, { value: 1, label: 'No' }] } default_option={1} label={it.L('Do you have a real money account?')} />
```
negarn: a bit strange that 0 means yes and 1 means no, can we use string values etc?

src/sass/static/pages.scss
```diff
@@ -1480,3 +1480,17 @@ body #not_authenticated_financial ul.checked > li {
         color: $COLOR_BLUE;
     }
 }
+
+#have_real_account {
```
negarn: should be in websocket pages scss since it is not for a static page

PR #4953, opened 2019-01-29
kelly/disable_weekends_and holidays

kellybinary:
- disable weekends and holidays on trading datepicker
- show Tooltip for Holidays
- fix animation to use React CSSTransitions
- minor fix on colors


src/javascript/app_2/App/Components/Form/DatePicker/date_picker.jsx
```diff
@@ -197,18 +230,32 @@ class DatePicker extends React.Component {
...
-                    className={classNames('datepicker__picker', {
-                        'datepicker__picker--show'                           : this.state.is_datepicker_visible,
-                        [`datepicker__picker--align-${this.props.alignment}`]: this.props.alignment,
-                    })}
+                <CSSTransition
+                    in={this.state.is_datepicker_visible}
+                    timeout={100}
+                    classNames={{
+                        enter    : 'datepicker__picker--enter',
+                        enterDone: 'datepicker__picker--enter--done',
```
zakibinary: Maybe `enterDone` modifier class should called `--enter-done` instead? It's actually a separate modifier `class` from `--enter`.

kellybinary: ok, sure

src/javascript/app_2/App/Components/Form/DatePicker/date_picker.jsx
```diff
@@ -113,6 +124,28 @@ class DatePicker extends React.Component {
         }
     };
 
+    async onChangeCalendarMonth(calendar_date) {
+        const trading_events = await getTradingEvents(calendar_date, this.props.underlying);
+        const holidays = [];
+        let weekends   = [];
+        trading_events.filter(events => {
```
OskarAhl: How come you are using `.filter` here? I think e.g. foreach is more appropriate since filter returns a new filtered array

kellybinary: good catch, thanks :)

PR #4966, opened 2019-01-30
Amam/barrier_input

cakasuma:
add styling design barrier input for barrier contract


src/javascript/app_2/Modules/Trading/Components/Form/TradeParams/barrier.jsx
```diff
@@ -12,6 +13,8 @@ const Barrier = ({
     onChange,
     validation_errors,
 }) =>  {
+    const barrier_title = barrier_count === 2 ? localize('Barriers') : localize('Barrier');
```
negarn: why not the other way? if it's 1 go to barrier, if it's more go to plural

src/javascript/app_2/Modules/Trading/Components/Form/TradeParams/barrier.jsx
```diff
@@ -35,7 +38,7 @@ const Barrier = ({
         );
     }
     return (
-        <Fieldset>
+        <Fieldset className='barriers' header={barrier_title}>
```
morteza-binary: @cakasuma would you please add each attribute in a separate line?

src/sass/app_2/modules/trading.scss
```diff
@@ -258,6 +258,48 @@
...
+                    padding-left: 3px;
+                }
+                &.multiple {
+                    display: inline-block;
+                    width: 46%;
+                    margin-left: 4px;
+                    margin-right: 4px;
+
+                    .up {
+                        mask: url('../images/sidebar/ic_barrier_high.svg');
```
zakibinary: Is mask supported widely by other browsers apart from Chromium based ones? Had a look at caniuse.com and it seems most browsers only have partial support.

src/sass/app_2/modules/trading.scss
```diff
@@ -258,6 +258,48 @@
...
+                    left: 0 !important;
+                }
+            }
+            .input-field {
+                input {
+                    padding-left: 3px;
+                }
+                &.multiple {
+                    display: inline-block;
+                    width: 46%;
```
zakibinary: Maybe we would be better off using something like `flexbox` or `cssgrids` since we just want a two column design with even widths.

src/sass/app_2/_common/form/input_field.scss
```diff
@@ -156,7 +157,8 @@
         &:active, &:focus {
             outline: 0;
             border-radius: $BORDER_RADIUS;
-            border-color: $COLOR_BLUE;
+            border-color: $COLOR_WHITE;
+            box-shadow: 0 0 0 0.2rem rgba(127, 131, 151, 0.3) // stylelint-disable-line number-leading-zero
```
zakibinary: Do we still need the `stylelint-disable-line` for this rule? For `number-leading-zero` seems like there are no rules violations here. It's only a violation if you use `.3` instead of `0.3`.

src/sass/app_2/_common/components/contracts_popup.scss
```diff
@@ -4,7 +4,7 @@ div.contracts-popup-container {
     background: $COLOR_WHITE;
     border: solid 1px $COLOR_GRAY;
     border-radius: 4px;
-    transition: transform 0.3s opacity 0.3s;
+    transition: transform 0.3s, opacity 0.3s, box-shadow 0.18s;
```
zakibinary: Personally I think we rather not use `transition` on `box-shadow` since it re-paints on every frame, having a performance impact on the page.

src/sass/app_2/modules/trading.scss
```diff
@@ -632,4 +674,18 @@
             }
         }
     }
+    fieldset {
+        &.barriers {
+            .input-field {
+                &.multiple {
+                    .up {
+                        background-color: $COLOR_WHITE;
```
zakibinary: You can combine `up` and `down` classes here since they're both the same `background-color`.

src/sass/app_2/_common/form/input_field.scss
cakasuma: its because the box-shadow start with 0 from `0 0 0 0.2rem`

src/sass/app_2/modules/trading.scss
cakasuma: most browsers supported but the -webkit-mask actually, and mask is directly translated into -webkit-mask, but despite that, IE not supported at all. do we need to support both IE and edge, or edge only ? (edge is actually supported)

src/sass/app_2/_common/form/input_field.scss
zakibinary: That's strange, does it not pass the rule if you remove it?

src/sass/app_2/modules/trading.scss
cakasuma: i think its better to make the image into accessible svg and avoid `mask` . i also think thats its quite dangerous in production

zakibinary: We can use `mask` if there's some special `clipping` or `filter` effects that we need, but in this case, we're just rendering an `svg` icon. 

So I think `mask` is a tad overkill, especially for something that is only partially supported in most modern browsers.

src/javascript/app_2/Modules/Trading/Components/Form/TradeParams/barrier.jsx
```diff
@@ -1,8 +1,11 @@
-import { observer } from 'mobx-react';
-import PropTypes    from 'prop-types';
-import React        from 'react';
-import Fieldset     from 'App/Components/Form/fieldset.jsx';
-import InputField   from 'App/Components/Form/input_field.jsx';
+import { observer }        from 'mobx-react';
+import PropTypes           from 'prop-types';
+import React               from 'react';
+import { IconBarrierUp }   from 'Assets/Trading/Barriers/icon_barrier_up.jsx';
+import { IconBarrierDown } from 'Assets/Trading/Barriers/icon_barrier_down.jsx';
```
kellybinary: these two lines can be imported together as such: 
`import { IconBarrierUp, IconBarrierDown }`, but please add an `index.js` file in Assets/Trading/Barriers and export those two files

src/javascript/app_2/Modules/Trading/Components/Form/TradeParams/barrier.jsx
```diff
@@ -35,28 +40,39 @@ const Barrier = ({
...
                 <InputField
                     type='number'
-                    name='barrier_2'
-                    value={barrier_2}
+                    name='barrier_1'
+                    value={barrier_1}
+                    className={barrier_count === 2 ? 'multiple' : 'single'}
                     onChange={onChange}
-                    error_messages = {validation_errors.barrier_2}
+                    error_messages = {validation_errors.barrier_1 || []}
```
kellybinary: could you remove the spaces between `=`? :)

src/javascript/app_2/Modules/Trading/Components/Form/TradeParams/barrier.jsx
```diff
@@ -35,28 +40,39 @@ const Barrier = ({
...
+    
+                {barrier_count === 2 &&
+                    <React.Fragment>
+                        <InputField
+                            type='number'
+                            name='barrier_2'
+                            value={barrier_2}
+                            className='multiple'
+                            onChange={onChange}
+                            error_messages = {validation_errors.barrier_2}
```
kellybinary: here as well, please remove the extra spaces

src/javascript/app_2/App/Components/Form/input_field.jsx
```diff
@@ -43,8 +43,8 @@ const InputField = ({
 
     const changeValue = (e) => {
         if (type === 'number') {
-            const is_empty = !e.target.value || e.target.value === '';
-            const signed_regex = is_signed ? '^[\+\-]?[0-9]*(\.?([0-9]*)?)?$' : '^';
+            const is_empty = !e.target.value || e.target.value === '' || e.target.value === '  ';
+            const signed_regex = is_signed ? '[\+\-\.0-9]$' : '^';
```
zakibinary: For `barrier` we need to check if value is leading with `+` or `-`. If you remove this it will falsely pass the validation in place for the `input_field` regex and return out of range error from API.

PR #4969, opened 2019-01-30
easteregg/fix-trackjs-offset

easteregg:


src/javascript/app/pages/trade/digit_ticker.js
```diff
@@ -139,7 +139,7 @@ const DigitTicker = (() => {
 
     // Calculate peek-box left offset.
     const calculateOffset = () => {
-        const left_offset = document.querySelector(`.digit-${current_spot}`).offsetLeft;
+        const left_offset = document.querySelector(`.digit-${current_spot}`) ? document.querySelector(`.digit-${current_spot}`).offsetLeft : 0;
```
negarn: why not store it in a variable instead of fetching it twice?

easteregg: You are right. pushing a commit now.

PR #4975, opened 2019-01-31
Nazanin/trade_type_info_height

nazaninreihani:


src/sass/app_2/_common/components/contracts_popup.scss
```diff
@@ -270,18 +270,25 @@ div.contracts-popup-container {
             }
             .info-content {
                 padding: 8px;
+                height: 215px;
+                overflow-y: scroll;
```
kellybinary: better to use `overflow-y: auto;`

PR #4979, opened 2019-02-03
easteregg/fix-one-tick-issues

easteregg:


src/javascript/app/pages/trade/purchase.js
```diff
@@ -327,6 +331,12 @@ const Purchase = (() => {
 
             if (CommonFunctions.isVisible(spots) && tick_d.epoch && tick_d.epoch > purchase_data.buy.start_time) {
                 const current_tick_count = spots.getElementsByClassName('row').length + 1;
+                if (contract_duration && +contract_duration < current_tick_count) {
+                    sellExpired();
+                    duration = 0;
+                    console.log('Fixed', this); // eslint-disable-line
```
negarn: do you want to be committing this?

src/javascript/app/pages/trade/purchase.js
```diff
@@ -327,6 +331,12 @@ const Purchase = (() => {
...
+                    break;
```
negarn: did you mean `return`?

src/javascript/app/pages/trade/purchase.js
easteregg: removed for production.

PR #4980, opened 2019-02-04
allow allow equal validation

cakasuma:
fix allow equal unable to check

### DIFF
--- a/src/javascript/app_2/App/Components/Form/input_field.jsx
+++ b/src/javascript/app_2/App/Components/Form/input_field.jsx
@@ -48,7 +48,7 @@ const InputField = ({
             e.target.value = e.target.value.replace(unit, '').trim();
         }
 
-        if (e.target.value === value) {
+        if (e.target.value === value && e.target.value !== 'is_equal') {
             return;
         }
 


src/javascript/app_2/App/Components/Form/input_field.jsx
```diff
@@ -48,7 +48,7 @@ const InputField = ({
             e.target.value = e.target.value.replace(unit, '').trim();
         }
 
-        if (e.target.value === value) {
+        if (e.target.value === value && e.target.value !== 'is_equal') {
```
negarn: wouldn't it be better to check that field is not a checkbox instead?

src/javascript/app_2/Stores/Modules/Trading/trade_store.js
```diff
@@ -448,6 +448,12 @@ export default class TradeStore extends BaseStore {
 
     @action.bound
     changeAllowEquals() {
+        if (!this.contract_types_list
+            || !this.contract_type
+            || !this.duration_unit
+            || !this.contract_start_type
+            || !this.expiry_type) return;
```
OskarAhl: I think it's cleaner if you add these checks to the functions where these variables are used?

PR #4981, opened 2019-02-04
msamprz/link-to-kid

msamprz:


src/templates/static/responsible_trading.jsx
```diff
@@ -41,7 +41,7 @@ const ResponsibleTrading = () => (
...
+            <p>{it.L('Investing in complex products, such as CFDs and FX may incur losses as well as gains. Prices may vary and/or fluctuate due to changes in current market prices and conditions, which may impact the return on your investment. Before making an Investment decision, you should refer to our <a href=\'[_1]\'>Key Information Documents</a> on our website, in particular, the amount of Margin required for particular instruments that we offer. The Products offered by Binary Investments (Europe) Ltd fall under the category of ‘complex products’ and may not be suitable for retail clients.', it.url_for('regulation'))}</p>
```
negarn: it's better to put the link in placeholders so translators don't have to deal with them

msamprz: noted and addressed.

src/templates/static/responsible_trading.jsx
```diff
@@ -41,7 +41,7 @@ const ResponsibleTrading = () => (
...
+            <p>{it.L('Investing in complex products, such as CFDs and FX may incur losses as well as gains. Prices may vary and/or fluctuate due to changes in current market prices and conditions, which may impact the return on your investment. Before making an Investment decision, you should refer to our [_1]Key Information Documents[_2] on our website, in particular, the amount of Margin required for particular instruments that we offer. The Products offered by Binary Investments (Europe) Ltd fall under the category of ‘complex products’ and may not be suitable for retail clients.', `<a href=\'${it.url_for('regulation')}\'>`,'</a>')}</p>
```
kellybinary: can we add anchor to KID tab? e.g.`url_for('regulation.html#key_information_documents')`

msamprz: sure, addressed in latest commit.

src/templates/static/legal/regulation.jsx
```diff
@@ -150,7 +150,7 @@ const Regulation = () => {
                         />
                     </div>
 
-                    <h3 id='key_information_documents'>{it.L('Key Information Documents')}</h3>
+                    <h3 id='key_information_documents' href='#key_information_documents'>{it.L('Key Information Documents')}</h3>
```
nazaninreihani: why using href? isn't id the way to do it?

msamprz: @nazaninreihani ID does work, and initially it was just with ID and it worked, but if the user is getting redirected from `responsible-trading` for the first time (as in they've never browsed that page again), it doesnt scroll exactly to KID, and after inspecting, I noticed that it scrolls to the beginning of `#accordion`, so in this case for the first time, accordion overrides browser's scroll and scrolls itself, and accordion uses the `href` attribute as selector.

msamprz: Although after thorough testing, i now find that this did not fix the issue, and the issue still persists, but it's seemingly very random. I've tested it a total of 20 times under different configurations (with `href`, without `href`, with `navigation` property on `accordion` init in `regulation.js`, and without) and the issue still only sometimes appears under each of those configurations including combinations of them. Do you have any suggestions for me, @nazaninreihani ?

nazaninreihani: @msamprz have you tried adding id attribute to other matching siblings? 
all h3 of tabs, not just this one

msamprz: @nazaninreihani I tried it now, and out of 15 times, at least 4 times it failed. But still majority of the times it was fine. I may have picked up something that the issue happens when I quickly click on the KID hyperlink (as in not letting the page load fully, or something else).

nazaninreihani: How fast you can click :-D? I removed all contents from responsible trading except the link and clicked on it even before it appeared on screen and it worked

nazaninreihani: when issue happens is url containing the hash?

msamprz: Maybe it helps that my computer is slow 😄, no but it may not be about speed then, I just thought I noticed something like that.

And yes in all circumstances the URL contains the hash. The issue isn't that the scroll or something doesn't work, the issue is that it scrolls to `EU Passport Rights` instead of KID.

PR #4984, opened 2019-02-04
msamprz/topup-popup-race

msamprz:


src/javascript/app/pages/user/update_balance.js
```diff
@@ -22,7 +22,11 @@ const updateBalance = (response) => {
         updateContractBalance(balance);
         $('.topMenuBalance, .binary-balance').html(view)
             .css('visibility', 'visible');
-        TopUpVirtualPopup.init(balance);
+        BinarySocket.send({ portfolio: 1 }).then((res) => {
```
negarn: what issue is this solving? I would rather not add a flood of portfolio calls every time balance updates

msamprz: As per discussions, this is noted and addressed.

PR #4985, opened 2019-02-04
easteregg/subscribe_on_websocket_reconnect

easteregg:


src/javascript/_common/base/socket_base.js
```diff
@@ -289,6 +298,8 @@ const BinarySocketBase = (() => {
         sendBuffered      : sendBufferedRequests,
         get               : () => binary_socket,
         setOnDisconnect   : (onDisconnect) => { config.onDisconnect = onDisconnect; },
+        setOnReconnect    : (onReconnect) => {config.onReconnect = onReconnect; },
```
negarn: space after `{`

src/javascript/_common/base/socket_base.js
```diff
@@ -289,6 +298,8 @@ const BinarySocketBase = (() => {
...
+        removeOnReconnect : () => { delete config.onReconnect; },
```
negarn: why not combine it with removeOnDisconnect? if we are using both in same places

src/javascript/app/pages/user/account/portfolio/portfolio.init.js
```diff
@@ -220,7 +220,20 @@ const PortfolioInit = (() => {
         is_initialized = false;
     };
 
+    const onReconnect = () => {
```
negarn: is it different from calling onunload and then load? do we need to set a custom onreconnect for every page?

src/javascript/_common/base/socket_base.js
easteregg: We need to do it on `onopen` of the reconnection. onDisconnect only happens when it is disconnected. and in that time, I don't think it'll help if we send the request.

src/javascript/app/pages/user/account/portfolio/portfolio.init.js
easteregg: There is no requirement for `onReconnect` on every page. If we need a page to do certain things when connection is back up, then we use this. if not, no command will be running.

src/javascript/app/base/page.js
```diff
@@ -54,7 +54,7 @@ const Page = (() => {
                 switch (evt.key) {
                     case 'active_loginid':
                         // not the active tab and logged out or loginid switch
-                        if (document.hidden && (evt.newValue === '' || !window.is_logging_in)) {
+                        if (evt.newValue === '' || !window.is_logging_in) {
```
negarn: I guess the comment now has to be updated as well? why was the code there? does removing it break any other behaviour?

easteregg: That is a good question. I thought about that, why we added that in a first place.. will follow this up with you on slack.

PR #4986, opened 2019-02-04
msamprz/elevio-lang-fix

msamprz:


src/javascript/_common/base/elevio.js
```diff
@@ -1,18 +1,20 @@
...
+const localize        = require('../localize').localize;
+const createElement   = require('../utility').createElement;
 
 const Elevio = (() => {
     const init = () => {
         if (!window._elev) return; // eslint-disable-line no-underscore-dangle
         window._elev.on('load', (elev) => { // eslint-disable-line no-underscore-dangle
-            const available_elev_languages = ['es', 'id', 'pt', 'ru'];
-            const current_language         = getLanguage().toLowerCase();
+            const available_elev_languages = ['de', 'fr', 'id', 'it', 'po', 'pt', 'ru', 'vi'];
```
negarn: we shouldn't be adding more languages, our custom translation is not fully ready from elevio

src/javascript/_common/base/elevio.js
```diff
@@ -1,18 +1,20 @@
...
+            const current_language         = languageFromUrl().toLowerCase();
```
negarn: getLanguage also calls languageFromUrl, what was the issue that this is solving?

msamprz: noted and addressed.

src/javascript/_common/base/elevio.js
msamprz: noted and addressed.

PR #4989, opened 2019-02-05
update colours

negarn:


src/sass/app_2/_common/components/range_slider.scss
```diff
@@ -142,57 +142,62 @@
...
     &__caption {
         display: flex;
         padding: 4px 0;
         justify-content: space-between;
+        line-height: 18px;
     }
 }
 
+.sidebar-container fieldset div.range-slider__caption span {
+    font-size: 0.75rem;
```
zakibinary: Will this rule not work if you move `span` inside the `&__caption` block above?

negarn: no, it gets over written, so I have to either add !important or override it with specificity

PR #4994, opened 2019-02-06
Nazanin/financial_assessment_null

nazaninreihani:


src/javascript/app/pages/user/account/settings/financial_assessment.js
```diff
@@ -90,7 +90,9 @@ const FinancialAssessment = (() => {
             const data = { set_financial_assessment: 1 };
             showLoadingImage(getElementById('msg_form'));
             $(form_selector).find('select').each(function () {
-                financial_assessment[$(this).attr('id')] = data[$(this).attr('id')] = $(this).val() || null;
+                if ($(this).val().length) {
```
negarn: I suggest storing `const $this = $(this);`,  `const value = $this.val();`, and also `const id = $this.attr('id');` since all are reused

PR #5001, opened 2019-02-08
oskar/store_name

OskarAhl:
* Add explicit store names to ui_store and trade_store to prevent uglifyjs minification to name the stores the same.


src/javascript/app_2/Stores/base_store.js
```diff
@@ -62,6 +64,11 @@ export default class BaseStore {
             enumerable: false,
             writable  : true,
         });
+        Object.defineProperty(this, 'store_name', {
+            value     : store_name,
```
mory0tiki: @OskarAhl I'd be better to throw an exception here when the `sotre_name` doesn't have a valid value since the `null` or invalid value affects the functionality of the `BaseStore` to load and save the snapshot to the storage. So it's better to prevent developers to pass null when they want to use the autosave feature of the base_store.

e.g. you can throw an exception when the `store_name` is null or invalid but one or both of `local_storage_properties` and 'session_storage_properties` have value.

src/javascript/app_2/Stores/base_store.js
```diff
@@ -139,9 +146,9 @@ export default class BaseStore {
         });
 
         if (storage === BaseStore.STORAGES.LOCAL_STORAGE) {
-            localStorage.setItem(this.constructor.name, snapshot);
+            localStorage.setItem(this.store_name ? this.store_name : this.constructor.name, snapshot);
```
mory0tiki: @OskarAhl  if you check the validity of `store_name` in the constructor it doesn't need to check it here anymore.

src/javascript/app_2/Stores/base_store.js
```diff
@@ -63,6 +65,21 @@ export default class BaseStore {
             writable  : true,
         });
 
+        const has_local_or_session_storage = local_storage_properties && local_storage_properties.length
+            || session_storage_properties && session_storage_properties.length;
+
+        if (has_local_or_session_storage) {
+            if (!store_name) {
```
OskarAhl: @mory0tiki would this be fine?

src/javascript/app_2/Stores/base_store.js
```diff
@@ -63,6 +65,21 @@ export default class BaseStore {
...
+                throw new Error('store_name is required for local/session storage');
+            } else {
```
mory0tiki: @OskarAhl it doesn't need `else` because after throwing an error the code will no execute anymore.

PR #5002, opened 2019-02-08
[new-app] zaki/font_smoothing_and_fixes

zakibinary:
Changelog
1. Added `font-smoothing` rules that should apply across supported browsers.
2. Updates to `margin` for `chartContainer` and `sideBar` and `positions` to follow latest design.
3. Color updates/fixes for certain components based on latest design.
4. Add `custom-scrollbars` to `app` contents and to `trade-categories` explanation when overflow is possible.
5. Fix contracts rendered out of viewport when in `contract` mode.
6. Style updates to `positions` drawer toggling.
7. Fix range slider `tick` steps moving out of position when window is scaled below `980px`.


src/sass/app_2/_common/base/typography.scss
```diff
@@ -28,6 +28,13 @@
     @return calc(#{$a*100}vw #{$sign} #{$b});
 }
 // Indrek Pass also has a fantastic mixin here: http://sassmeister.com/gist/7f22e44ace49b5124eec
+html {
+    text-shadow: 1px 1px 1px rgba(0, 0, 0, 0.004); //stylelint-disable-line number-max-precision
+    text-rendering: optimizeLegibility !important;
```
negarn: does this mess with translations? I remember something similar messing up some special characters in some other languages

zakibinary: We have the same `optimizeLegibility` rule in `binary-style` typography as well. 
It shouldn't affect translations. Same goes with `text-shadow`.

PR #5003, opened 2019-02-08
negar/style_guide

negarn:
write styleguide for sass


src/sass/README.md
```diff
@@ -0,0 +1,100 @@
...
+}
+```
+
+
+---
+
+### SVG
+
+<a id="svg-template-versus-sass"></a>
+**[Template versus Sass:](#svg-template-versus-sass)** Add SVGs as components if you need to add classes to modify them in different themes. Otherwise you may import them in Sass.
```
zakibinary: Just to add that we can also import the `SVG` directly to `Component` from `src/images` with  something like this `import SomeIcon from 'Images/app_2/header/some_icon.svg';`.

negarn: do we need to specify that? isn't that how we import all components anyway?

zakibinary: In this use case, we don't need to embed the `SVG` inside a `JSX` file, so we just directly import from `src/images/app_2`. This is useful if we want to have the `SVG` inline but don't need to pass `classes` to it.

negarn: I see what you mean. I'll add it.

src/sass/README.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+Keep all common variables in the [constants.scss](https://github.com/binary-com/binary-static/blob/master/src/sass/app_2/_common/base/constants.scss) file.
+
+---
+
+### Units
+
+<a id="units-flexibility"></a>
+**[Flexibility:](#units-flexibility)** If flexibility is needed, for example for font-size, use units such as `rem`, `vh`, `vw`, `fr`, and only use `px` if it's supposed to be a fixed value.
```
OskarAhl: I can live without it if no one agrees but I find using scalable rems as a base unit a convenient way of making a responsive layout 😄
```
html {
    font-size: 62.5%; /* 1rem = 10px */
    box-sizing: border-box;
    @include respond(big-desktop) {
        font-size: 75.5%; /* 1rem = 12px */
    }
    @include respond(tab-land) {
        font-size: 50%; /* 1rem = 8px */
    }
    @include respond(tab-port) {
        font-size: 47%; /* 1rem = 7.5px */
    }
    @include respond(phone-xl) {
        font-size: 38%; /* 1rem = 6px */
    }
    @include respond(phone-lg) {
        font-size: 35%; /* 1rem = 5.5px */
    }
}
```
(but can still use the other units for exceptions)

ashkanx: @OskarAhl Actually this is what I discussed with Negar and Kelly. my suggestion is we use `rem` and components will be relative to the root. then inside the container, we use `em` which will be relative to its component.

kellybinary: thanks @OskarAhl, will add this to my PR :)

PR #5006, opened 2019-02-08
msamprz/mt5-accounts-tooltip

msamprz:
Added Counterparty and Jurisdiction tooltip to MT5 accounts


src/javascript/app/pages/user/metatrader/metatrader.ui.js
```diff
@@ -185,6 +185,21 @@ const MetaTraderUI = (() => {
                     login   : () => (`${info} (${/demo/.test(accounts_info[acc_type].account_type) ? localize('Demo Account') : localize('Real-Money Account')})`),
                 };
                 $(this).html(typeof mapping[key] === 'function' ? mapping[key]() : info);
+
+                if (key === 'login') {
+                    const mt_financial_company = State.getResponse('landing_company.mt_financial_company');
```
negarn: better to do this outside of the loop, no need to fetch it every time, same as below

src/javascript/app/pages/user/metatrader/metatrader.ui.js
```diff
@@ -185,6 +185,21 @@ const MetaTraderUI = (() => {
...
+                    const mt_gaming_company = State.getResponse('landing_company.mt_gaming_company');
+                    const company = /standard/.test(acc_type) ?
+                                    mt_financial_company.standard
+                                    : /advanced/.test(acc_type) ?
+                                        mt_financial_company.advanced
+                                        : mt_gaming_company.standard;
```
negarn: please put a bracket around these three lines.
what happens if they add mt_gaming_company.advanced later? do we need to write it in a way that it won't break in that scenario?

src/javascript/app/pages/user/metatrader/metatrader.ui.js
msamprz: Noted and addressed in [#8fa5c77](https://github.com/binary-com/binary-static/pull/5006/commits/8fa5c77fa22915f8134b94b78381cc1f9114fe13)

src/javascript/app/pages/user/metatrader/metatrader.ui.js
msamprz: Noted and addressed in [#8fa5c77](https://github.com/binary-com/binary-static/pull/5006/commits/8fa5c77fa22915f8134b94b78381cc1f9114fe13)

src/javascript/app/pages/user/metatrader/metatrader.ui.js
```diff
@@ -186,6 +186,9 @@ const MetaTraderUI = (() => {
                 };
                 $(this).html(typeof mapping[key] === 'function' ? mapping[key]() : info);
             });
+
+            setCounterpartyAndJurisdictionTooltip($('.acc-info div[data="login"]'), acc_type);
```
negarn: better store `$('.acc-info div[data="login"]')` as a variable outside the map so we don't keep fetching the same element every time

src/javascript/app/pages/user/metatrader/metatrader.ui.js
```diff
@@ -561,6 +564,26 @@ const MetaTraderUI = (() => {
         }
     };
 
+    const setCounterpartyAndJurisdictionTooltip = ($el, acc_type) => {
+        const mt_financial_company = State.getResponse('landing_company.mt_financial_company');
+        const mt_gaming_company = State.getResponse('landing_company.mt_gaming_company');
```
negarn: this doesn't fix the issue that I mentioned. this function gets called multiple times in `.map`, and it keeps fetching the state response every time. better store it outside the map function.

msamprz: This function is being called outside of the `map()`, I have only found a single instance of `.map()` in the file:
```
            // Update account info
            $detail.find('.acc-info div[data]').map(function () {
                const key     = $(this).attr('data');
                const info    = accounts_info[acc_type].info[key];
                const mapping = {
                    balance : () => (isNaN(info) ? '' : Currency.formatMoney(MetaTraderConfig.getCurrency(acc_type), +info)),
                    leverage: () => `1:${info}`,
                    login   : () => (`${info} (${/demo/.test(accounts_info[acc_type].account_type) ? localize('Demo Account') : localize('Real-Money Account')})`),
                };
                $(this).html(typeof mapping[key] === 'function' ? mapping[key]() : info);
            });

            setCounterpartyAndJurisdictionTooltip($('.acc-info div[data="login"]'), acc_type);
```
I can't find any other instance of `.map()`, maybe I'm misunderstanding.

PR #5011, opened 2019-02-12
ahkuma/Allow client to reapply if rejected

ahkuma:


src/templates/app/_includes/professional_client.jsx
```diff
@@ -1,3 +1,4 @@
+
```
negarn: empty line

src/templates/app/user/settings/professional.jsx
```diff
@@ -19,7 +19,8 @@ const Professional = () => (
                 <p>{it.L('You are categorised as a professional client.')}</p>
             </div>
             <div className='invisible' id='rejected'>
-                <p>{it.L('Your request to be treated as a professional client has been rejected. Please check your inbox for more details. Your account remains categorised as a retail client account.')}</p>
+                <p>{it.L('Your request to be treated as a  professional is <strong>not approved</strong>.')}</p>
```
negarn: better use a placeholder rather than tags in string

src/templates/app/user/settings/professional.jsx
```diff
@@ -19,7 +19,8 @@ const Professional = () => (
...
+                <p>{it.L('Please check your inbox for details. Your account remains under the retail category.')} <br /> {it.L(' You are welcome at any time if you\'d like to reapply as a professional client.')} </p>
```
negarn: extra space before/after br tag and before closing p tag
please don't put a space in translation string. use `nbsp;` before it instead.

src/templates/app/user/settings/professional.jsx
```diff
@@ -19,7 +19,8 @@ const Professional = () => (
                 <p>{it.L('You are categorised as a professional client.')}</p>
             </div>
             <div className='invisible' id='rejected'>
-                <p>{it.L('Your request to be treated as a professional client has been rejected. Please check your inbox for more details. Your account remains categorised as a retail client account.')}</p>
+                <p>{it.L('Your request to be treated as a  professional is [_1]not approved[_2].', '<strong>', '</strong>')}</p>
```
negarn: there are two spaces before `professional`, and are we missing the word `client` after that?

negarn: as per the design only `approved` is supposed to be bold?

negarn: but I agree that making `not approved` bold is better, perhaps you can check with Lucas

ahkuma: Already checked with lucas earlier. We should be bolding `not approved`.

ahkuma: @negarn

src/javascript/app/pages/user/account/settings/professional_client.js
```diff
@@ -33,6 +33,10 @@ const professionalClient = (() => {
             return;
         }
 
+        $('#professional').setVisibility(0);
+        $('#processing').setVisibility(0);
+        $('#rejected').setVisibility(0);
```
negarn: we are fetching the selectors here and also in the setvisible function, it would be better to store them once and reuse them

src/templates/app/_includes/professional_client.jsx
```diff
@@ -4,10 +4,10 @@ import { Fieldset } from '../../_common/components/forms.jsx';
 const ProfessionalClient = () => (
     <Fieldset legend={it.L('Professional Client')} id='fs_professional' className='invisible'>
         <div className='gr-padding-10 gr-12'>
-            <p>{it.L('By default, all clients are treated as retail but anyone can request to be treated as professional.')}</p>
+            <p>{it.L('By default, all Binary.com clients are retail clients but anyone can request to be treated as a professional client.')}</p>
```
negarn: we don't hardcode Binary.com, please use it.website_name as a placeholder instead

src/javascript/app/pages/user/account/settings/professional_client.js
```diff
@@ -21,7 +21,7 @@ const professionalClient = (() => {
     const setVisible = (selector) => {
         $('#loading').remove();
         $('#frm_professional').setVisibility(0);
-        $(selector).setVisibility(1);
+        selector.setVisibility(1);
```
negarn: better keep the name as `$selector` since it's a jquery variable

PR #5032, opened 2019-02-17
easteregg/allow_non_authenticated_cr_clients_to_update

easteregg:


src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -98,11 +175,12 @@ const PersonalDetails = (() => {
     };
 
     const show_label_if_any_value   = ['account_opening_reason', 'citizen', 'place_of_birth', 'tax_residence', 'tax_identification_number'];
-    const force_update_fields       = ['tax_residence', 'tax_identification_number'];
+    const force_update_fields       = ['tax_residence', 'tax_identification_number'].concat(changeable_fields);
+    const has_changeable_fields     = Client.hasCostaricaAccount() && /^CR/.exec(Client.get('loginid')) && !is_fully_authenticated && !is_virtual;
```
negarn: this isn't global, but you're using it globally?

negarn: IMO we shouldn't be using `/^CR/.exec(Client.get('loginid'))`, we try not to hardcode login id as much as possible. just checking that current landing company shortcode is costarica can replace `Client.hasCostaricaAccount() && /^CR/.exec(Client.get('loginid'))`

negarn: then we can remove `!is_virtual` as well

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -362,16 +471,16 @@ const PersonalDetails = (() => {
...
             init();
             const account_status = State.getResponse('get_account_status').status;
             get_settings_data = State.getResponse('get_settings');
             is_fully_authenticated = checkStatus(account_status , 'authenticated');
             if (is_virtual) {
                 getDetailsResponse(get_settings_data);
             }
-
-            if (!is_virtual || !residence) {
+            if (!is_virtual || !residence || has_changeable_fields) {
```
negarn: will there every be a case where client is virtual or doesn't have residence but has changeable fields? I think no, if client is virtual there are no changeable fields because we only allow CR to do that, and if client doesn't have residence the first thing they have to do is select residence before the rest of their details is displayed to them.

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -98,11 +175,12 @@ const PersonalDetails = (() => {
...
 
     const show_label_if_any_value   = ['account_opening_reason', 'citizen', 'place_of_birth', 'tax_residence', 'tax_identification_number'];
-    const force_update_fields       = ['tax_residence', 'tax_identification_number'];
+    const force_update_fields       = ['tax_residence', 'tax_identification_number'].concat(changeable_fields);
+    const has_changeable_fields     = Client.hasCostaricaAccount() && /^CR/.exec(Client.get('loginid')) && !is_fully_authenticated && !is_virtual;
 
     const displayGetSettingsData = (get_settings, populate = true) => {
         Object.keys(get_settings).forEach((key) => {
-            const has_label         = show_label_if_any_value.includes(key);
+            const has_label         = show_label_if_any_value.includes(key) && !changeable_fields.includes(key);
```
negarn: please add a comment to explain this exception in the code

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -143,6 +221,10 @@ const PersonalDetails = (() => {
                         if (force_update) { // Force pushing values, used for (API-)expected values
                             $element.attr({ 'data-force': true, 'data-value': el_value });
                         }
+                        // Update data-value on change for inputs
+                        if (should_update_value) {
+                            $(element_key).change(function () {this.setAttribute('data-value', this.value);});
```
negarn: why is this needed? doesn't existing behaviour handle change of value?

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -199,6 +281,29 @@ const PersonalDetails = (() => {
                 { selector: '#chk_tax_id',     validations: is_financial ? [['req', { hide_asterisk: true, message: localize('Please confirm that all the information above is true and complete.') }]] : '', exclude_request: 1 },
             ];
 
+            // Push validations for changeable fields.
+            if (has_changeable_fields) {
```
negarn: I don't think we need an exception for pushing these to validations, we can include them in the validations list, since validator will handle if the selector for these doesn't exist or is invisible

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -54,6 +58,76 @@ const PersonalDetails = (() => {
...
+            changeable_fields = changeable.only_before_auth;
+        }
+    };
+
+    /**
+     * Remove labels and static fields and replace them with input when fields are changeable.
+     *
+     * @param {get_settings} to prepopulate some of the values.
+     */
+    const displayChangeableFields = (get_settings) => {
```
negarn: can't we use `displayGetSettingsData` to also handle showing/hiding input for changeable fields? it feels like we're hardcoding a lot of fields in this function, while they are automated in `displayGetSettingsData`. it feels like repeating the work, but in a non-dynamic way.

src/templates/app/user/settings/personal_details.jsx
```diff
@@ -36,8 +41,14 @@ const PersonalDetails = () => (
 
         <form className='form gr-padding-10 invisible' id='frmPersonalDetails'>
             <Fieldset legend={it.L('Details')}>
+                <Salutation className='invisible' />
+                <FirstName className='invisible' />
+                <LastName className='invisible' />
+                <DateOfBirth className='invisible' />
+                <Citizenship className='invisible' />
+                <AccountOpeningReason className='invisible' />
```
negarn: usually we put the label next to the input row so we know that both exist, e.g. look at citizenship below

src/javascript/app/pages/user/account/settings/personal_details.js
easteregg: Consider it done. Regarding global, I didn't get what you mean. Could you elaborate on that?

src/javascript/app/pages/user/account/settings/personal_details.js
easteregg: Unfortunately no, `data-value` does not get updated when changing data (value does) but `form.handleSubmit` does not pick it up.

src/javascript/app/pages/user/account/settings/personal_details.js
easteregg: Removing if statement. But I think letting the `forEach` loop stay, since it is more dynamic (with a few exceptions).

src/javascript/app/pages/user/account/settings/personal_details.js
easteregg: I'll refactor this to separate the previous behavior from injecting `has_changeable_field`

src/templates/app/user/settings/personal_details.jsx
easteregg: done

src/javascript/app/pages/user/account/settings/personal_details.js
easteregg: I can't think of a way that we can. Since I have to init data differently, we can't streamline it the way it has been done in `get_settings`. I would glad to refactor this if you have got some pointers to lead me to the right direction.

src/javascript/app/pages/user/account/settings/personal_details.js
negarn: should we add that in a bigger scale rather than only fixing it in this file then?

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -54,6 +59,70 @@ const PersonalDetails = (() => {
...
+        if (!has_changeable_fields) return;
+        changeable_fields.forEach(field => {
+            CommonFunctions.getElementById(`row_${field}`).setVisibility(1);
+        });
+
+        if (changeable_fields.includes('first_name') || changeable_fields.includes('last_name') || changeable_fields.includes('salutation')) {
```
negarn: would it be better to put these three in an array and then loop them? since adding a new field in the future would mean having to make it even longer

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -54,6 +59,70 @@ const PersonalDetails = (() => {
...
+            CommonFunctions.getElementById('row_name').setVisibility(0);
+            CommonFunctions.getElementById('first_name').setVisibility(1);
+            CommonFunctions.getElementById('last_name').setVisibility(1);
+            CommonFunctions.getElementById('salutation').setVisibility(1);
+        }
+
+        if (changeable_fields.includes('date_of_birth')) {
+            const $input_el = $('#date_of_birth');
+            $input_el.setVisibility(1);
```
negarn: why not chain this with below?

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -54,6 +59,70 @@ const PersonalDetails = (() => {
...
+                .change(function() {
+                    return CommonFunctions.dateValueChanged(this, 'date');
+                });
+
+            DatePicker.init({
+                selector: '#date_of_birth',
+                minDate : -100 * 365,
+                maxDate : (
+                    -18 * 365
+                ) - 5,
```
negarn: can you explain this?

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -54,6 +59,70 @@ const PersonalDetails = (() => {
...
+                yearRange: '-100:-18',
+            });
+        }
+
+        const $options               = $('<div/>');
```
negarn: extra space before equal

negarn: should we put it inside the if? seems it's only used in there?

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -113,7 +187,7 @@ const PersonalDetails = (() => {
                 if (populate) {
                     const should_update_value = /select|text/i.test(element_key.type);
                     if (has_label) {
-                        CommonFunctions.getElementById(`row_${element_id}`).setVisibility(1);
+                        CommonFunctions.getElementById(`row_lbl_${element_id}`).setVisibility(1);
```
negarn: how was it working before?

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -362,38 +469,45 @@ const PersonalDetails = (() => {
...
             get_settings_data = State.getResponse('get_settings');
             is_fully_authenticated = checkStatus(account_status , 'authenticated');
+            has_changeable_fields = Client.get('landing_company_shortcode') === 'costarica' && !is_fully_authenticated;
+
             if (is_virtual) {
                 getDetailsResponse(get_settings_data);
+            } else if (has_changeable_fields) {
+                populateChangeableFields();
+                displayResidenceList();
+            } else if (!is_virtual || !residence) {
```
negarn: doesn't `!residence` take priority over `has_changeable_fields`?

negarn: we don't need to check `!is_virtual` here anymore, if client was virtual they would have gone to first if.

src/javascript/app/pages/user/account/settings/personal_details.js
easteregg: template naming was not consistent. I've made a change to prefix all labels with `row_lbl`.

src/javascript/app/pages/user/account/settings/personal_details.js
easteregg: I'll refactor to (is_virtual -> residence -> changeable_fields)

src/javascript/app/pages/user/account/settings/personal_details.js
easteregg: I don't get it. Do you want me to add a comment in the code? or just explain it here?

negarn: comment in the code :)

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -54,6 +59,70 @@ const PersonalDetails = (() => {
...
+        if (!has_changeable_fields) return;
+        changeable_fields.forEach(field => {
+            CommonFunctions.getElementById(`row_${field}`).setVisibility(1);
+        });
+
+        if (['first_name', 'last_name', 'salutation'].some(key => changeable_fields.includes(key))) {
+            CommonFunctions.getElementById('row_name').setVisibility(0);
+            CommonFunctions.getElementById('first_name').setVisibility(1);
+            CommonFunctions.getElementById('last_name').setVisibility(1);
+            CommonFunctions.getElementById('salutation').setVisibility(1);
```
negarn: I meant if we loop the array, we can combine these three lines to `CommonFunctions.getElementById(looping_item).setVisibility(1);`, and handle setting row_name to invisible once

negarn: but if you think this is better then keep it

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -21,11 +23,14 @@ const PersonalDetails = (() => {
...
         residence,
-        get_settings_data;
+        get_settings_data,
+        has_changeable_fields,
+        changeable_fields;
 
     const init = () => {
         editable_fields   = {};
         get_settings_data = {};
+        changeable_fields = [];
```
negarn: shouldn't we add tax res and tax id here until API returns it?

src/sass/app/websocket_pages.scss
```diff
@@ -283,3 +283,9 @@
         justify-content: center;
     }
 }
+
+.realws-content {
+    div.select-dropdown {
+        width: 100%;
```
negarn: we have some very long state names, or country names, I can't remember which. sometimes 100% causes it to become too long. please check this,

easteregg: Yeah, countries, this doesn't affect that one. Thanks for pointing that out. I'll double check the mobile ones.

PR #5033, opened 2019-02-18
kelly/refactor_font_size

kellybinary:


src/sass/README.md
```diff
@@ -52,6 +52,64 @@ Keep all common variables in the [constants.scss](https://github.com/binary-com/
...
+    @include toEm(padding, 10px, 1.4em); // font-size in em
+}
+```
+3. Or any online converter tool.
+
+---
+
+### Typography
+The `@typeface($var, $text-transform, $text-align)` mixin can be used to style any text element. Simply pass in a typeface `$var` name to the mixin.
+The `$var` name is in the format `--$FONT_SIZE_-$FONT_WEIGHT-$COLOR`.
```
negarn: is there an extra `_` after font_size?

kellybinary: nope, that's a typo. thanks!

PR #5040, opened 2019-02-19
oskar/subscribe_on_buy

OskarAhl:
* Add subscription to proposal_open_contract from buy request


src/javascript/app_2/Services/ws-methods.js
```diff
@@ -73,6 +73,14 @@ const WS = (() => {
     const subscribeProposalOpenContract = (contract_id = null, cb, should_forget_first) =>
         SubscriptionManager.subscribe('proposal_open_contract', { proposal_open_contract: 1, subscribe: 1, ...(contract_id && { contract_id }) }, cb, should_forget_first);
 
+    const subscribeProposalOpenContractAfterBuy = (buy_request) =>
+        SubscriptionManager.addSubscriptionFromRequest(
+            'proposal_open_contract',
+            { ...buy_request, subscribe: 1 },
+            { proposal_open_contract: 1, subscribe: 1, contract_id: '' },
```
kellybinary: do you need `contract_id`? i think you can remove `contract_id: ''` here

src/javascript/app_2/Stores/Modules/Portfolio/portfolio-store.js
```diff
@@ -70,11 +70,7 @@ export default class PortfolioStore extends BaseStore {
             WS.subscribeProposalOpenContract(contract_id, this.proposalOpenContractHandler, false);
         } else if (act === 'sell') {
             // TODO: Refactor with contract-store and use common helpers to handle contract result
```
zakibinary: Can you remove the comment here?

src/javascript/_common/base/subscription_manager.js
```diff
@@ -63,6 +63,45 @@ const SubscriptionManager = (() => {
...
+     * @param {Object}   send_request           the object of the request to be made
+     * @param {Object}   subscribe_request      the object of the subscription request
+     * @param {String}   subscription_prop      prop to add to subscribe_request from initial request, e.g. contract_id
+     */
+    const addSubscriptionFromRequest = (msg_type, send_request, subscribe_request, subscription_prop = '') =>
+        new Promise((resolve, reject) => {
+            let sub_id;
+            let is_stream = false;
+
+            BinarySocket.send(send_request, {
```
morteza-binary: @OskarAhl Shouldn't it manage to prevent from sending duplicate buy requests and just subscribes them to the current subscription?

src/javascript/app_2/Services/ws-methods.js
```diff
@@ -73,6 +73,14 @@ const WS = (() => {
     const subscribeProposalOpenContract = (contract_id = null, cb, should_forget_first) =>
         SubscriptionManager.subscribe('proposal_open_contract', { proposal_open_contract: 1, subscribe: 1, ...(contract_id && { contract_id }) }, cb, should_forget_first);
 
+    const subscribeProposalOpenContractAfterBuy = (buy_request) =>
```
morteza-binary: @OskarAhl IMO, it'd be better to change the function name, it conveys that we should call it after sending buy response.

src/javascript/_common/base/subscription_manager.js
OskarAhl: @morteza-binary Should it? Currently it's the same buy as the buy in the `ws-methods.js`, and the purpose is to add a subscription without any subscribers - since the subscription is implicitly added as a side effect from sending the buy request.

src/javascript/app_2/Services/ws-methods.js
OskarAhl: Would `subscribeProposalOpenContractOnBuy` be more clear?

src/javascript/_common/base/subscription_manager.js
morteza-binary: so it's fine.

PR #5050, opened 2019-02-21
[new-app] zaki/positions_update_part1

zakibinary:
Changelog
- Update positions drawer/sidebar design.
- Fixed issues with positions drawer when switching accounts.
- Added sell button on sidebar for each running position, temporary toast message is used before refactoring `contract_store` along with `contract_mode` components.


src/javascript/app_2/App/Components/Elements/PositionsDrawer/ProgressSlider/positions-progress-slider.jsx
```diff
@@ -0,0 +1,58 @@
...
+    has_result,
+    ticks_count,
+    current_tick,
+    percentage,
+    remaining_time,
+}) => {
+    if (!percentage && !ticks_count || has_result) return <div className='progress-slider--completed' />;
+    return (
+        <div className={classNames('progress-slider', className)}>
+            {(ticks_count < 5) ?
```
OskarAhl: Why `5`?

zakibinary: Because design isn't ready yet for ticks duration and after testing it only makes sense to show duration slider in place ticks slider for `tick_counts` above 5.  Regardless this is still a work in progress and will be handled in the subsequent card/pr.

PR #5051, opened 2019-02-21
ahkuma/advance_duration_bug

ahkuma:


src/javascript/app_2/App/Components/Form/TimePicker/dialog.jsx
```diff
@@ -6,16 +6,20 @@ import { toMoment } from 'Utils/Date';
 
 const Dialog = ({
     preClass,
-    value,
-    start_time,
-    end_time,
+    selected_time,
+    boundary_end,
+    boundary_start,
```
negarn: let's stick to start_time and start_end usage, as it's common with API naming and we all know what it refers to

package.json
```diff
@@ -136,9 +136,9 @@
     "react-transition-group": "2.4.0",
     "scriptjs": "2.5.8",
     "sinon": "7.2.2",
-    "smartcharts-beta": "0.3.92",
+    "smartcharts-beta": "0.3.93",
```
zakibinary: If you update any dependency please update the accompanying Trello card as a `SRP` (Special Release Procedure) and also add it to the PR's changelog / description. It will also be helpful to document/log what changes are added to this PR for future reference.

src/javascript/app_2/Modules/Trading/Components/Form/TimePicker/trading-time-picker.jsx
```diff
@@ -2,41 +2,73 @@ import PropTypes   from 'prop-types';
...
+    const is_expired_next_day = contract_expiry.diff(contract_start, 'day') === 1;
+    const expiry_time_start = is_expired_next_day
+        ? contract_expiry.clone().startOf('day')
+        : contract_expiry.clone();
+    const expiry_time_end   = is_expired_next_day
+        ? minDate(
+            contract_expiry.clone().subtract(10, 'minute'),
+            market_next_day_close)
+        : market_next_day_close.clone();
+    */
```
OskarAhl: Why is this commented out?

ahkuma: Mainly because we are not using `start_time` right now, and I want to implement it differently because I think it's quite confusing. I feel like it is trying to do 2 things at once, so rather than that, I think we should separate them, this is where the helper files come in. If we support `start_time` again, we should add the required logic in a helper file then do a check to see which modes is used and use the functions respectively, and if there is any more modes to be added, we could easily add more.

negarn: IMO rewriting the function, if it doesn't have any part in resolving the advanced duration bug, shouldn't be done in the same PR. if it is related to fixing the bug, then it should be rewriting and removing the old code, not keeping the old code as comments.

ahkuma: I am not sure if it qualifies as related to fixing the bug or not, however I rewrote it to fix the bug because the old code is too confusing and complicated (everytime I fix one bug, another bug appears). Keeping the old code is just as a reference because Morteza said we should keep the code for when we need to re-enable start time again. I think once we reimplement start time, we can remove the comments.

OskarAhl: I think the cleaner way would be to save these changes in either a new PR,  or in the git history and If we reimplement start time in the future -  add it back.

PR #5055, opened 2019-02-22
msamprz/acc-switcher-light

msamprz:
+ Changed placement of add account title symbol to match design spec
+ Increased border size of selected account radio option
+ Changed array reduction implementation for getting virtual accounts
+ Added check to only show virtual account in switcher only if it exists
+ Added check for real account title to be based on number of accounts
+ Fixed condition that sometimes prevented real accs from being shown

### DIFF
--- a/src/javascript/app_2/App/Containers/AccountSwitcher/account-switcher.jsx
+++ b/src/javascript/app_2/App/Containers/AccountSwitcher/account-switcher.jsx
@@ -2,6 +2,7 @@ import classNames               from 'classnames';
 import PropTypes                from 'prop-types';
 import React                    from 'react';
 import { localize }             from '_common/localize';
+import { isEmptyObject }        from '_common/utility';
 import { IconAccountsCurrency } from 'Assets/Header/AccountsCurrency';
 import { IconLogout }           from 'Assets/Header/Drawer';
 import { requestLogout }        from 'Services/index';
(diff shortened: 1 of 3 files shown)


src/javascript/app_2/App/Containers/AccountSwitcher/account-switcher.jsx
```diff
@@ -46,20 +47,22 @@ class AccountSwitcher extends React.Component {
...
 
-        const main_account_title = localize('Real account');
+        const main_account_title = real_accounts.length > 1 ? localize('Real accounts') : localize('Real account');
 
         return (
             <div className='acc-switcher__list' ref={this.setWrapperRef}>
                 {
                     // Make sure this block is not rendered if there are no real accounts
-                    !!(this.props.account_list.length > 0 && !(this.props.account_list[0].is_virtual)) &&
+                    !!(this.props.account_list.length > 0 && real_accounts.length > 0) &&
```
nazaninreihani: suggestion:
in conditionals better to use `real_accounts.length`  instead of `real_accounts.length > 0`, the first one is shorter and also better in performance

src/javascript/app_2/App/Containers/AccountSwitcher/account-switcher.jsx
```diff
@@ -88,22 +91,25 @@ class AccountSwitcher extends React.Component {
...
+                    <div className='acc-switcher__list--virtual'>
+                        <span className='acc-switcher__list-title'>
+                            {localize('Virtual account')}
                         </span>
-                        <span className='acc-switcher__radio' />
+                        <div
+                            className={classNames('acc-switcher__account', {
+                                'acc-switcher__account--selected': (this.props.virtual_loginid === this.props.account_loginid),
+                            })}
+                            onClick={this.doSwitch.bind(this, this.props.virtual_loginid)}
```
nazaninreihani: why not binding this to doSwitch in constructor method?  to prevent react re-render (arrow functions and inline bindings make the component re-render every time)

src/javascript/app_2/App/Containers/AccountSwitcher/account-switcher.jsx
msamprz: Agreed, good idea. Thanks, will make the change in the new branch.

src/javascript/app_2/App/Containers/AccountSwitcher/account-switcher.jsx
msamprz: @nazaninreihani that's for virtual account, right, and so I can't seem to work that out with the other part for Real Accounts as it relies on the `map()` call to map the individual accounts, and the `loginid` of the account is passed to `doSwitch()`. Do you have any suggestions? here's the code:

```
real_accounts.map(account => (
  <div
    key={account.loginid}
    className={classNames("acc-switcher__account", {
      "acc-switcher__account--selected":
        account.loginid === this.props.account_loginid
    })}
    onClick={this.doSwitch.bind(this, account.loginid)}
  >
    <span className={"acc-switcher__id"}>
      <IconAccountsCurrency
        className={`acc-switcher__id-icon acc-switcher__id-icon--${
          account.icon
        }`}
        type={account.icon}
      />
      {account.loginid}
    </span>
  </div>
));
                            
```

msamprz: Okay, I could do all of this before `render()` is called I guess, but for now we're going to create a card for `Optimizing re-rendering components`, so I'll just add this in there too.

PR #5062, opened 2019-02-25
negar/add_bem

negarn:
add more BEM


src/javascript/app_2/App/Components/Layout/Header/login-button.jsx
```diff
@@ -7,7 +7,7 @@ import Button              from '../../Form/button.jsx';
 
 const LoginButton = ({ className }) => (
     <Button
-        className={classNames(className, 'secondary orange')}
+        className={classNames(className, 'btn-secondary btn--secondary--orange')}
```
OskarAhl: `btn-secondary` should be `btn--secondary`?

src/javascript/app_2/App/Components/Form/DatePicker/date-picker.jsx
```diff
@@ -230,7 +230,7 @@ class DatePicker extends React.Component {
             >
                 { this.renderInputField() }
                 <IconCalendar
-                    className={classNames('datepicker__icon datepicker__icon--calendar', {
+                    className={classNames('input-field__icon datepicker__icon datepicker__icon--calendar', {
```
OskarAhl: `input-field__icon `?

negarn: these icons that project inside the input field have common styling. I can rename it and move it out of input-field.scss if you think that's better.

OskarAhl: Where is the parent block `input-field`?

negarn: there's no parent, it's a sibling. in the old code it was handled with a +

OskarAhl: Oh I see, would it be less coupled if we added a mixin for this common icon style instead?

negarn: sure, I will do that instead.

PR #5075, opened 2019-02-26
kelly/add_longcode_component

kellybinary:
- add InfoBoxLongcode component to InfoBox for SmartCharts's top widget


src/sass/app_2/modules/contract.scss
```diff
@@ -119,7 +119,17 @@ $CONTRACT_INFO_BOX_PADDING: 1.5em;
...
     transition: transform 0.4s ease, opacity 0.4s linear;
+    display: flex;
+    justify-content: space-between;
+    width: 100%;
 
+    &-button:hover {
+        cursor: pointer;
+    }
+    &-icon {
+        width: 3.2em;
```
zakibinary: I still have my reservations with using `em` for `width` and `height` for icons. In some cases, it seems like the scaling is off when using this method.

kellybinary: sure, i've change it to px

PR #5078, opened 2019-02-27
negar/add_bem

negarn:
more BEM


src/sass/app_2/_common/layout/footer.scss
```diff
@@ -122,7 +123,7 @@ footer {
             top: 0;
             left: 0;
         }
-        .inline-icon {
+        .ic-settings__icon {
```
kellybinary: `&__icon` ?

negarn: then I would have to change the class to `ic-settings--active__icon`, which is not the pattern we want to follow

negarn: the icon for ic-settings--active has a different styling than non-active

kellybinary: ahh, didn't see the `--active` please ignore this

PR #5080, opened 2019-02-27
ahkuma/Notify Currency selection to Clients on Cashier page.

ahkuma:


src/javascript/app/pages/cashier/payment_agent_list.js
```diff
@@ -1,6 +1,7 @@
 const BinaryPjax   = require('../../base/binary_pjax');
 const Client       = require('../../base/client');
 const BinarySocket = require('../../base/socket');
+const Url          = require('../../../_common/url');
 const urlForStatic = require('../../../_common/url').urlForStatic;
```
negarn: if we have Url, we can remove urlForStatic and instead use `Url.UrlForStatic` in code below

ahkuma: Yes, I fixed it now.

src/javascript/app/pages/cashier/payment_agent_list.js
```diff
@@ -16,6 +16,11 @@ const PaymentAgentList = (() => {
             });
         });
 
+        if (!Client.get('currency')) {
```
negarn: do we need to check if client is logged in then only check the currency? we should do that if we allow logged out clients to see this page

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -172,6 +173,11 @@ const PaymentAgentWithdraw = (() => {
 
     const onLoad = () => {
         BinarySocket.wait('get_account_status').then((data) => {
+            if (!Client.get('currency')) {
```
negarn: why do we need to wait for account status to check currency? don't we already get that from authorize? and this page only loads when authorize response is received

src/javascript/app/pages/cashier/payment_agent_list.js
negarn: same here, why do we go through doing the accordion action before redirecting? it should be the first thing we do

PR #5085, opened 2019-02-28
amam/fix regex signed for barrier

cakasuma:


src/javascript/app_2/Utils/URL/url-helper.js
```diff
@@ -10,7 +10,8 @@ export default class URLHelper {
      */
     static getQueryParams(url) {
         const query_string =  url ? new URL(url).search : window.location.search;
-        const query_params = new URLSearchParams(query_string.slice(1));
+        const query_encoded = encodeURIComponent(query_string);
```
negarn: we don't need the slice(1) anymore?

cakasuma: `slice(1)` is to remove the `?` before uri, however, since this is encoded, it will become `%3F` instead, if we use `slice(1)` it will remove the `%` instead and the decoder will not be able to decode

PR #5092, opened 2019-03-01
Amam/digit audit

cakasuma:
- combine contract start and contract end audit


src/javascript/app/pages/user/view_popup/view_popup.js
```diff
@@ -554,26 +557,45 @@ const ViewPopup = (() => {
...
-            } else {
-                contract_starts.div.remove();
-            }
-            // don't show exit tick information if missing or manual sold
-            if (contract.audit_details.contract_end && contract.status !== 'sold') {
-                parseAuditResponse(contract_starts.table, contract.audit_details.contract_end).then(() => {
-                    if (contract.audit_details.contract_end) {
-                        appendAuditLink('trade_details_current_spot');
-                    }
+        if (!is_tick_contract) {
```
negarn: can we simply check if `contract.tick_count` instead of making a new global variable?

src/javascript/app/pages/user/view_popup/view_popup.js
```diff
@@ -553,30 +554,47 @@ const ViewPopup = (() => {
...
+            parseAuditResponse(contract_details.table, contract.audit_details.all_ticks).then(() => {
+                if (contract.audit_details.all_ticks) {
+                    createAuditHeader(contract_details.table);
+                    appendAuditLink('trade_details_entry_spot');
+                    appendAuditLink('trade_details_current_spot');
+                } else {
+                    contract_details.div.remove();
+                }
+
+                if (contract.status === 'sold') {
```
negarn: is this the best way to check that contract has ended? why not check contract.status !== 'open'?

cakasuma: yes, you are correct. its better to check if the contract.status is not open 👍

PR #5094, opened 2019-03-04
negar/add_bem

negarn:
add drawer component to BEM


src/javascript/app_2/App/Components/Elements/Drawer/drawer.jsx
```diff
@@ -46,16 +46,20 @@ class Drawer extends React.Component {
         const { is_this_drawer_on } = this.state;
         const { alignment, closeBtn, children } = this.props;
 
-        const drawer_bg_class = classNames('drawer-bg', {
-            'show': is_this_drawer_on,
+        const drawer_bg_class = classNames('drawer__bg', {
+            'drawer--show': is_this_drawer_on,
         });
-        const drawer_class = classNames('drawer', alignment);
+        const drawer_class = classNames('drawer', `drawer--${alignment}`);
```
zakibinary: Maybe check if alignment prop is `left` or `right` before adding `className`?

PR #5098, opened 2019-03-05
amam/fix_barrier_drag

cakasuma:
when barrier dragged from smartcharts, the + is parsed to int and disappeared


src/javascript/app_2/Stores/Modules/Trading/Constants/validation-rules.js
```diff
@@ -17,7 +17,7 @@ const getValidationRules = () => ({
         rules: [
             ['req'    , { condition: store => store.barrier_count && store.form_components.indexOf('barrier') > -1, message: localize('Barrier is a required field.') }],
             ['barrier', { condition: store => store.contract_expiry_type !== 'daily' && store.barrier_count }],
-            ['number' , { condition: store => store.contract_expiry_type === 'daily' && store.barrier_count, type: 'float' }],
+            ['number' , { condition: store => store.contract_expiry_type === 'daily' && store.barrier_count, type: 'float', input: 'barrier' }],
```
OskarAhl: Removing this number validation + changing the barrier validation to also apply for daily expiry type should work without having to make changes to `declarative-validation-rules.js`?

PR #5100, opened 2019-03-05
oskar/chart_marker_components

OskarAhl:
* Add marker component (handles spot without label)

* Add marker component with label

* Add styling for: entry spot, exit spot, middle spot


src/javascript/app_2/Modules/SmartChart/Components/Markers/marker-spot-label.jsx
```diff
@@ -0,0 +1,69 @@
+import { observer }  from 'mobx-react';
+import PropTypes     from 'prop-types';
+import React         from 'react';
+import { addComma }  from '_common/base/currency_base';
+import { toMoment }  from 'Utils/Date';
+import MarkerSpot    from './marker-spot.jsx';
+import { IconClock } from '../../../../Assets/Common/icon-clock.jsx';
```
kellybinary: think you can use absolute import here: `'Assets/Common/icon-clock.jsx'`

PR #5115, opened 2019-03-06
Nazanin/tests_failing

nazaninreihani:


src/javascript/app_2/Utils/Date/__tests__/date-time.spec.js
```diff
@@ -63,7 +63,7 @@ describe('daysFromTodayTo', () => {
 
     it('return difference value between selected date and today', () => {
         //get date three days from now
-        const date = moment().utc().startOf('day').add('3', 'days').format('YYYY-MM-DD');
+        const date = moment().utc().startOf('day').add('3', 'days');
```
negarn: why do we need to remove the formatting? `daysFromTodayTo` changes date to moment format, so it shouldn't be passed a moment for test

nazaninreihani: @negarn it fails in here because when we set it to YYYY-MM-DD it removes the hour and the minute from UTC date and that causes to fail the tests in Canada, it returns 2 days diff instead of 3 days because of removing hours from UTC date

negarn: so we should fix the hours before adding 3 days, to make sure it's really three days, and then keep the formatting at the end

nazaninreihani: the issue is here https://github.com/binary-com/binary-static/blob/761780e77b267bd370a110c8a68595aff3150a3f/src/javascript/app_2/Utils/Date/date-time.js#L23, when we format the date YYYY-MM-DD it's no longer UTC and skips the previous condition, and in this line, moment(new Date(value)) returns yesterday instead of today (in some timezones), so the solution is to convert it to utc right before formatiing it, `moment(new Date(value)).utc().format('YYYY-MM-DD')`

msamprz: This specific issue has been addressed and resolved now, but further fixes/changes to the datepicker is incoming, to switch datepicker from consuming local date/time to server date/time, so please don't make any changes in the meanwhile.

PR #5116, opened 2019-03-07
msamprz/remove-open-roles

msamprz:
Remove `Translator` and `Proofreader` positions


src/javascript/static/pages/job_details.js
```diff
@@ -30,7 +31,16 @@ const JobDetails = (() => {
...
+
+        window.location.hash = ($sidebar_dept.length > 0 && window.location.hash === '')
+            ? $sidebar_dept.find('a')[0].hash
+            : window.location.hash;
+
+        if ($sidebar_dept.length === 0 || $sidebar_dept.find(`a[href="${window.location.hash}"]`).length === 0) {
+            load(urlFor('404'));
+        }
+
+        // show sectionl
```
negarn: extra `l` at the end

src/javascript/static/pages/job_details.js
```diff
@@ -30,7 +31,16 @@ const JobDetails = (() => {
         $sidebar.setVisibility(0);
         $('#title').find('h1').setVisibility(0);
         $('#image').find('img').setVisibility(0);
-        // show section
+
+        window.location.hash = ($sidebar_dept.length > 0 && window.location.hash === '')
+            ? $sidebar_dept.find('a')[0].hash
```
negarn: I remember we don't want to use something like this, we want to have a list of valid hash, because of the injection security concern raised by Shashank. so we decided to simply leave it blank rather than dynamically show content.

src/javascript/static/pages/job_details.js
```diff
@@ -30,7 +31,14 @@ const JobDetails = (() => {
         $sidebar.setVisibility(0);
         $('#title').find('h1').setVisibility(0);
         $('#image').find('img').setVisibility(0);
-        // show section
+
+        if ($sidebar_dept.length && window.location.hash === '') {
+            load(`${urlFor('open-positions/job-details')}?dept=${dept}${$sidebar_dept.find('a')[0].hash}`);
```
negarn: can you use BinaryPjax.Load instead? I think we use it that way everywhere, it's easier to find that way

PR #5117, opened 2019-03-07
negar/rename_run

negarn:
rename to only ups/only downs


src/javascript/app/pages/user/view_popup/view_popup.js
```diff
@@ -120,8 +120,8 @@ const ViewPopup = (() => {
             PUTSPREAD   : localize('Put Spread'),
             TICKHIGH    : localize('High Tick'),
             TICKLOW     : localize('Low Tick'),
-            RUNHIGH     : localize('Run Ups'),
-            RUNLOW      : localize('Run Downs'),
+            RUNHIGH     : localize('Only Ups'),
```
easteregg: Does the API still send the RUNHIGH/RUNLOW as a category name, shouldn't they change it too? 🤔

negarn: I believe they still send runs. I don't think we can change that, it will break any apps using it. changing display name is not consequential though.

PR #5119, opened 2019-03-07
easteregg/add-loading-popup

easteregg:


src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -38,12 +41,20 @@ const DigitDisplay = (() => {
...
                     .append($('<strong />', { class: 'gr-6', text: localize('Spot Time (GMT)') }))))
             .append($('<div />', { class: 'digit-ticker invisible', id: 'digit_ticker_container' }));
+        $digit_table = $('#table_digits');
+        $digit_table
+            .append($('<div />', { class: 'barspinner dark' })
+                .append($('<div />', { class: 'rect1' }))
+                .append($('<div />', { class: 'rect2' }))
+                .append($('<div />', { class: 'rect3' }))
+                .append($('<div />', { class: 'rect4' }))
+                .append($('<div />', { class: 'rect5' })));
```
negarn: is it worth moving the adding of a spinner to a utility function and passing `$digit_table` to it?

src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -101,7 +112,8 @@ const DigitDisplay = (() => {
         if (getPropertyValue(response, ['tick', 'id']) && document.getElementById('sell_content_wrapper')) {
             ViewPopupUI.storeSubscriptionID(response.tick.id);
         }
-
+        $digit_table.find('.barspinner').remove();
+        // Hide loading component
```
negarn: why is the comment below the line the does the action? 😅

easteregg: :joy: totally my bad.

src/javascript/app/pages/trade/digit_trade.js
easteregg: Yeah, move it to it's own function.

src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -29,7 +30,20 @@ const DigitDisplay = (() => {
         }
     };
 
+    function setLoadingSpinner(content_id) {
```
negarn: you're expecting content_id, but on line 61 you're not passing any parameters.

src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -29,7 +30,20 @@ const DigitDisplay = (() => {
...
+        $digit_table = $(`#${content_id}`);
+        $digit_table
+            .append($('<div />', { class: 'barspinner dark' })
```
negarn: this isn't really what I meant. I meant moving this function to utility file etc so all pages can call it if needed. if we're still only specifically using it for digit table there's not much point in adding a separate function for it.

easteregg: I see what you mean now. This is a great idea. I generalize it a bit and move it to a separate module. (with show/hide functionality).

PR #5125, opened 2019-03-07
Nazanin/asset_index

nazaninreihani:
show an error to user when asset index is empty (restricted countries)


src/templates/app/resources/asset_index.jsx
```diff
@@ -6,6 +6,9 @@ const AssetIndex = () => (
         <div className='gr-padding-10'>
             <p id='errorMsg' className='error-msg invisible' />
             <div id='asset-index' className='has-tabs gr-parent' />
+            <p className='notice-msg center-text invisible' id='empty-asset-index'>
+                {it.L('Asset index is unavailable at this time.')}
```
negarn: @ashkanx is this text approved?

src/javascript/app/pages/resources/asset_index/asset_index.ui.js
```diff
@@ -29,6 +30,8 @@ const AssetIndexUI = (() => {
     const populateTable = () => {
         if (!active_symbols || !asset_index) return;
 
+        if (!asset_index.length) $('#empty-asset-index').setVisibility(1);
```
negarn: do we need to return after this?

nazaninreihani: return here causes the showLoadingImage fire and the loader is visible constantly

negarn: so we should hide the loader and return, doesn't make sense to do all these actions to get asset index data and build the table when there's no data to show

src/templates/app/resources/asset_index.jsx
ashkanx: No. I added a ticket

PR #5128, opened 2019-03-08
amam/interview_eu_modal

cakasuma:


src/javascript/app/base/interview_popup.js
```diff
@@ -0,0 +1,37 @@
+const Cookies        = require('js-cookie');
+const BinarySocket   = require('./socket');
+const isEuCountry    = require('../common/country_base').isEuCountry;
+const Client         = require('../base/client');
```
negarn: extra spaces before `=`, client should be on line 3 (base comes before common alphabetically)

src/javascript/app/base/interview_popup.js
```diff
@@ -0,0 +1,37 @@
...
+
+const InterviewPopup = (() => {
+    const onLoad = () => {
+        BinarySocket.wait('authorize', 'website_status', 'get_account_status').then(() => {
+            if (Client.isLoggedIn() && isEuCountry() && !Cookies.get('InterviewConsent')) {
```
negarn: better use a boolean variable format, something like `is_interview_consent` etc

src/javascript/app/base/interview_popup.js
```diff
@@ -0,0 +1,37 @@
...
+                const interview_popup = $('#interview_popup_container');
+                const interview_no_thanks = $('#interview_no_thanks');
+                const interview_ask_later = $('#interview_ask_later');
+                const interview_interested = $('#interview_interested');
```
negarn: these are all jquery elements, so the variable name should start with `$`, i.e. `$interview_popup` as per the coding styleguide

src/javascript/app/base/interview_popup.js
```diff
@@ -0,0 +1,37 @@
...
+
+                interview_popup.removeClass('invisible');
+                interview_no_thanks.on('click', () => {
```
negarn: do we want to use `one` (http://api.jquery.com/one/) so it triggers only once?

src/javascript/app/base/page.js
```diff
@@ -86,6 +87,7 @@ const Page = (() => {
             updateLinksURL('body');
             recordAffiliateExposure();
             endpointNotification();
+            InterviewPopup.onLoad();
```
negarn: I think we also don't want to call this on isLoginPages() similar to line 81, so we should move it or if we want to keep the hierarchy of function calls we should store isLoginPages() it in a variable and check that again here

src/templates/_common/components/interview_popup.jsx
```diff
@@ -6,22 +6,21 @@ const InterviewPopup = () => (
         <HTMLComment>googleoff: all</HTMLComment>{/* https://support.google.com/gsa/answer/6329153?hl=en#82542 */}
         <div className='popup'>
             <div className='popup__head'>
-                <img className='popup__present_img' src={it.url_for('images/interview_popup/present.svg')} alt='$30 present' />
-                <div className='header-1'>{it.L('Do you have 40 minutes for an interview?')}</div>
+                <div className='header-1'>{it.L('Can you spare 15 minutes?')}</div>
             </div>
             <div className='popup__body'>
-                <div className='header-2'>{it.L('Earn $30 to trade on [_1]', it.website_name.toLowerCase())}</div>
+                <div className='header-2'>{it.L('We’d love to hear what you think.')}</div>
```
negarn: instead of `’` please use `\'`

src/templates/_common/components/interview_popup.jsx
```diff
@@ -6,22 +6,21 @@ const InterviewPopup = () => (
...
                 <p>
-                    {it.L('We\'re looking for users of [_1] to participate in a [_2]40-minute video or phone interview[_3]. To qualify, just answer a few short questions. If selected, you will receive an email or a phone call from one of our researchers.', it.website_name.toLowerCase(), '<strong>', '</strong>')}
+                    {it.L('We’re looking to improve our products and services, and we want to understand your needs better. We’d like to interview you via phone, to know what you like about us, what you don’t like, and where we can do better.')}
```
negarn: same here

src/templates/_common/components/interview_popup.jsx
```diff
@@ -6,22 +6,21 @@ const InterviewPopup = () => (
...
-                        <a href={'javascript:void(0);'} id='interview_no_thanks'>{it.L('No thanks')}</a>
+                        <a id='interview_no_thanks'>{it.L('No thanks')}</a>
                         <span className='popup__separator' />
-                        <a href={'javascript:void(0);'} id='interview_ask_later'>{it.L('Ask me later')}</a>
+                        <a id='interview_ask_later'>{it.L('Ask me later')}</a>
                     </div>
-                    <a href={'javascript:void(0);'} id='interview_interested' className='button' target='_blank' rel='noopener noreferrer'>
-                        <span>{it.L('I\'m interested')}</span>
+                    <a id='interview_interested' className='button'>
+                        <span>{it.L('I’m interested')}</span>
```
negarn: and here

src/templates/static/_layout/layout.jsx
```diff
@@ -54,7 +54,7 @@ const Layout = () => {
                         </div>
                         <Footer />
                     </div>
-                    <InterviewPopup /> {/* TODO: remove when UX research is finished */}
+                    <InterviewPopup />
```
negarn: let's keep the comments for these two, we still want to remove them when the UX research is finished.

src/javascript/app/base/page.js
cakasuma: i'll move it to `!Login.isLoginPages` if bracket

PR #5131, opened 2019-03-10
arash-binary/br_error-message-dismiss-too-fast

arash-binary:
When updating customer profile with duplicate same last name, first name and DOB, it throw error.


src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -375,7 +375,7 @@ const PersonalDetails = (() => {
             .attr('class', is_success ? 'success-msg' : 'errorfield')
             .html(is_success ? $ul : localized_text)
             .css('display', 'block')
-            .delay(5000)
```
negarn: what about the other pages where we delay by 5000? should we update those to 15k too?

arash-binary: we have 2 other pages with delay(5000). I check every error message for these two pages.
1- financial_assessment.js the longest message is: 'Sorry, an error occurred while processing your request.' I think 5000 delay is ok for that.
2- self_exclusion.js the longest message is: 'When you click "OK" you will be excluded from trading on the site until the selected date.' I change the delay time to 10000.

PR #5132, opened 2019-03-10
easteregg/speed-up-svg-test

easteregg:


scripts/__tests__/svg_test.js
```diff
@@ -1,39 +1,37 @@
...
 
-            return arr;
-        };
-
-        const svgs = fetchSvgs(path.resolve(common.root_path, 'src/images'));
+    it('should be valid svgs', async () => {
+        try {
+            changed_files = [
+                ...await fetchFiles('git diff --name-only -- *.svg'),
+                ...await fetchFiles('git diff HEAD origin/master --name-only -- *.svg'),
```
negarn: does this mean you constantly need to make sure that origin/master is up to date with upstream/master? mine never is, I can't imagine everyone else remembers to always update theirs either

easteregg: No, you don't need to keep your branch updated. So, we are looking for changed files, if nothing is changed since x commits ago, then we are good to go 👍.

easteregg: In a situation that origin had a file, but the dev's branch did not include that one, an error popped up. the new commit fixes that.

PR #5140, opened 2019-03-11
amam/themify_components

cakasuma:
components need to be themified
- account switcher
- account info
- datepicker
- icons
- fullpage modal
- rangeslider
- button-menu

this also
- fix svg icons
- add signup button
- fix misplaced positions
- add missing icons
- fix missing validation for minimum date for duration


src/sass/app_2/_common/base/constants.scss
```diff
@@ -33,6 +34,7 @@ $COLOR_DARK_BLUE_6         : #555975;
 $COLOR_DARK_BLUE_7         : #7f8397;
 $COLOR_DARK_BLUE_8         : rgba(127, 131, 151, 0.3);
 $COLOR_DARK_GREEN_GRADIENT : linear-gradient(to top, rgba(16, 19, 32, 0), rgba(0, 148, 117, 0.16));
+$COLOR_DARK_TRANSPARENT    : rgba(31, 36, 61, 0.8);
```
ashkanx: @cakasuma I don't see these colours in Zeplin style guide.

cakasuma: yes, i was not aware of the `transparentize mixin`. i'll fix it

src/sass/app_2/_common/components/account-switcher.scss
```diff
@@ -101,40 +113,77 @@
         &-title {
             @include typeface(--small-center-normal-grey, none);
             @include toEm(padding, 8px 16px, 1.2em);
-            border: 1px dashed $COLOR_LIGHT_BLACK_2;
             width: 100%;
             border-radius: 4px;
             cursor: pointer;
+            @include themify($themes) {
```
kellybinary: @cakasuma just curious, do we have to include `themify` for individual classes? since`themify` was already added to the top-most parent (the `.body`), shouldn't using `themed('color_var')` suffice to change the colors (without having to include `themify` again)?

cakasuma: @kellybinary themify works as a mixin to produce the current selector with the object themes provided. basically themify function will add the `theme--dark` class plus the current selector. and then themed('color_var'). and yes, we can put the @include themify to the most parent for example here `.acc-switcher` and just use themed('color_var') all the way to its child. however. if we do that, the file of css will be very high since it will doubled the styling.

kellybinary: you mean, it's better to use `@include themify` for all parent?

cakasuma: i mean, it can be used for all parent. however it will doubled the css yes. that is why i believe its better to put it under individual class.

src/sass/app_2/_common/base/common.scss
```diff
@@ -1,3 +1,13 @@
+/** @define theme */
+.theme {
+    &--light {
```
cakasuma: since this is the most top parent, i guess there is no other way other than put the color manually

src/sass/app_2/_common/components/account-switcher.scss
```diff
@@ -101,40 +113,77 @@
...
+                }
+                &--selected {
+                    background: themed('active_color');
+
+                    .acc-switcher__id {
+                        color: themed('text_active_color');
+                        font-weight: 700;
+
+                        &--virtual:before {
+                            color: $COLOR_WHITE;
```
kellybinary: i think this can be `color: themed('text_active_color');`

src/sass/app_2/_common/components/account-switcher.scss
```diff
@@ -101,40 +113,77 @@
...
+                            border: 1px solid $COLOR_WHITE;
```
kellybinary: `border: 1px solid themed('text_active_color')`?

PR #5141, opened 2019-03-11
Nazanin/forward_starting

nazaninreihani:
Add popup when symbol has only forward starting type


src/javascript/app_2/Stores/Modules/Trading/Helpers/contract-type.js
```diff
@@ -25,8 +25,11 @@ const ContractType = (() => {
     let available_categories     = {};
     let contract_types;
     const trading_times          = {};
+    let is_only_forwarding = false;
```
OskarAhl: Perhaps `has_only_forward_starting_contracts` would be more suitable?

src/javascript/app_2/Stores/Modules/Trading/trade-store.js
```diff
@@ -334,13 +334,20 @@ export default class TradeStore extends BaseStore {
...
         }
 
+        let has_only_forward_starting_contracts;
+
+        if (/symbol/.test(Object.keys(obj_new_values))) {
+            await Symbol.onChangeSymbolAsync(obj_new_values.symbol);
+            has_only_forward_starting_contracts =
+                ContractType.getContractCategories().has_only_forward_starting_contracts;
+        }
+        this.root_store.ui.setHasOnlyForwardingContracts(has_only_forward_starting_contracts);
```
OskarAhl: LGTM, think we can add a `// TODO: remove this once future contracts are implemented`?  I'm a bit concerned that this couples the trade-store with the ui-store but I can't think of a better way 😄

nazaninreihani: yes I'll add a comment

PR #5142, opened 2019-03-12
msamprz/update-open-roles

msamprz:
Updated open roles to add `Data Scientist` under the new `Data Analytics` section


src/templates/static/about/job_details.jsx
```diff
@@ -94,6 +96,13 @@ const JobDetails = () => (
                             ]}
                         />
 
+                        <SideBar
+                            className='data_analytics'
+                            sidebar_items={[
+                                { href: 'data_scientist',      string: it.L('Data Scientist') },
```
negarn: can you fix the spacing here too?

src/templates/static/about/job_details.jsx
```diff
@@ -587,6 +596,67 @@ const JobDetails = () => (
...
+                        <div className='invisible' id='data_scientist'>
+                            <h1>{it.L('Data Scientist')}</h1>
+
+                            <p>{it.L('We need a talented Data Scientist with the technical skills and creative curiosity to explore and solve complex business problems.')}</p>
+
+                            <UlText
+                                text={it.L('You should be passionate about:')}
+                                className='bullet'
+                                items={[
+                                    { text: it.L('Focusing on customer’s needs in everything that you do.') },
```
negarn: can you use `\'` instead of `’`?

PR #5148, opened 2019-03-12
ahkuma/Disable ea4eu

ahkuma:


src/javascript/app/pages/user/metatrader/metatrader.ui.js
```diff
@@ -638,6 +638,10 @@ const MetaTraderUI = (() => {
         }
     };
 
+    const showEA = (is_shown) => {
```
negarn: I prefer `should_show`, what do you think?

src/javascript/static/pages/get_started.js
```diff
@@ -22,7 +24,19 @@ module.exports = {
...
-        onLoad  : () => { MenuSelector.init(['what-are-binary-options', 'how-to-trade-binary', 'types-of-trades']); },
+        onLoad: () => {
+            let menu_sections = [
+                'what-are-binary-options',
+                'how-to-trade-binary',
+                'types-of-trades',
+            ];
+            const landing_company_shortcode = ClientBase.get('landing_company_shortcode');
+            if (landing_company_shortcode === 'malta' || landing_company_shortcode === 'maltainvest') {
+                getElementById('how-to-trade-binary').setVisibility(false);
```
negarn: better to use `0` instead of `false` to be consistent with the rest of the code

src/javascript/static/pages/get_started.js
```diff
@@ -22,7 +24,19 @@ module.exports = {
...
+                menu_sections = menu_sections.filter(menu_item => menu_item !== 'how-to-trade-binary');
```
negarn: is this better, or to push how-to-trade-binary into the array in else?

src/templates/app/user/metatrader.jsx
```diff
@@ -156,7 +156,7 @@ const Metatrader = () => (
                         <div id='account_desc' className='mt-container border-bottom' />
                     </div>
                     <div className='mt-panel'>
-                        <div className='mt-sidebar-button border-bottom'>
+                        <div className='mt-sidebar-button border-bottom' id='mt-ea'>
```
negarn: we have a property data-hide, you can use it to hide this section for malta and maltainvest, it doesn't work in this case?

src/javascript/app/pages/user/metatrader/metatrader.ui.js
ahkuma: tbh, this thing is just a one liner, do we really need this? I think should_show rolls of the tongue better.

src/javascript/static/pages/get_started.js
ahkuma: I think this is better than pushing it in else since we remove one logical point we need to think of when reading the code. One thing that bothers me, is it not very obvious what it is trying to do, I wish JS have a `.remove` function that takes a value and remove it just like in Python, would make it more readable, but I still like this better because I can just think, 

`if landingcompany is malta or maltainvest, remove this section` instead of `if landingcompany is malta or maltainvest, remove this part otherwise add this part`

src/javascript/app/pages/user/metatrader/metatrader.ui.js
negarn: yes, better keep the function since that's the behaviour in metatrader to hide/show elements in .ui

src/javascript/static/pages/get_started.js
```diff
@@ -22,7 +25,20 @@ module.exports = {
...
     BinaryOptionsForMT5: {
-        onLoad  : () => { MenuSelector.init(['what-are-binary-options', 'how-to-trade-binary', 'types-of-trades']); },
+        onLoad: () => {
+            let menu_sections = [
+                'what-are-binary-options',
+                'how-to-trade-binary',
+                'types-of-trades',
+            ];
+            BinarySocket.wait('authorize', 'website_status', 'landing_company').then(() => {
+                if (isEuCountry()) {
```
kellybinary: do you need this check? think it's redundant with the usage of `data-show='-eucountry'` 🤔

ahkuma: Yes, it still need to check for the `next` and `previous` on the bottom to remove it, otherwise when user press next, they will go to the `how-to-binary-mt5` page.

PR #5153, opened 2019-03-12
easteregg/fix-substr-of-undefined

easteregg:


src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -126,8 +126,10 @@ const DigitDisplay = (() => {
 
     const end = (proposal_open_contract) => {
         if (proposal_open_contract.status !== 'open') {
+            // if there is no exit tick inside proposal open contract, select a fallback from history instead.
+            const fallback_exit_tick = spot_times.find(spot => spot.time === proposal_open_contract.exit_tick_time);
```
OskarAhl: Will this need the number coercion for when BE sends strings instead of numbers? 
`+spot.time === +proposal_open_contract.exit_tick_time`

easteregg: Yeah, why not. more safe. I'll push another commit to address it.

PR #5165, opened 2019-03-14
Nazanin/trading_times

nazaninreihani:
show notice when there is no trading times


src/templates/app/resources/trading_times.jsx
```diff
@@ -3,15 +3,18 @@ import React from 'react';
...
             <input type='text' id='trading-date' readOnly='readonly' size='20' />
         </p>
-        <p>{it.L('All times are in GMT (Greenwich Mean Time).')}</p>
+        <p id='trading-date-notice'>{it.L('All times are in GMT (Greenwich Mean Time).')}</p>
 
         <div className='gr-padding-10'>
             <p className='error-msg invisible' id='errorMsg' />
             <div id='trading-times' className='has-tabs gr-parent' />
+            <p className='notice-msg center-text invisible' id='empty-trading-times'>
+                {it.L('Trading times is unavailable in this country. If you have an active [_1] account, please log in for full access.', it.website_name)}
```
kellybinary: do we need to add link to `log in`?

ashkanx: yes

PR #5166, opened 2019-03-14
kelly/fix

kellybinary:
- update symbol in `trade_store` when selecting an active position in position drawer
- add `getQueryString` method in `url-helper.js`
- store query string in `trade_store` properly
- refactor allow-equals logic from `trade_store` to it's own component


src/javascript/app_2/Utils/URL/url-helper.js
```diff
@@ -32,12 +31,13 @@ export default class URLHelper {
 
             const value = params[name];
 
+            // console.log(name, value);
```
OskarAhl: Think this can be removed

src/javascript/app_2/Utils/URL/url-helper.js
```diff
@@ -10,8 +10,7 @@ export default class URLHelper {
      */
     static getQueryParams(url) {
         const query_string =  url ? new URL(url).search : window.location.search;
-        const query_encoded = encodeURIComponent(query_string);
```
OskarAhl: What was the reason for the encode again? @cakasuma

src/javascript/app_2/Stores/Modules/Trading/trade-store.js
```diff
@@ -97,6 +97,9 @@ export default class TradeStore extends BaseStore {
     @observable proposal_info        = {};
     @observable purchase_info        = {};
 
+    // Query string
+    @observable query = '';
```
OskarAhl: What's the reason for adding the query as an observable?

src/javascript/app_2/Utils/URL/url-helper.js
kellybinary: @OskarAhl @cakasuma thanks, i've added a fix

PR #5184, opened 2019-03-19
[new-app] zaki/trade_chart_updates

zakibinary:
Changelog
- Refactored the portfolio store, removed redundant object responses and only set the object response that have modifiers.
- Add leftscroll epoch and leftscroll offset, as well as zoom props in smart-charts store.
- Switch granularity to `0`, set `purchase_time` as `start_epoch` and zoom in for `tick` contracts.
- Fixed bug with account switching causing symbol to set to default i.e - `AUD/JPY` instead.


src/javascript/app_2/Modules/SmartChart/Containers/smart-chart.jsx
```diff
@@ -41,7 +41,7 @@ class Chart extends React.Component {
             <SmartChart
                 barriers={this.props.barriers_array}
                 bottomWidgets={this.props.should_show_last_digit_stats ? undefined : this.bottomWidgets}
-                chartControlsWidgets={this.chartControlsWidgets}
+                chartControlsWidgets={this.props.is_contract_mode ? null : this.chartControlsWidgets}
```
zakibinary: Added this here but it wouldn't work until SmartCharts fixes the issue where it's overriding `null` passed to the `chartControlsWidgets` prop and still showing the chart controls.

src/javascript/app_2/Stores/Modules/Contract/contract-store.js
```diff
@@ -41,52 +42,50 @@ export default class ContractStore extends BaseStore {
     @observable has_error         = false;
     @observable error_message     = '';
     @observable is_sell_requested = false;
+    @observable is_left_epoch_set = false;
```
OskarAhl: Could this be moved to the smart-chart-store? Since it seems related to the chart.

src/javascript/app_2/Stores/Modules/Contract/contract-store.js
```diff
@@ -100,14 +99,17 @@ export default class ContractStore extends BaseStore {
...
         this.sell_info         = {};
         this.is_sell_requested = false;
         this.chart_config      = {};
+        this.is_left_epoch_set = false;
 
         destroyChartTickMarkers();
         this.smart_chart.removeBarriers();
         this.smart_chart.removeMarkers();
+        this.smart_chart.resetScrollZoom();
         this.smart_chart.setContractMode(false);
```
OskarAhl: Could these all be put in a `this.smart_chart.cleanupContractChartView()` or similar?

src/javascript/app_2/Stores/Modules/Portfolio/portfolio-store.js
```diff
@@ -114,7 +106,7 @@ export default class PortfolioStore extends BaseStore {
     @action.bound
     onClickSell(contract_id) {
         const i = this.getPositionIndexById(contract_id);
-        const bid_price = this.positions[i].bid_price;
+        const bid_price = this.positions[i].contract_info.bid_price;
```
OskarAhl: nit: could be `const { bid_price } = this.positions[i].contract_info;`

src/javascript/app_2/Stores/Modules/Contract/contract-store.js
zakibinary: It's related to SmartCharts but the variable is only relevant for when the contract is running.

PR #5190, opened 2019-03-20
amam/revert_interview

cakasuma:
*Merge this PR if we want to disable the interview popup dialog


src/javascript/app/base/page.js
```diff
@@ -82,7 +81,8 @@ const Page = (() => {
                 Language.setCookie(Language.urlLang());
 
                 if (!ClientBase.get('is_virtual')) {
-                    InterviewPopup.onLoad();
+                    // TODO: uncomment below to enable interview popup dialog
```
ashkanx: should we just remove this?

src/javascript/app/pages/user/reality_check/reality_check.ui.js
```diff
@@ -101,7 +100,8 @@ const RealityCheckUI = (() => {
     const closePopUp = () => {
         $('#reality_check').remove();
         startSummaryTimer();
-        InterviewPopup.onLoad();
+        // TODO: uncomment below to enable interview popup dialog
```
ashkanx: same here

negarn: I think let's keep it, we will use it in the future, same as before

PR #5196, opened 2019-03-20
ashkan/add country_ip to datalayer

ashkanx:


src/javascript/_common/base/gtm.js
```diff
@@ -12,13 +12,14 @@ const getPropertyValue = require('../utility').getPropertyValue;
 const getAppId         = require('../../config').getAppId;
 
 const GTM = (() => {
-    const isGtmApplicable = () => (/^(1|1098|14473|15284)$/.test(getAppId()));
+    const isGtmApplicable = () => (/^(1|1098|14473|15284|1108)$/.test(getAppId()));
```
ashkanx: I should remove this before merging the PR

src/javascript/_common/base/gtm.js
```diff
@@ -12,13 +12,14 @@ const getPropertyValue = require('../utility').getPropertyValue;
...
     const getCommonVariables = () => ({
-        language : getLanguage(),
-        pageTitle: pageTitle(),
-        pjax     : State.get('is_loaded_by_pjax'),
-        url      : document.URL,
+        language  : getLanguage(),
+        pageTitle : pageTitle(),
+        pjax      : State.get('is_loaded_by_pjax'),
+        url       : document.URL,
+        country_ip: State.getResponse('website_status.clients_country'),
```
negarn: please sort it alphabetically similar to above lines, line 22 should be on line 18

negarn: not sure if it is guaranteed that we will have website_status here, since we're not able to explicitly wait for website_status. we can expect some times this can be empty, if the race condition hits. it's unlikely, but seems possible from the code, for page_load event etc

ashkanx: I tested it several times and it didn't happen. although it really doesn't matter if it's randomly empty. we can handle that in GTM

src/javascript/_common/base/gtm.js
negarn: please remember to remove this

PR #5212, opened 2019-03-25
ahkuma/poidwed

ahkuma:


src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -39,6 +39,11 @@ const Authenticate = (() => {
                         link = Url.urlForCurrentDomain('https://marketing.binary.com/authentication/MF_Authentication_Process.pdf');
                     }
                     $not_authenticated.find('.learn_more').setVisibility(1).find('a').attr('href', link);
+
+                    if (isIdentificationNoExpiry(Client.get('residence'))) {
+                        $('#expiry_datepicker_proofid').setVisibility(0);
+                        $('#exp_date_2').datepicker('setDate', '2099-12-31');
```
negarn: does this work for mobile too? or just desktop?

ahkuma: @negarn It should. Seems like it use the same jQuery date picker for both.

negarn: @ahkuma please test it. in mobile we use the native datepicker.

ahkuma: Sorry, I should have said, I tested it before and it works fine. I tested it again using both the mac and my phone and it works fine.

PR #5213, opened 2019-03-25
oskar/markers_from_proposal_ticks_arr

OskarAhl:
* Add ongoing / finished contract tick markers from new `ticks_stream` proposal_open_contract response
* Remove old chart-tick-markers.js file


src/javascript/_common/utility.js
```diff
@@ -229,6 +233,7 @@ module.exports = {
     template,
     isEmptyObject,
     cloneObject,
+    unique,
```
negarn: I think this should be on line 237, we seem to be sorting by function placement in this file

PR #5219, opened 2019-03-26
easteregg/platform-settings-assets

easteregg:


src/javascript/app_2/App/Containers/SettingsDialog/settings-chart.jsx
```diff
@@ -1,19 +1,32 @@
...
+import ChartPositionEnabledLight     from 'Images/app_2/settings/bottom.svg';
+import OHLCDisabledDark              from 'Images/app_2/settings/dark/OHLC-disabled.svg';
+import OHLCEnabledDark               from 'Images/app_2/settings/dark/OHLC-enabled.svg';
+import ChartPositionEnabledDark      from 'Images/app_2/settings/dark/bottom.svg';
+import IntervalDurationDisabledDark  from 'Images/app_2/settings/dark/interval-disabled.svg';
+import IntervalDurationEnabledDark   from 'Images/app_2/settings/dark/interval-enabled.svg';
+import ChartPositionDisabledDark     from 'Images/app_2/settings/dark/left.svg';
+import IntervalDurationDisabledLight from 'Images/app_2/settings/interval-disabled.svg';
+import IntervalDurationEnabledLight  from 'Images/app_2/settings/interval-enabled.svg';
+import ChartPositionDisabledLight    from 'Images/app_2/settings/left.svg';
```
OskarAhl: Would it be worth naming these as `...Icon`? e.g. ChartPositionDisabledDarkIcon

src/javascript/app_2/App/Components/Elements/Media/media-icon.jsx
```diff
@@ -1,7 +1,15 @@
...
-    <div className='media__icon' />
-);
+const MediaIcon = ({ is_enabled, enabled, disabled }) => {
+    const Icon = is_enabled ? enabled : disabled;
+    return <Icon className='media__icon' />;
+};
+
+MediaIcon.propTypes = {
+    disabled  : PropTypes.func,
+    enabled   : PropTypes.func,
```
OskarAhl: Are `enabled` and `disabled` functions, looks like they are svg elements?

src/javascript/app_2/App/Containers/SettingsDialog/settings-chart.jsx
easteregg: Yeap. Fine by me. Will do.

src/javascript/app_2/App/Components/Elements/Media/media-icon.jsx
easteregg: imported svg elements only work as functions, tried `node`, `element` and `string` nothing else works on them.

OskarAhl: strange, `PropTypes.instanceOf(Element)` does not work? If it doesn't work perhaps we can add a comment that mentions that this component expects svg elements as parameters?

PR #5223, opened 2019-03-27
add validation for malta shortcode landing company

cakasuma:
- check redirect url
- append for malta landing company shortcode
- redirect to self exclusion after set currency

### DIFF
--- a/src/javascript/app/pages/user/set_currency.js
+++ b/src/javascript/app/pages/user/set_currency.js
@@ -85,6 +85,10 @@ const SetCurrency = (() => {
                                         redirect_url = Url.urlFor('user/authenticate');
                                     }
                                 }
+                                // Do not redirect MF clients to cashier, because they need to set self exclusion before trading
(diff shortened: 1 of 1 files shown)


src/javascript/app/pages/user/set_currency.js
```diff
@@ -85,6 +85,10 @@ const SetCurrency = (() => {
                                         redirect_url = Url.urlFor('user/authenticate');
                                     }
                                 }
+                                // Do not redirect MF clients to cashier, because they need to set self exclusion before trading
+                                if (!redirect_url && !/^(malta)$/i.test(Client.get('landing_company_shortcode'))) {
```
negarn: have you tested this? as per the code IMO this won't work, on line 82 we are deciding for financial clients (MF) send them to authenticate page, and here we want to say that send gaming (MLT) clients to self exclusion instead, but in the comment you are mentioning `MF` instead of `MLT` and also you are checking that shortcode is not equal to `malta` while it should be equal to `malta`

PR #5234, opened 2019-03-28
msamprz/change-currency

msamprz:


src/javascript/app/pages/cashier/cashier.js
```diff
@@ -65,14 +69,86 @@ const Cashier = (() => {
...
+
+        const currency_hint       = isCryptocurrency(currency)
+            ? localize('Don\'t want to trade in [_1]? You can open another cryptocurrency account. [_2]Manage your accounts[_3].', [`${currency}`, `<a href=${urlFor('user/accounts')}>`, '</a>'])
+            : has_no_mt5 && has_no_tx
+                ? localize('You can set a new currency before you deposit for the first time or create an MT5 account. [_1]Manage your accounts[_2].', [`<a href=${urlFor('user/accounts')}>`, '</a>'])
+                : missingCriteria(!has_no_mt5, !has_no_tx);
+
+        elementInnerHtml(el_current_currency, currency_message);
+        elementInnerHtml(el_current_hint, currency_hint);
+        el_currency_image.src = `/images/pages/cashier/icons/icon-${currency}.svg`;
```
kellybinary: @msamprz you can use the currencies icon from `images/pages/set_currency` instead of adding the new ones (unless they are different), they looked the same to me 😄

msamprz: Due to inconsistency of the shape of icons of the new set vs the ones in `images/pages/set_currency`, this will be addressed by adding a wrapper to the image to be able to scale the image (so nothing is cropped) and have a circular window of it.

To avoid this:
image

And look like this (only issue with this is that the image will be scaled):
image

kellybinary: okay, let's keep these new icons then 👍

src/javascript/app/pages/cashier/cashier.js
```diff
@@ -1,10 +1,14 @@
+/* eslint-disable no-nested-ternary */
```
negarn: would prefer not disabling this

src/javascript/app/pages/cashier/cashier.js
```diff
@@ -1,10 +1,14 @@
...
 const Client           = require('../../base/client');
 const BinarySocket     = require('../../base/socket');
 const isCryptocurrency = require('../../common/currency').isCryptocurrency;
+const elementInnerHtml = require('../../../_common/common_functions').elementInnerHtml;
 const getElementById   = require('../../../_common/common_functions').getElementById;
 const localize         = require('../../../_common/localize').localize;
+const State            = require('../../../_common/storage').State;
 const paramsHash       = require('../../../_common/url').paramsHash;
 const urlFor           = require('../../../_common/url').urlFor;
+const urlForStatic     = require('../../../_common/url').urlForStatic;
```
negarn: may be worth importing the whole url file and calling the functions since we are using three of them?

src/javascript/app/pages/cashier/cashier.js
```diff
@@ -65,14 +69,60 @@ const Cashier = (() => {
...
     };
 
+    const showCurrentCurrency = (currency, has_no_mt5, has_no_tx) => {
+        const el_acc_currency     = getElementById('account_currency');
+        const el_currency_image   = getElementById('account_currency_img');
+        const el_current_currency = getElementById('account_currency_current');
+        const el_current_hint     = getElementById('account_currency_hint');
+
+        const missingCriteria     = (has_mt5, has_tx) => {
+            const existing_mt5_msg = localize('You can no longer change the currency because you’ve created an MT5 account. [_1]Manage your accounts[_2].', [`<a href=${urlFor('user/accounts')}>`, '</a>']);
```
negarn: should we use `you have` instead?

src/javascript/app/pages/cashier/cashier.js
```diff
@@ -65,14 +69,60 @@ const Cashier = (() => {
...
+            const existing_tx_msg  = localize('You can no longer change the currency because you’ve made a first-time deposit. [_1]Manage your accounts[_2].', [`<a href=${urlFor('user/accounts')}>`, '</a>']);
+
+            return !has_mt5 && has_tx ? existing_tx_msg : (has_mt5 && !has_tx ? existing_mt5_msg : existing_tx_msg);
+        };
+
+        // Set messages based on if currency is crypto or fiat
```
negarn: can we change it to `Set messages based on currency being crypto or fiat`

src/javascript/app/pages/cashier/cashier.js
```diff
@@ -65,14 +69,60 @@ const Cashier = (() => {
...
+        const el_acc_currency     = getElementById('account_currency');
+        const el_currency_image   = getElementById('account_currency_img');
+        const el_current_currency = getElementById('account_currency_current');
+        const el_current_hint     = getElementById('account_currency_hint');
+
+        const missingCriteria     = (has_mt5, has_tx) => {
+            const existing_mt5_msg = localize('You can no longer change the currency because you’ve created an MT5 account. [_1]Manage your accounts[_2].', [`<a href=${urlFor('user/accounts')}>`, '</a>']);
+            const existing_tx_msg  = localize('You can no longer change the currency because you’ve made a first-time deposit. [_1]Manage your accounts[_2].', [`<a href=${urlFor('user/accounts')}>`, '</a>']);
+
+            return !has_mt5 && has_tx ? existing_tx_msg : (has_mt5 && !has_tx ? existing_mt5_msg : existing_tx_msg);
```
negarn: can this be simplified to below? seems to do the same
```
return has_mt5 ? existing_mt5_msg : existing_tx_msg;
```

src/javascript/app/pages/cashier/cashier.js
```diff
@@ -65,14 +69,60 @@ const Cashier = (() => {
...
+        };
+
+        // Set messages based on if currency is crypto or fiat
+        // If fiat, set message based on if they're allowed to change currency or not
+        // Condition is to have no MT5 accounts *and* have no transactions
+        const currency_message    = isCryptocurrency(currency)
```
negarn: extra space before `=`

src/javascript/app/pages/cashier/cashier.js
```diff
@@ -65,14 +69,60 @@ const Cashier = (() => {
         });
     };
 
+    const showCurrentCurrency = (currency, has_no_mt5, has_no_tx) => {
+        const el_acc_currency     = getElementById('account_currency');
+        const el_currency_image   = getElementById('account_currency_img');
+        const el_current_currency = getElementById('account_currency_current');
+        const el_current_hint     = getElementById('account_currency_hint');
+
+        const missingCriteria     = (has_mt5, has_tx) => {
```
negarn: extra space before `=`, we don't need to align it with above since there's an empty line between them

src/javascript/app/pages/cashier/cashier.js
```diff
@@ -65,14 +69,60 @@ const Cashier = (() => {
...
+            return !has_mt5 && has_tx ? existing_tx_msg : (has_mt5 && !has_tx ? existing_mt5_msg : existing_tx_msg);
+        };
+
+        // Set messages based on if currency is crypto or fiat
+        // If fiat, set message based on if they're allowed to change currency or not
+        // Condition is to have no MT5 accounts *and* have no transactions
+        const currency_message    = isCryptocurrency(currency)
+            ? localize('This is your [_1] account.', `${currency}`)
+            : has_no_mt5 && has_no_tx
+                ? localize('Your fiat account’s currency is currently set to [_1].', `${currency}`)
```
negarn: please use a normal `'` with a `\`

src/javascript/app/pages/cashier/cashier.js
```diff
@@ -65,14 +69,60 @@ const Cashier = (() => {
...
+                : localize('Your fiat account’s currency is set to [_1].', `${currency}`);
+
+        const currency_hint       = isCryptocurrency(currency)
```
negarn: extra space before `=`

src/javascript/app/common/attach_dom/dialog.js
```diff
@@ -17,16 +17,43 @@ const Dialog = (() => {
...
+                    // Used for setting the dialog title, message, and footnote
+                    const setMessage = (option_item, $selector) => {
+                        const item = Array.isArray(options[option_item]) ? options[option_item].join('<p />') : options[option_item];
+                        elementInnerHtml(container.querySelector($selector), item);
+                    };
 
-                    const localized_title = options.localized_title;
-                    if (localized_title && el_title) {
+                    // Set title if `localized_title` field in `options` param exists
+                    if (options.localized_title && el_title) {
```
OskarAhl: What about this instead of the comment? 
```suggestion
                    const has_title = options.localized_title && el_title
                    if (has_title) {
```

src/javascript/app/pages/cashier/deposit_withdraw.js
```diff
@@ -202,6 +203,19 @@ const DepositWithdraw = (() => {
         });
     };
 
+    const canChangeCurrency = () => {
+        const statement  = State.getResponse('statement');
+        const has_no_mt5 = State.getResponse('mt5_login_list').length === 0;
+        const has_no_tx  = (statement.count === 0 && statement.transactions.length === 0);
+
+        // Current BE requirements for user successfully changing their account's currency:
```
negarn: please use API instead of BE

src/javascript/app/common/attach_dom/dialog.js
```diff
@@ -17,16 +17,43 @@ const Dialog = (() => {
...
                         el_title.setVisibility(1);
-                        elementInnerHtml(container.querySelector('#dialog_title'), localized_title);
+                        setMessage('localized_title', '#dialog_title');
+                    }
+
+                    // Set dialog message
+                    setMessage('localized_message', '#dialog_message');
+
+                    // Set footnote if `localized_footnote` field in `options` param exists
+                    if (options.localized_footnote && el_footnote) {
```
OskarAhl: Same could be added here

src/javascript/app/pages/user/accounts.js
```diff
@@ -71,6 +80,20 @@ const Accounts = (() => {
...
+        const has_no_mt5 = State.getResponse('mt5_login_list').length === 0;
+        const has_no_tx  = (statement.count === 0 && statement.transactions.length === 0);
+        const currency    = Client.get('currency');
+
+        // Current BE requirements for user successfully changing their account's currency:
+        // 1. User must not have made any transactions
+        // 2. User must not have any MT5 account
+        // 3. Not be a crypto account
+        // 4. Not be a virtual account
+        return !Client.get('is_virtual') && has_no_tx && has_no_mt5 && !isCryptocurrency(currency);
```
negarn: are we repeating this? can we use a common function?

src/javascript/app/pages/user/accounts.js
```diff
@@ -95,6 +118,90 @@ const Accounts = (() => {
                         .html($('<span/>', { text: localize('Create') })))));
     };
 
+    const addChangeCurrencyOption = () => {
+        const table_headers        = TableHeaders.get();
+        const loginid              = Client.get('loginid');
```
negarn: extra space before `=`

src/javascript/app/pages/user/accounts.js
```diff
@@ -244,9 +337,12 @@ const Accounts = (() => {
         }
     };
 
-    const showError = (localized_text) => {
-        $('#new_account_error').remove();
-        $('#new_account_opening').find('button').parent().append($('<p/>', { class: 'error-msg', id: 'new_account_error', text: localized_text }));
+    const showError = (localized_text, is_currency_change = false) => {
```
OskarAhl: Instead of passing a `is_currency_change ` boolean parameter maybe passing the ids would make this function more reusable and general?

src/javascript/app/pages/user/set_currency.js
```diff
@@ -51,10 +51,15 @@ const SetCurrency = (() => {
...
             }
-            const crytpo_currencies = $cryptocurrencies.html();
-            if (crytpo_currencies) {
+            const crypto_currencies = $cryptocurrencies.html();
+            if (crypto_currencies) {
                 $('#crypto_currencies').setVisibility(1);
-                $('#crypto_currency_list').html(crytpo_currencies);
+                $('#crypto_currency_list').html(crypto_currencies);
+            }
+            if ((!fiat_currencies && crypto_currencies) || (fiat_currencies && !crypto_currencies)) {
```
OskarAhl: I think we could also add this to a constant? perhaps `const has_currency_selected = ...`

PR #5239, opened 2019-03-28
Nazanin/Mga disclosure requirements

nazaninreihani:


src/templates/app/_layout/footer.jsx
```diff
@@ -181,6 +184,9 @@ const Footer = () => (
                                     <a className='lga-icon' href='https://www.authorisation.mga.org.mt/verification.aspx?lang=EN&company=a5fd1edc-d072-4c26-b0cd-ab3fa0f0cc40&details=1' target='_blank' rel='noopener noreferrer'>
                                         <img className='responsive' src={it.url_for('images/pages/footer/mga-logo-footer.svg')} />
                                     </a>
+                                    <div className='age-restriction'>
```
kellybinary: i think you can move this icon outside of `lga-gamstop-icon-container`

nazaninreihani: yes I just found out what it looks like for EUs :-D

kellybinary: hehe, i wonder why it's named `lga` instead of `mga` tho.. 🤔

nazaninreihani: I think we can change that, it was supposed to be mga probably

kellybinary: that'd be great, thank you :)

PR #5246, opened 2019-04-01
negar/expect_fields

negarn:
get the error fields correctly


src/javascript/app/pages/cashier/deposit_withdraw.js
```diff
@@ -162,20 +162,22 @@ const DepositWithdraw = (() => {
...
+                address_state   : localize('State/Province'),
+                email           : localize('Email address'),
+                phone           : localize('Telephone'),
+                residence       : localize('Country of Residence'),
             };
+            details_fields = details.fields.map(field => (error_fields[field] || field));
         }
         const $el     = $(`#${msg_id}`);
-        const err_msg = template($el.html(), [details ? error_fields[details] : localize('details')]);
+        const err_msg = template($el.html(), [details_fields ? details_fields.join(', ') : localize('details')]);
```
Anthony-Wong: Is it fine to be showing a message like `Please update your phone, email here.` grammar wise?

negarn: IMO yes, we do it in other places as well. we can't really translate the `and` in the middle.

PR #5248, opened 2019-04-01
kelly/prompt to install PWA in notification bar

kellybinary:
- prompt to install PWA in notification bar


src/javascript/app_2/Stores/ui-store.js
```diff
@@ -243,4 +237,16 @@ export default class UIStore extends BaseStore {
     setHasOnlyForwardingContracts(has_only_forward_starting_contracts) {
         this.has_only_forward_starting_contracts = has_only_forward_starting_contracts;
     }
+
+    @action.bound
+    addNotificationBar(message) {
+        if (!this.push_notifications.length) {
+            this.push_notifications.push(message);
+        }
+        this.push_notifications.map(notification => {
```
nazaninreihani: Isn't it going to make duplicates in push_notifications array?

src/javascript/app_2/Stores/ui-store.js
```diff
@@ -243,4 +237,16 @@ export default class UIStore extends BaseStore {
...
+            if (notification.msg_type !== message.msg_type) {
```
negarn: I think to avoid duplicates you need to do a find instead, since in mapping if the message.msg_type of the first notification is different but the second one is the same, it never reaches the check for the second one and just pushes it into the array after the first check passes.

src/javascript/app_2/Stores/ui-store.js
kellybinary: true, idk what i was thinking 😅

PR #5264, opened 2019-04-02
Handle age restriction sign's visibility

nazaninreihani:


src/javascript/app/base/footer.js
```diff
@@ -15,8 +15,11 @@ const Footer = (() => {
             if (Client.isLoggedIn()) {
                 showWarning((Client.get('landing_company_shortcode') === 'maltainvest' ||
                     (Client.get('is_virtual') && State.getResponse('landing_company.financial_company.shortcode') === 'maltainvest')));
+                showAgeRestrictionSign(/iom|maltainvest/.test(Client.get('landing_company_shortcode')) ||
+                    (Client.get('is_virtual') && /iom|maltainvest/.test(Client.get('landing_company_shortcode'))));
```
negarn: I don't see the difference, either way we are checking `/iom|maltainvest/.test(Client.get('landing_company_shortcode'))`, right? why do we need to check `is_virtual` then?

src/javascript/app/base/footer.js
```diff
@@ -15,8 +15,11 @@ const Footer = (() => {
...
             } else {
                 showWarning(isEuCountry());
+                showAgeRestrictionSign(isEuCountry());
```
negarn: we are duplicating `isEuCountry()`, better store it in a variable inside this `else`

src/javascript/app/base/footer.js
```diff
@@ -25,6 +28,10 @@ const Footer = (() => {
         $('#footer-regulatory .eu-only').setVisibility(should_show_warning);
     };
 
+    const showAgeRestrictionSign = (should_show_age_restriction_sign) => {
```
negarn: this doesn't always show, it sometimes hides. it's better to update the name, something like `showHideAgeRestrictionSign` or `setVisibilityAgeRestrictionSign` etc

src/javascript/app/base/footer.js
```diff
@@ -25,6 +28,10 @@ const Footer = (() => {
...
+        $('#footer-regulatory .age-restriction').setVisibility(should_show_age_restriction_sign);
```
negarn: does it need to be a classname? if we change it to an id, we can easily use javascript instead of jQuery for it

src/javascript/app/base/footer.js
nazaninreihani: my bad, we do not need that

src/javascript/app/base/footer.js
nazaninreihani: I prefer to not do that because we're using it twice in one file

src/javascript/app/base/footer.js
```diff
@@ -13,10 +13,14 @@ const Footer = (() => {
             // logged in virtual clients with maltainvest financial landing company or
             // logged out clients with EU IP address
             if (Client.isLoggedIn()) {
-                showWarning((Client.get('landing_company_shortcode') === 'maltainvest' ||
+                const landing_company_shortcode = Client.get('landing_company_shortcode');
```
ashkanx: @negar-binary @nazaninreihani don't we have `hide/show` logic already? 
https://github.com/binary-com/binary-static/blob/master/src/javascript/app/common/content_visibility.js

nazaninreihani: we need to check for loggedin and logged out clients to exclude maltainvest, that does not have the logic I think

PR #5284, opened 2019-04-05
khalid/minus google+

Khalidxx:
Removed google plus social link from footer and made footer a little more responsive


src/sass/_common/footer.scss
```diff
@@ -218,4 +218,24 @@
             margin-right: 0;
         }
     }
+    @media screen and (max-width: 380px) {
+        .regulation-logos {
+            .vanuatu-icon {
```
cakasuma: do we really need to add these styling ?, i thought by just removing the element of google plus, since its display flex, it will auto responsive

Khalidxx: Oh it is already responsive. This is actually for the regulatory logos. They were non-responsive for small phones even before. I'm just fixing it in the same branch.

PR #5295, opened 2019-04-08
arash/show all tabs in get started regardless of offerings

arash-binary:


negarn:
@arash-binary please resolve the conflict

PR #5298, opened 2019-04-08
negar/wait_website_status_gtm

negarn:
wait for website_status before sending page_load event


ashkanx:
@negarn this is solving the issue we discussed. I'll merge this so I can test all other tags in staging

PR #5301, opened 2019-04-09
arash/dropdown issue in microsoft edge

arash-binary:


src/javascript/app/pages/trade/process.js
```diff
@@ -301,6 +301,8 @@ const Process = (() => {
     const onExpiryTypeChange = (value) => {
         const $expiry_type    = $('#expiry_type');
         const validated_value = value && $expiry_type.find(`option[value=${value}]`).length ? value : 'duration';
+        const is_ie = false || !!document.documentMode;
```
nazaninreihani: why is the `false` in this statement?

arash-binary: I change it.

PR #5303, opened 2019-04-10
khalid/Audit issue fix

Khalidxx:
Fixed the back button issue in the audit popup for mobile


src/sass/app/sell_popup.scss
```diff
@@ -120,6 +120,12 @@
                         font-weight: normal;
                     }
                 }
+                @media screen and (max-width: 480px) {
```
cakasuma: nit, please put the media screen for the respective classs , for example view-chart class is inside of `#sell_details_chart_wrapper` .

PR #5304, opened 2019-04-10
mamali|msamprz/optimize-rerendering

msamprz:


src/javascript/app_2/App/Components/Elements/Notifications/empty-notification.jsx
```diff
@@ -1,12 +1,13 @@
 import React        from 'react';
 import { localize } from '_common/localize';
+import { Icon } from 'Assets/Common';
```
negarn: alignment of `from`

src/javascript/app_2/App/Components/Elements/PositionsDrawer/positions-drawer.jsx
```diff
@@ -5,7 +5,7 @@ import React                          from 'react';
 import { CSSTransition }              from 'react-transition-group';
 import { Scrollbars }                 from 'tt-react-custom-scrollbars';
 import { localize }                   from '_common/localize';
-import { IconMinimize }               from 'Assets/Common';
+import { Icon, IconMinimize }         from 'Assets/Common';
```
negarn: do we split this to two lines?

src/javascript/app_2/Modules/Portfolio/Components/contract-type-cell.jsx
```diff
@@ -1,12 +1,13 @@
 import React                      from 'react';
 import PropTypes                  from 'prop-types';
 import { getContractTypeDisplay } from 'Constants/contract';
+import { Icon }          from 'Assets/Common/icon.jsx';
```
zakibinary: `import` alignment here

src/javascript/app_2/Modules/Trading/Containers/trade-params.jsx
```diff
@@ -47,10 +130,61 @@ TradeParams.propTypes = {
...
+        onChangeMultiple      : modules.trade.onChangeMultiple,
+        simple_duration_unit  : ui.simple_duration_unit,
+        server_time           : common.server_time,
+        start_date            : modules.trade.start_date,
+        // validation_errors     : modules.trade.validation_errors,
+        market_open_times     : modules.trade.market_open_times,
+        // LastDigit component props
+        // is_minimized          :,
+        last_digit            : modules.trade.last_digit,
+        // onChange              : modules.trade.onChange,
```
OskarAhl: What's the reason for passing down all the mobx props through the `TradeParams` component? Can we connect the child components inside Duration, Barrier, LastDigit, Amount instead?

msamprz: @OskarAhl That is exactly the plan but the change will come in the next PR.

src/javascript/app_2/Stores/common-store.js
```diff
@@ -26,6 +26,19 @@ export default class CommonStore extends BaseStore {
         this.is_socket_opened = is_socket_opened;
     }
 
+    @action.bound
+    setNetworkStatus(status, is_online) {
+        if (this.network_status.class) {
+            this.network_status = Object.assign(
```
OskarAhl: Maybe for consistency we can use the object spread instead of assign here since we are mostly using that in other places?

msamprz: @OskarAhl They don't do the same thing though. With `Object.assign` we can keep the reference and only change the value, but if we use spread syntax, it will change the reference.

OskarAhl: Ah I see now, so the purpose is to only update the class and tooltip value - then why not update it this way?
```suggestion
            this.network_status.class = status.class;
            this.network_status.tooltip = status.tooltip;
```

msamprz: @OskarAhl Yeah, for clarity we can do that. Thanks.

PR #5305, opened 2019-04-10
[binarynex] zaki/animation_refinements

zakibinary:
Changelog
- Added animations for `full-page-modal` using `react-pose`.
- Added animation transition for `settings-dialog`.
- Added button highlight animation for `button-toggle-menu` component.
- Added Purchase button animation for hover state.
- Refactor page `blur` effect, split into use for `full` and `app` only blur.
- Fixed positions drawer jumping bug when `modal-root` has a `modal` component mounted.
- Fixed missing key prop `react` error message.


src/javascript/app_2/App/Components/Elements/FullPageModal/full-page-modal.jsx
```diff
@@ -1,54 +1,137 @@
...
+});
+
+class FullPageModal extends React.Component {
+    componentDidMount() {
+        if (this.props.is_visible) {
+            this.props.showAppBlur();
+        }
+    }
+
+    componentDidUpdate() {
```
kellybinary: is this necessary since adding the blur effect is done onMount?

zakibinary: yes, since there are instances where the prop is updated and we need to re-toggle the `blur` effect. It's mainly for the `is_virtual_account` prop check when client switches account from the accounts switcher and not the `modal`.

src/javascript/app_2/App/Components/Form/ButtonToggleMenu/button-highlight.jsx
```diff
@@ -0,0 +1,24 @@
+import PropTypes from 'prop-types';
+import React     from 'react';
+
+const Highlight = ({ left, width }) => {
+    const highlight_style = {
+        width,
+        transform                : `translate3d(${left}px, 0, 0)`,
+        'borderTopLeftRadius'    : (left === 0) ? '4px' : '0',
```
OskarAhl: Can we add a `const border_radius_size = '4px'` and use it here and below instead of `4px`?

PR #5321, opened 2019-04-12
ashkan/update AppIDs

ashkanx:


src/javascript/config.js
```diff
@@ -36,15 +36,15 @@ const getAppId = () => {
         app_id = binary_desktop_app_id;
     } else if (/staging\.binary\.com/i.test(window.location.hostname)) {
         window.localStorage.removeItem('config.default_app_id');
-        app_id = is_new_app ? 16298 : 1098;
+        app_id = is_new_app ? 16303 : 1098;
```
negarn: do we actually want a different app ID for staging? or we just keep 1098 for both app and current?

ashkanx: so let's keep it for now. but we might remove it later

PR #5322, opened 2019-04-12
Remove crypto icons on home page

ashkanx:


negarn:
@nazaninreihani no, they are being aligned by `internet_bank_transfer` so we need to keep it as is

nazaninreihani:
@negar-binary internet_bank_transfer does not have data-show, why we align data-show to it?

negarn:
@nazaninreihani misunderstood your original comment, my bad.

PR #5330, opened 2019-04-15
negar/reduce_delay

negarn:
reduce delay


src/root_files/app/index.html
```diff
@@ -49,14 +48,6 @@
...
-            function isDelayedRedirect() {
-                if(isGtmApplicable() && !redirectTimeOut) {
-                    redirectTimeOut = setTimeout(redirect, 5000);
-                } else {
-                    clearTimeout(redirectTimeOut);
-                    redirect();
-                }
-            }
             function isGtmApplicable() {
                 return /(www|staging)\.binary\.com/i.test(window.location.hostname);
```
ashkanx: How about `binary.me` and `desktop_app` handling?

negarn: desktop_app has the same URL right? I'll just add the handler for binary.me

src/root_files/app_2/index.html
```diff
@@ -49,54 +49,40 @@
...
-                if(isGtmApplicable() && !redirectTimeOut) {
-                    redirectTimeOut = setTimeout(redirect, 5000);
-                } else {
-                    clearTimeout(redirectTimeOut);
-                    redirect();
-                }
-            }
             function isGtmApplicable() {
-                return /www\.binary\.com/i.test(window.location.hostname);
+                return /(www|staging)\.binary\.(com|me)/i.test(window.location.hostname);
```
ashkanx: I think we don't need `isGtmApplicable` and ` pushEvents` as we don't wait for GTM anymore

negarn: so do we want to remove GTM from app_2/index.html completely?

PR #5338, opened 2019-04-16
amam/update_dependencies

cakasuma:
### Update packages
`High || Moderate vulnerability`
- grunt
- grunt-sass
- stylelint
- svgo

*Note: 
- added `node-sass` as new requirement for grunt-sass newer version `implementation sass is required to be passed on option`
-  also added new aliases `start` to skip test for development compile and serve
- there still left `tar` dependency of node-gyp error. we can wait until they release the `3.8.1`


build/aliases.yaml
```diff
@@ -20,6 +20,13 @@ default:
   - 'css'
   - 'js'
 
+start:
+  - 'css'
+  - 'copy'
+  - 'shell:compile_dev'
+  - 'connect:livereload'
+  - 'concurrent:serve'
```
negarn: don't we also need `'js'`?

build/webpack/plugins.js
```diff
@@ -26,11 +26,11 @@ const getPlugins = (app, grunt) => ([
...
-                cachedId                 : 'app_2',
-                dontCacheBustUrlsMatching: /\.\w{8}\./,
-                minify                   : false,
-                navigateFallback         : '',
-                filepath                 : path.resolve(PATHS.DIST, `app/${lang.toLowerCase()}/service-worker.js`),
+                cachedId                     : 'app_2',
+                dontCacheBustUrlsMatching    : /\.\w{8}\./,
+                minify                       : false,
+                navigateFallback             : '',
+                filepath                     : path.resolve(PATHS.DIST, `app/${lang.toLowerCase()}/service-worker.js`),
```
negarn: why?

package.json
```diff
@@ -63,7 +63,7 @@
     "estraverse": "4.2.0",
     "extend": "3.0.2",
     "gettext-parser": "1.3.0",
-    "grunt": "1.0.3",
+    "grunt": "^1.0.4",
```
negarn: AFAIK we don't want to use `^`, we want to use exact version

package.json
```diff
@@ -137,6 +138,7 @@
     "scriptjs": "2.5.8",
     "sinon": "7.2.2",
     "smartcharts-beta": "0.4.15",
+    "tar": "^4.4.8",
```
negarn: what is this for?

build/webpack/plugins.js
cakasuma: my eslint gave me error, i'll revert it

package.json
cakasuma: sure 👍

package.json
cakasuma: that is the sub of sub dependency update, but, i think we just wait until that dependency update then we can update it. for now, ill remove it

build/aliases.yaml
```diff
@@ -20,6 +20,14 @@ default:
   - 'css'
   - 'js'
 
+start:
+  - 'css'
+  - 'js'
+  - 'copy'
```
negarn: should we do copy before css and js, same as in default?

build/sass.js
```diff
@@ -1,6 +1,9 @@
+const sass = require('node-sass');
+
 module.exports = function (grunt) {
     const options = {
         style: 'expanded',
+        implementation: sass,
```
negarn: what does this do?

cakasuma: the new grunt-sass needs implementation passed as an argument in the option, the sass implementation comes from node-sass dependency

PR #5347, opened 2019-04-17
arash/remove the won/lost status shade on position when dropdown is opened

arash-binary:


src/javascript/app_2/App/Components/Elements/PositionsDrawer/positions-drawer-card.jsx
```diff
@@ -13,177 +13,191 @@ import { getTimePercentage } from './helpers';
...
-            <ContractLink
-                className={classNames(
-                    'positions-drawer-card', {
-                        'positions-drawer-card--active': (parseInt(active_position) === id),
-                        'positions-drawer-card--green' : (profit_loss > 0) && !result,
-                        'positions-drawer-card--red'   : (profit_loss < 0) && !result,
-                    }
-                )}
-                contract_id={id}
+class PositionsDrawerCard extends React.Component {
```
zakibinary: Should be PureComponent

src/javascript/app_2/App/Components/Elements/PositionsDrawer/positions-drawer-card.jsx
```diff
@@ -13,177 +13,191 @@ import { getTimePercentage } from './helpers';
...
                         >
-                            {result ? localize('P/L:') : localize('Potential P/L:')}
+                            <div className='positions-drawer-card__underlying-name'>
+                                <div
+                                    className={classNames(
+                                        'icons-underlying',
+                                        `icons-underlying__ic-${contract_info.underlying || 'unknown'}`
+                                    )}
+                                />
+                                <span className='positions-drawer-card__symbol'>{contract_info.display_name}</span>
```
zakibinary: nitpick: break into multiple lines?

src/javascript/app_2/App/Components/Elements/PositionsDrawer/result-overlay.jsx
```diff
@@ -11,6 +11,7 @@ const ResultOverlay = ({
     onClick,
     onClickRemove,
     result,
+    display_shade,
```
zakibinary: For `boolean` prop we could use `is_shade_visible` to make it consistent with other props in the project. `is_visible` is fine too, since this prop only applies within the context of this component.

src/javascript/app_2/App/Components/Elements/PositionsDrawer/result-overlay.jsx
```diff
@@ -60,6 +62,7 @@ const ResultOverlay = ({
 
 ResultOverlay.propTypes = {
     contract_id  : PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+    display_shade: PropTypes.bool,
     onClick      : PropTypes.func,
     onClickRemove: PropTypes.func,
```
zakibinary: FYI, the `close` or `X` button will still work but is not visible to users since it is part of the `overlay` which has `visibility: hidden;` when `shade` is not visible. Maybe we need to separate it out from the `overlay` component and make it `visible` once contract has ended.

src/javascript/app_2/App/Components/Elements/PositionsDrawer/positions-drawer-card.jsx
arash-binary: done.

src/javascript/app_2/App/Components/Elements/PositionsDrawer/result-overlay.jsx
arash-binary: Sure,
Done.

src/javascript/app_2/App/Components/Elements/PositionsDrawer/result-overlay.jsx
arash-binary: Working on it, Thanks for noticing that.

src/javascript/app_2/App/Components/Elements/PositionsDrawer/positions-drawer-card.jsx
arash-binary: It's done.

PR #5349, opened 2019-04-18
negar/hi_low_labels

negarn:
stop showing highest/lowest tick and draw chart


negarn:
@cakasuma we change highest/lowest tracking and labels and just show exit tick same as other contracts

PR #5362, opened 2019-04-22
arash/Changing of Binary (BVI) Registered Address {9}

arash-binary:


src/javascript/config.js
```diff
@@ -25,7 +25,7 @@ const binary_desktop_app_id = 14473;
 
 const getAppId = () => {
     let app_id = null;
-    const user_app_id   = ''; // you can insert Application ID of your registered application here
+    const user_app_id   = '16125'; // you can insert Application ID of your registered application here
```
ashkanx: You should NOT commit this.

ashkanx: You can do `git update-index --assume-unchanged src/javascript/config.js`

arash-binary: Sure,

PR #5368, opened 2019-04-23
Smartcharts 0.4.17

ashkanx:


package-lock.json
```diff
@@ -924,7 +924,7 @@
     },
     "ansi-colors": {
       "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/ansi-colors/-/ansi-colors-1.1.0.tgz",
+      "resolved": "http://registry.npmjs.org/ansi-colors/-/ansi-colors-1.1.0.tgz",
```
negarn: why are we changing to http?

ashkanx: no idea. @morteza-binary ?

negarn: should be related to your node version or npm version right? some people will change to http, some people will change to https, which is why we run `npm ci` but still face this when a different person runs `npm i` to install a new package

morteza-binary: @ashkanx I don't have any idea, AFAIK, it was an issue in the older version of NPM but it should be solved.

PR #5376, opened 2019-04-24
`Install PWA` Design Updated

msamprz:


src/javascript/app_2/App/Containers/Layout/install-pwa.jsx
```diff
@@ -22,18 +23,25 @@ const InstallPWA = ({
 
     return (
         <React.Fragment>
-            <p>{localize('Want quick and easy access to world markets?')}</p>
+            <p className='notification-bar__message'>{localize('Install Binary.com app?')}</p>
```
negarn: use website_name please

PR #5383, opened 2019-04-25
negar/send_referrer

negarn:
send referrer


src/javascript/app/base/binary_loader.js
```diff
@@ -74,7 +74,8 @@ const BinaryLoader = (() => {
         ContentVisibility.init();
 
         BinarySocket.wait('authorize', 'website_status', 'landing_company').then(() => {
-            GTM.pushDataLayer({ event: 'page_load' }); // we need website_status.clients_country
+            const utm_source = Url.paramsHash().utm_source;
+            GTM.pushDataLayer({ event: 'page_load', ...(utm_source && { utm_source }) }); // we need website_status.clients_country
```
ashkanx: We don't need to push it to datalayer. as long as it's in the URL, tracking tools can access it.

PR #5423, opened 2019-05-03
negar/change_ws

negarn:
use frontend instead of ws

### DIFF
--- a/src/root_files/app_2/index.html
+++ b/src/root_files/app_2/index.html
@@ -64,7 +64,7 @@
             if (lang) {
                 redirect();
             } else {
-                var ws = new WebSocket('wss://ws.binaryws.com/websockets/v3?app_id=1');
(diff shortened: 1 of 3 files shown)


src/root_files/app_2/index.html
```diff
@@ -64,7 +64,7 @@
             if (lang) {
                 redirect();
             } else {
-                var ws = new WebSocket('wss://ws.binaryws.com/websockets/v3?app_id=1');
+                var ws = new WebSocket('wss://frontend.binaryws.com/websockets/v3?app_id=1');
```
morteza-binary: @negar-binary Should we use the same app_id for `binary-2`?

negarn: I think it doesn't matter, this is just to get the language for a logged out user without language cookie, @ashkanx let me know if we want to track that with the app_2 app id

PR #5429, opened 2019-05-06
Logout without page reload

easteregg:


src/javascript/app_2/App/Containers/AccountSwitcher/account-switcher.jsx
```diff
@@ -161,6 +162,7 @@ const account_switcher = connect(
         is_virtual            : client.is_virtual,
         switchAccount         : client.switchAccount,
         upgrade_info          : client.upgrade_info,
+        cleanUp               : client.cleanUp,
```
OskarAhl: LGTM, can we just add this as a propType too?

easteregg: Ah, You are right @OskarAhl. right away.

PR #5431, opened 2019-05-07
amam/age_verification

cakasuma:
add new message for age verification ukgc


src/javascript/app/pages/trade/purchase.js
```diff
@@ -134,6 +134,19 @@ const Purchase = (() => {
                                 additional_message = localize('Try our other markets.');
                             }
                             message = `${error.message}. ${additional_message}`;
+                        } else if (/ClientUnwelcome/.test(error.code) && /gb/.test(State.getResponse('authorize.country'))) {
+                            let message_text = '';
+                            let additional_message = '';
+
+                            if (!Client.hasAccountType('real') && /^VRTC/.test(State.getResponse('authorize.loginid'))) {
```
negarn: please instead of the second condition use `Client.get('is_virtual')`

src/javascript/app/pages/trade/purchase.js
```diff
@@ -134,6 +134,19 @@ const Purchase = (() => {
...
+                                message_text = localize('Please complete the [_1]Real Account form[_2] to verify your age as required by the <strong>UK Gambling</strong> Commission (UKGC).', [`<a href='${urlFor('new_account/realws')}'>`, '</a>']);
```
negarn: do we want to use placeholders or keep the `<strong>` tag in the translation string?

src/javascript/app/pages/trade/purchase.js
```diff
@@ -134,6 +134,19 @@ const Purchase = (() => {
                                 additional_message = localize('Try our other markets.');
                             }
                             message = `${error.message}. ${additional_message}`;
+                        } else if (/ClientUnwelcome/.test(error.code) && /gb/.test(State.getResponse('authorize.country'))) {
```
negarn: please instead use `Client.get('residence')` to get the country of residence

src/javascript/app/pages/trade/purchase.js
cakasuma: if I am not mistaken, we want to avoid using too many placeholders since it will make confusion to the translator?

src/javascript/app/pages/trade/purchase.js
```diff
@@ -134,6 +134,19 @@ const Purchase = (() => {
...
+                            let message_text = '';
+                            let additional_message = '';
+
+                            if (!Client.hasAccountType('real') && /^VRTC/.test(State.getResponse('authorize.loginid'))) {
+                                message_text = localize('Please complete the [_1]Real Account form[_2] to verify your age as required by the <strong>UK Gambling</strong> Commission (UKGC).', [`<a href='${urlFor('new_account/realws')}'>`, '</a>']);
+                                additional_message = localize('Once verified, you will be able to trade on both real and virtual accounts.');
+                            } else if (Client.hasAccountType('real') && /^MX|VRTC/.test(State.getResponse('authorize.loginid'))) {
```
negarn: we don't want to check the broker code unless we have to, it's always better to check landing company or existing functions. and for responses of authorize, they are all stored in Client so we should get them from there. in this case my suggestion is to use something similar to `/^virtual|iom$/i.test(Client.get('landing_company_shortcode')` instead.

src/javascript/app/pages/trade/purchase.js
```diff
@@ -134,6 +134,19 @@ const Purchase = (() => {
...
+                                message_text = localize('Your age verification failed. Please contact customer service for assistance. <br/><br/> <strong>Telephone:</strong> <br/> United Kingdom <br/> +44 (0) 1666 800042 <br/> 0800 011 9847 (Toll Free)');
```
negarn: do we want to keep all the HTML tags in the translation string?

src/javascript/app/pages/trade/purchase.js
```diff
@@ -134,6 +134,19 @@ const Purchase = (() => {
...
+                                additional_message = localize('[_1]Telephone numbers in other locations[_2]', [`<a href='${urlFor('contact')}'>`, '</a>']);
+                            }
+
+                            message = `${message_text} <br/><br/> ${additional_message}`;
```
negarn: is there a case where message_text and additional_message are both empty? should we in that case have message be empty instead of two br tags?

src/javascript/app/pages/user/new_account/real_acc_opening.js
```diff
@@ -13,12 +13,15 @@ const RealAccOpening = (() => {
         if (Client.get('residence')) {
             if (AccountOpening.redirectAccount()) return;
 
-            BinarySocket.wait('landing_company').then(() => {
+            BinarySocket.wait('authorize', 'landing_company').then(() => {
```
negarn: this page is already an authorized page, meaning we already wait for authorize before loading this page, handled in binary_loader.js. you don't need this additional wait for authorize.

src/javascript/app/pages/trade/purchase.js
cakasuma: I fix it by adding an else statement, will get the original message of unwelcome

src/javascript/app/pages/trade/purchase.js
cakasuma: maybe placeholders for <br/>?

src/javascript/app/pages/trade/purchase.js
cakasuma: let me keep it as placeholders instead

src/javascript/app/pages/user/new_account/real_acc_opening.js
```diff
@@ -19,6 +19,9 @@ const RealAccOpening = (() => {
                     State.getResponse('authorize.upgradeable_landing_companies').indexOf('costarica') !== -1) {
                     getElementById('risk_disclaimer').setVisibility(1);
                 }
+                if (/gb/.test(Client.get('residence'))) {
```
negarn: do we want to check that client is also having status unwelcome instead of assuming?

src/javascript/app/pages/user/new_account/virtual_acc_opening.js
```diff
@@ -121,7 +121,7 @@ const VirtualAccOpening = (() => {
                         loginid     : new_account.client_id,
                         token       : new_account.oauth_token,
                         is_virtual  : true,
-                        redirect_url: urlFor('new_account/welcome'),
+                        redirect_url: /gb/.test(residence) ? urlFor('new_account/realws') : urlFor('new_account/welcome'),
```
negarn: do we want to check that client is also having status unwelcome instead of assuming?

PR #5448, opened 2019-05-08
sash/grid LP improvemenets

sashcy:


negarn:
@sashcy where is src/images/grid_lp/phone@2x.png used? can we remove it if unused?

sashcy:
> @sashcy where is src/images/grid_lp/phone@2x.png used? can we remove it if unused?

sorry, i thought i deleted this

PR #5461, opened 2019-05-13
negar/remove_app_2

negarn:
remove app_2


build/config/init.js
```diff
@@ -29,18 +29,15 @@ const initGlobals = (grunt) => {
 
         if (global.release_target === 'staging' && global.section === 'all') {
```
OskarAhl: `global.section === 'all'` should also be removed now?

build/copy.js
```diff
@@ -24,39 +24,20 @@ module.exports = function (grunt) {
...
-
-                // binary-style
-                { expand: true, cwd: `${global.node_modules_paths.binary_style}/src/images/favicons`, src: ['**'], dest: `${global.dist_app_2}/images/favicons/` },
-
-                // ChartIQ dependencies
-                { expand: true, cwd: `${global.node_modules_paths.smartcharts}/dist/`, src: ['*.smartcharts.*'],  dest: `${global.dist_app_2}/js/smartcharts/` },
-                { expand: true, cwd: `${global.node_modules_paths.smartcharts}/dist/`, src: ['smartcharts.css*'], dest: `${global.dist_app_2}/css/` },
-            ],
-        },
         get all() {
```
OskarAhl: I don't think we need all() here either since it's only one section now? Same in the files below

package.json
```diff
@@ -119,27 +118,17 @@
...
     "jquery.scrollto": "2.1.2",
     "js-cookie": "2.2.0",
     "kinetic": "5.2.0",
-    "lodash.debounce": "4.0.8",
-    "mobx": "5.0.3",
-    "mobx-react": "^5.4.2",
-    "mobx-utils": "^5.4.0",
     "moment": "2.19.4",
     "promise-polyfill": "6.1.0",
     "prop-types": "15.6.1",
```
OskarAhl: Can be removed too since it's not used in static?

negarn: it's used in contract jsx

build/copy.js
negarn: I kept it in case we want to maintain the config to add sections again in the future. should I remove it?

negarn: talked to ashkan, will remove it for now

PR #5474, opened 2019-05-14
Dubai Career Page

msamprz:


src/templates/static/about/careers.jsx
```diff
@@ -127,8 +127,8 @@ const Careers = () => {
                         <ColumnLG image='malta2x'    text={it.L('Malta')}                       link={it.url_for('malta')} />
                         <ColumnLG image='my2x'       text={it.L('Cyberjaya, Malaysia')}         link={it.url_for('cyberjaya')} />
                         <ColumnLG image='labuan2x'   text={it.L('Labuan, Malaysia')}            link={it.url_for('labuan')} />
-                        <ColumnLG image='asuncion2x' text={it.L('Asunción, Paraguay')}          link={it.url_for('asuncion')} />
-                        <ColumnLG image='dubai2x'    text={it.L('Dubai, United Arab Emirates')} link={it.url_for('dubai')} />
+                        <ColumnLG image='asuncion2x' text={it.L('Asunción, Paraguay')}          link={it.url_for('asuncion')} margin_top={true} />
+                        <ColumnLG image='dubai2x'    text={it.L('Dubai, United Arab Emirates')} link={it.url_for('dubai')}    margin_top={true} />
```
negarn: I think our style guide is not to send {true} just pass the variable, can you change it?

negarn: https://github.com/binary-com/binary-static/tree/master/src/templates#jsx-rules

PR #5496, opened 2019-05-20
Elevio pageview

msamprz:


src/javascript/_common/base/elevio.js
```diff
@@ -59,6 +58,10 @@ const Elevio = (() => {
             setTranslations(elev);
             addEventListenerGTM();
             makeLauncherVisible();
+
+            if (is_open) {
+                elev.open();
```
negarn: should this be `window._elev.open(); // eslint-disable-line no-underscore-dangle`?

PR #5500, opened 2019-05-21
khalid/removed undefined header

Khalidxx:


src/javascript/app/base/binary_loader.js
```diff
@@ -149,7 +149,7 @@ const BinaryLoader = (() => {
             return;
         }
 
-        const div_container = createElement('div', { class: 'logged_out_title_container', html: content.getElementsByTagName('h1')[0] });
+        const div_container = createElement('div', { class: 'logged_out_title_container', html: content.getElementsByTagName('h1')[0] ? content.getElementsByTagName('h1')[0] : '' });
```
negarn: better store it in a variable so we don't need to retrieve the element twice

negarn: or use `content.getElementsByTagName('h1')[0] || ''`

Khalidxx: Okay 👍

src/javascript/app/base/binary_loader.js
```diff
@@ -149,7 +149,8 @@ const BinaryLoader = (() => {
             return;
         }
 
-        const div_container = createElement('div', { class: 'logged_out_title_container', html: content.getElementsByTagName('h1')[0] });
+        const title = content.getElementsByTagName('h1')[0];
```
negarn: please align the `=` with below

Khalidxx: I removed the title 😅 I guess it looks better without

PR #5506, opened 2019-05-21
arash/mt5 dashboard

arash-binary:


src/templates/app/user/metatrader.jsx
```diff
@@ -180,6 +180,18 @@ const Metatrader = () => (
...
                             </div>
                         </div>
+                        <div className='mt-sidebar-button'>
+                            <div className='small-icon'>
+                                <a href={it.url_for('metatrader/download')}>
+                                    <img src={it.url_for('images/pages/metatrader/dashboard/img-app-mac@2x.png')} />
+                                </a>
+                            </div>
+                            <div className='mt-link'>
+                                <a href='https://trade.mql5.com'>
```
ashkanx: @arash-binary  please add `target` and `rel` attributes

arash-binary: Sure

arash-binary: @ashkanx ` target='_blank' rel='noopener noreferrer'` is it right?

ashkanx: yes

PR #5513, opened 2019-05-23
Khalid/PA withdrawal limit

Khalidxx:
Get the Payment Agent withdrawal limit from the API instead of being hardcoded. If there's an error getting the value from the API, the default value is left as the original hardcoded one.


src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -81,8 +96,8 @@ const PaymentAgentWithdraw = (() => {
             $form.find('label[for="txtAmount"]').text(`${localize('Amount in')} ${currency}`);
             trimDescriptionContent();
             FormManager.init(form_id, [
-                { selector: field_ids.txt_amount,        validations: ['req', ['number', { type: 'float', decimals: getDecimalPlaces(currency), min, max }], ['custom', { func: () => +Client.get('balance') >= +$(field_ids.txt_amount).val(), message: localize('Insufficient balance.') }]], request_field: 'amount' },
-                { selector: field_ids.txt_desc,          validations: ['general'], request_field: 'description' },
+                { selector: field_ids.txt_amount,         validations: ['req', ['number', { type: 'float', decimals: getDecimalPlaces(currency), min, max }], ['custom', { func: () => +Client.get('balance') >= +$(field_ids.txt_amount).val(), message: localize('Insufficient balance.') }]], request_field: 'amount' },
```
cakasuma:
```suggestion
                { selector: field_ids.txt_amount, validations: ['req', ['number', { type: 'float', decimals: getDecimalPlaces(currency), min, max }], ['custom', { func: () => +Client.get('balance') >= +$(field_ids.txt_amount).val(), message: localize('Insufficient balance.') }]], request_field: 'amount' },
```

cakasuma:
```suggestion
                { selector: field_ids.txt_amount, validations: ['req', ['number', { type: 'float', decimals: getDecimalPlaces(currency), min, max }], ['custom', { func: () => +Client.get('balance') >= +$(field_ids.txt_amount).val(), message: localize('Insufficient balance.') }]], request_field: 'amount' },
```

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -81,8 +96,8 @@ const PaymentAgentWithdraw = (() => {
...
+                { selector: field_ids.txt_desc,           validations: ['general'], request_field: 'description' },
```
cakasuma:
```suggestion
                { selector: field_ids.txt_desc,   validations: ['general'], request_field: 'description' },
```

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -71,8 +71,23 @@ const PaymentAgentWithdraw = (() => {
...
+                return getPaWithdrawalLimit(currency, 'min');
+            };
+            const max  = () => {
+                const selected_val = $ddl_agents.val() || $txt_agents.val();
+                if (selected_val){
+                    const selected_pa = pa_list.find(pa => pa.paymentagent_loginid === selected_val);
+                    if (selected_pa) return selected_pa.max_withdrawal;
+                }
+                return getPaWithdrawalLimit(currency, 'max');
+            };
```
OskarAhl: I think you could combine these 2 functions to 1, or move the function body to a new shared function called in min and max since the logic is more or less the same

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -71,8 +71,23 @@ const PaymentAgentWithdraw = (() => {
 
             const form_id = `#${$(view_ids.form).find('form').attr('id')}`;
             const $form   = $(form_id);
-            const min     = getPaWithdrawalLimit(currency, 'min');
-            const max     = getPaWithdrawalLimit(currency, 'max');
+
+            const min  = () => {
+                const selected_val = $ddl_agents.val() || $txt_agents.val();
```
negarn: why not use the function `getPALoginID`?

src/javascript/app/pages/cashier/payment_agent_withdraw.js
Khalidxx: Alright 👍

src/javascript/app/pages/cashier/payment_agent_withdraw.js
Khalidxx: Didn't notice the function 😄

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -109,6 +110,9 @@ const PaymentAgentWithdraw = (() => {
                     // error handling
                     $agent_error.setVisibility(1);
                 }
+                if ($(field_ids.txt_amount).val()) {
+                    Validation.validate(form_id);
```
negarn: don't we also need it for `$txt_agents.on('keyup'`?

Khalidxx: But it scrolls down to the `txtAmount` field every time a key is pressed

negarn: then better use a different listener to value change instead of keyup on txt_agents if possible

negarn: or, I like your idea better, let's go for that.

Khalidxx: Alright, done 👍

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -113,6 +127,12 @@ const PaymentAgentWithdraw = (() => {
                 }
             });
 
+            $txt_agents.on('focusout', () => {
+                if ($(field_ids.txt_amount).val()) {
```
negarn: @Khalidxx can you define this element outside the listener so we don't have to keep fetching it? something like on line 129 `const $txt_amount = $(field_ids.txt_amount)` and then on line 131 `if ($txt_amount.val()) {`

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -99,6 +110,9 @@ const PaymentAgentWithdraw = (() => {
                     // error handling
                     $agent_error.setVisibility(1);
                 }
+                if ($(field_ids.txt_amount).val()) {
+                    Validation.validate(form_id);
```
negarn: can we put this in a function and call it both on line 131 and 113? since it's the same if condition and function call.

src/javascript/app/pages/cashier/payment_agent_withdraw.js
Khalidxx: I defined it up there with the other elements

src/javascript/app/pages/cashier/payment_agent_withdraw.js
Khalidxx: Done!

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -1,4 +1,5 @@
 const refreshDropdown      = require('@binary-com/binary-style').selectDropdown;
+const Validation           = require('../../common/form_validation');
```
negarn: could you also sort this alphabetically? we actually already have form_validation on line 9, perhaps you can remove `.validEmailToken` from that line, remove this line, and call that function using `Validation.validEmailToken` throughout the code below instead @Khalidxx

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -113,6 +127,16 @@ const PaymentAgentWithdraw = (() => {
                 }
             });
 
+            $txt_agents.on('focusout', () => {
+                checkValidation();
+            });
+
+            const checkValidation = () => {
```
negarn: better use a more specific function name, e.g. `validateAmount`

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -71,18 +72,29 @@ const PaymentAgentWithdraw = (() => {
...
+                const selected_val = getPALoginID();
+                if (selected_val){
+                    const selected_pa = pa_list.find(pa => pa.paymentagent_loginid === selected_val);
+                    if (selected_pa) return selected_pa[`${limit}_withdrawal`];
+                }
+                return getPaWithdrawalLimit(currency, limit);
+            };
+            
+            const min  = () => getAPILimit('min');
+            const max  = () => getAPILimit('max');
```
negarn: there should only be a single space before `=` (currently two)

PR #5525, opened 2019-05-27
amam/add_affiliate_token

cakasuma:
add affiliate token and UTMs to social login URL

### DIFF
--- a/src/javascript/_common/base/gtm.js
+++ b/src/javascript/_common/base/gtm.js
@@ -26,8 +26,8 @@ const GTM = (() => {
     });
 
     const pushDataLayer = (data) => {
-        const is_login_page = /logged_inws|redirect/i.test(window.location.pathname);
-        if (isGtmApplicable() && !is_login_page) {
+        const is_login_pages = /logged_inws|redirect/i.test(window.location.pathname);
+        if (isGtmApplicable() && !is_login_pages) {
             dataLayer.push({
                 ...getCommonVariables(),
                 ...data,
--- a/src/javascript/_common/base/gtm.js
+++ b/src/javascript/_common/base/gtm.js
@@ -1,7 +1,6 @@
 const Cookies          = require('js-cookie');
 const moment           = require('moment');
 const ClientBase       = require('./client_base');
-const Login            = require('./login');
 const ServerTime       = require('./server_time');
 const BinarySocket     = require('./socket_base');
 const getElementById   = require('../common_functions').getElementById;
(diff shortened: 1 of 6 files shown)


src/javascript/_common/base/gtm.js
```diff
@@ -27,7 +26,8 @@ const GTM = (() => {
     });
 
     const pushDataLayer = (data) => {
-        if (isGtmApplicable() && !Login.isLoginPages()) {
+        const is_login_pages = /logged_inws|redirect/i.test(window.location.pathname);
```
negarn: why don't we fix the function?

src/javascript/_common/base/login.js
```diff
@@ -1,4 +1,6 @@
+const Cookies             = require('js-cookie');
 const Client              = require('./client_base');
+const TrafficSource       = require('../../app/common/traffic_source');
```
negarn: should be on line 9 instead

src/javascript/_common/base/login.js
```diff
@@ -8,7 +10,8 @@ const getAppId            = require('../../config').getAppId;
 
 const Login = (() => {
     const redirectToLogin = () => {
-        if (!Client.isLoggedIn() && !isLoginPages() && isStorageSupported(sessionStorage)) {
+        const is_login_pages = /logged_inws|redirect/i.test(window.location.pathname);
```
negarn: same here, why not fix the function?

src/javascript/_common/base/login.js
```diff
@@ -19,30 +22,36 @@ const Login = (() => {
...
+    );
 
     const initOneAll = () => {
         ['google', 'facebook'].forEach(provider => {
             $(`#button_${provider}`).off('click').on('click', e => {
                 e.preventDefault();
-                window.location.href = socialLoginUrl(provider);
+
+                const utm_data     = TrafficSource.getData();
+                const utm_source   = TrafficSource.getSource(utm_data) ? `&utm_source=${TrafficSource.getSource(utm_data)}` : '';
```
negarn: better store `TrafficSource.getSource(utm_data)` in a variable so we don't call it twice

src/javascript/_common/base/gtm.js
cakasuma: i put the function back to normal. however. we can't import login here since it will be circular dependency here

negarn: what is the circular dependency?

negarn: as discussed, please move it to a utility function to eliminate the race condition.

src/javascript/_common/base/login.js
```diff
@@ -1,9 +1,12 @@
+const Cookies             = require('js-cookie');
 const Client              = require('./client_base');
 const getLanguage         = require('../language').get;
+const isLoginPages        = require('../utility').isLoginPages;
```
negarn: should be on line 8

PR #5528, opened 2019-05-28
Khalid/empty barrier error 

Khalidxx:
- An error is shown when the barrier entered is `0`
- `+0` and `-0` barriers are now allowed


negarn:
@Khalidxx in your PR you are also committing @cakasuma 's changes, perhaps a git conflict resolve wrong? please reset your branch and only include your changes

src/javascript/app/pages/trade/barriers.js
```diff
@@ -198,6 +198,17 @@ const Barriers = (() => {
         });
     };
 
+    const isAbsoluteZero = num =>  {
+        if (parseFloat(num) !== 0) {
+            return false;
+        } else if (Object.is(parseFloat(num), -0)) {
+            return false;
+        } else if (num.toString().charAt(0) === '+') {
```
negarn: why not do `/+|-/.test(num.toString().charAt(0))` and remove the second condition?

src/javascript/app/pages/trade/barriers.js
```diff
@@ -198,6 +198,17 @@ const Barriers = (() => {
...
+            return false;
+        }
+        return true;
```
negarn: so it would be something like:
```
const isAbsoluteZero = num => !(parseFloat(num) !== 0 || /+|-/.test(num.toString().charAt(0)));
```

negarn: in that case it doesn't need to be a function, it can be a variable and moved to above `const empty`

Khalidxx: That looks leaner! I'll update it

PR #5529, opened 2019-05-28
khalid/scroll to error message

Khalidxx:
When a new account cannot be opened, the view is automatically scrolled to the error message


src/javascript/app/common/account_opening.js
```diff
@@ -197,9 +197,11 @@ const AccountOpening = (() => {
     const handleNewAccount = (response, message_type) => {
         if (response.error) {
             const errorMessage = response.error.message;
+            const noticeBox    = $('#client_message').find('.notice-msg');
```
negarn: for jQuery elements we use `$` in front of the name to make it obvious, and also the format should be lowercase letters separated by `_`, so `$notice_box`. could you also fix `error_message` above that? seems it was missed

src/javascript/app/common/account_opening.js
```diff
@@ -197,9 +197,11 @@ const AccountOpening = (() => {
...
             $('#submit-message').empty();
-            $('#client_message').find('.notice-msg').text(response.msg_type === 'sanity_check' ? localize('There was some invalid character in an input field.') : errorMessage).end()
+            noticeBox.text(response.msg_type === 'sanity_check' ? localize('There was some invalid character in an input field.') : errorMessage).end()
                 .setVisibility(1);
+            $.scrollTo(noticeBox, 500, { offset: -150 });
```
negarn: why such a large offset?

src/javascript/app/common/account_opening.js
Khalidxx: Roger!

src/javascript/app/common/account_opening.js
Khalidxx: It scrolls so as to keep the message at the top of the view. I was trying to get the message in the middle

PR #5534, opened 2019-05-29
amam|khalid/account_closure

cakasuma:


src/templates/app/user/security/account_closure.jsx
```diff
@@ -0,0 +1,132 @@
...
+                            title={it.L('Change my affiliate')}
+                            subtitle={it.L('Contact [_1]affiliates@binary.com[_2] for more info on changing your affiliate.', '<a href="mailto:affiliates@binary.com">', '</a>')}
+                        />
+                        <ClosureDescription
+                            title={it.L('Change my account limits')}
+                            subtitle={it.L('You may set limits in your account to help prevent unwanted losses.[_1]Go to [_2]self-exclusion page[_3] to manage your account limits.', '<br />', `<a href=${it.url_for('user/security/self_exclusionws')}>`, '</a>')}
+                        />
+                    </div>
+                </fieldset>
+                {/* TODO: complete these component */}
```
zakibinary: Is component completed? If yes, can you please remove the comment.

src/templates/app/user/security/account_closure.jsx
```diff
@@ -0,0 +1,132 @@
...
+const AccountClosure = () => (
+    <React.Fragment>
+        <div id='msg_main' className='center-text gr-gutter gr-padding-10 invisible'>
+            <h1>{it.L('Account closure confirmed')}</h1>
+            <p className='notice-msg'>{it.L(`Accounts closed successfully. A confirmation email will be sent to your email.[_1]This page will redirect to the ${it.website_name} homepage after 10 seconds.`, '<br />')}</p>
+        </div>
+        <div id='closure-container'>
+            <div id='main-header' className='gr-padding-30'>
+                <h1 id='heading'>{it.L('Account Closure')}</h1>
+                <p>{it.L(`Closing your ${it.website_name} accounts involves closing all open positions in your accounts, and withdrawing your funds, and deactivating your accounts with ${it.website_name}`)}</p>
```
zakibinary: Same here. Should be:
 ```<p>{it.L('Closing your [_1] accounts involves closing all open positions in your accounts, and withdrawing your funds, and deactivating your accounts with [_1]', it.website_name)}</p>```

src/templates/app/user/security/account_closure.jsx
```diff
@@ -0,0 +1,132 @@
+import React from 'react';
+import Loading from '../../../_common/components/loading.jsx';
+import { SeparatorLine } from '../../../_common/components/separator_line.jsx';
+import { FormRow, SubmitButton } from '../../../_common/components/forms.jsx';
+
+const AccountClosure = () => (
+    <React.Fragment>
+        <div id='msg_main' className='center-text gr-gutter gr-padding-10 invisible'>
+            <h1>{it.L('Account closure confirmed')}</h1>
+            <p className='notice-msg'>{it.L(`Accounts closed successfully. A confirmation email will be sent to your email.[_1]This page will redirect to the ${it.website_name} homepage after 10 seconds.`, '<br />')}</p>
```
zakibinary: We shouldn't include `it.website_name` as part of the localized string.
Should be:
```<p className='notice-msg'>{it.L('Accounts closed successfully. A confirmation email will be sent to your email.[_1]This page will redirect to the [_2] homepage after 10 seconds.' , '<br />', it.website_name)}</p>```

File: src/javascript/app/pages/user/account/settings/account_closure.js [hunk 4]
```diff
@@ -0,0 +1,86 @@
...
+        const radioText  = $(`label[for=${id}]`).text();
+        const textInput  = $textField.val();
+        if (value){
+            if (value === 'other') {
+                if (!textInput) {
+                    $textField.addClass('error-field');
+                    showFormMessage(localize('Please specify your reason.'));
+                    return false;
+                } else if (textInput.length > 3 && textInput.length < 50) return textInput;
+                showFormMessage(localize('Maximum lenght: 50 characters'));
```
zakibinary: `length` typo.

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,86 @@
...
+
+    const showFormMessage = (localized_msg, is_success) => {
+        if (is_success) {
+            $.scrollTo($('h1#heading'), 500, { offset: -10 });
+            $(form_selector).setVisibility(0);
+            $('#msg_main').setVisibility(1);
+        } else {
+            $.scrollTo($('#reason'), 500, { offset: -20 });
+            $('#msg_form')
+                .attr('class', is_success ? 'success-msg' : 'errorfield')
```
zakibinary: `className` should be `error-field` or `errorfield`?

src/templates/app/user/security.jsx
```diff
@@ -56,6 +56,9 @@ const Security = () => (
                 <Column url='two_factor_authentication' image='2fa' header={it.L('Two-Factor Authentication')} text={it.L('Enable two-factor authentication for an extra layer of security.')} />
 
                 <Column url='vpn_app' image='vpn' header={it.L('VPN App')} text={it.L('Establish a secure Internet connection and protect your privacy.')} />
+
+                {/* TODO: new image and text description from ui design */}
```
zakibinary: Ditto, remove comment if it's no longer applicable.

src/templates/app/user/security/account_closure.jsx
Khalidxx: @cakasuma left these comments, but it seems completed. I'll remove it for now.

scripts/config/pages.js
```diff
@@ -37,6 +37,7 @@ module.exports = [
     ['user/reality_check_summary',              'app/user/reality_check/summary',              'default',    'Reality Check'],
     ['user/reset_passwordws',                   'app/user/reset_password',                     'default',    'Password Reset'],
     ['user/securityws',                         'app/user/security',                           'default',    'Security'],
+    ['user/security/closurews',                 'app/user/security/account_closure',           'default',    'Account Closure'],
```
negarn: please don't put `ws` in new urls, we used to add them before when marking which pages have moved to websocket, but now we can't change them since other people may have links to them. there's no need to add it for new urls

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,86 @@
+const BinarySocket     = require('../../../../base/socket');
+const localize         = require('../../../../../_common/localize').localize;
+const Url              = require('../../../../../_common/url');
```
negarn: empty spaces before `=`

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,86 @@
...
+const AccountClosure = (() => {
+    const form_selector = '#form_closure';
+
+    const onLoad = () => {
+        $(form_selector).on('submit', (event) => {
+            event.preventDefault();
+            submitForm();
+        });
+        $('#closure-container').setVisibility(1);
+        $('#other-reason').on('keyup', () => $('#other-reason').removeClass('error-field'));
```
negarn: better store `$('#other-reason')` in a variable so we don't have to retrieve it every time on keyup

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,86 @@
+const BinarySocket     = require('../../../../base/socket');
+const localize         = require('../../../../../_common/localize').localize;
+const Url              = require('../../../../../_common/url');
+
+const AccountClosure = (() => {
+    const form_selector = '#form_closure';
+
+    const onLoad = () => {
+        $(form_selector).on('submit', (event) => {
```
negarn: any reason we're not using our form manager for submission and error handling?

src/javascript/app/pages/user/account/settings/account_closure.js
Khalidxx: It is `errorfield`, cause it's a different class and I guess it was meant for the error message not the textfield

scripts/config/pages.js
Khalidxx: Okay 👍

src/javascript/app/pages/user/account/settings/account_closure.js
negarn: then above you should also remove errorfield no error-field

src/javascript/app/pages/user/account/settings/account_closure.js
Khalidxx: Yeah, should be better. I tried defining it globally, but didn't work. I'll declare it and then assign its value on load

src/javascript/app/pages/user/account/settings/account_closure.js
Khalidxx: Because of the validation. There wasn't a conditional validation for radio button (if this radio button is selected, validate the textfield this way)

negarn: I see

src/javascript/app/pages/user/account/settings/account_closure.js
negarn: that's because jquery hasn't loaded in the global yet, so we can't assign jquery globally. you can create a let variable globally and set the jquery value onload as well.

Khalidxx: Yeah, I thought so too

src/javascript/app/pages/user/account/settings/account_closure.js
Khalidxx: I'll remove the conditions since we never call this function on success

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,81 @@
+const BinarySocket = require('../../../../base/socket');
+const localize     = require('../../../../../_common/localize').localize;
+const Url          = require('../../../../../_common/url');
+
+const AccountClosure = (() => {
+    const form_selector = '#form_closure';
+    let $textField;
```
negarn: variable names should be underscore separated, so `$text_field`, or better yet `$txt_field` to use a 3-letter input type

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,81 @@
...
+    
+    const onLoad = () => {
+        $textField = $('#other-reason');
```
negarn: better use a more specific variable name, perhaps `$txt_other_reason`

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,81 @@
...
+        const $btn_submit = $(`${form_selector} #btn_submit`);
+        const reason = getReason();
+        if (reason) {
+            $('#closure_loading').setVisibility(1);
+            $btn_submit.attr('disabled', true);
+
+            const data  = { account_closure: 1, reason };
+            BinarySocket.send(data).then((response) => {
+                if (response.error) {
+                    showFormMessage(localize('Sorry, an error occurred while processing your request.'), false);
```
negarn: why not show response.error.message to show exactly what went wrong? `response.error.message || localize('Sorry, an error occurred while processing your request.')`

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,81 @@
...
+                    $btn_submit.attr('disabled', false);
+                } else {
+                    $('#closure_loading').setVisibility(0);
```
negarn: this is called twice, better store it in a variable above.

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,81 @@
...
+                    $('#closure-container').setVisibility(0);
```
negarn: this is called twice, better store it globally.

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,81 @@
...
+                }
+            });
+        } else {
+            setTimeout(() => { $btn_submit.removeAttr('disabled'); }, 1000);
+        }
+    };
+
+    const showFormMessage = (localized_msg) => {
+        $.scrollTo($('#reason'), 500, { offset: -20 });
+        $('#msg_form')
```
negarn: I suggest storing all of these. who knows how many times this function gets called.

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,81 @@
...
+            .attr('class', 'errorfield')
+            .html(localized_msg)
+            .css('display', 'block')
```
negarn: can't we use `.setVisibility`?

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,81 @@
...
+            .delay(5000)
+            .fadeOut(500);
+    };
+
+    const getReason = () => {
+        const value      = $('input[type=radio]:checked').val();
+        const id         = $('input[type=radio]:checked').attr('id');
```
negarn: `$('input[type=radio]:checked')` is repeated twice so better store that separately. also better if we can use the id of the input field instead of `input[type=radio]` in case more radio buttons are added in the future.

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,81 @@
...
+        const radioText  = $(`label[for=${id}]`).text();
+        const textInput  = $textField.val();
```
negarn: should be lowercase separated, so `radio_text` and I suggest `other_reason_text` since it's more specific than just text input

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,81 @@
...
+        if (value){
```
negarn: missing space before `{`

src/javascript/app/pages/user/account/settings/account_closure.js
negarn: even for `radio_text`, better be specific what text that refers to

src/javascript/app/pages/user/account/settings/account_closure.js
negarn: please also use more specific variable names for these. something like value and id are too general to know what they mean

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,81 @@
...
+            if (value === 'other') {
+                if (!textInput) {
+                    $textField.addClass('error-field');
+                    showFormMessage(localize('Please specify your reason.'));
+                    return false;
+                } else if (textInput.length > 3 && textInput.length < 50) return textInput;
```
negarn: please use `{}` on a new line to match the if format

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,81 @@
...
+                showFormMessage(localize('Maximum length: 50 characters'));
```
negarn: how do you know the issue is not with a minimum value?

src/sass/static/pages.scss
```diff
@@ -1539,6 +1539,24 @@ body #not_authenticated_financial ul.checked > li {
...
+        padding-left: 15px;
+    }
+}
+
+.reason-textfield {
+    border-color: $COLOR_GRAY;
+    color: $COLOR_BLACK;
+}
+
+#invisible-label {
```
negarn: why do we need this? you can add the class 'invisible' instead

src/templates/app/user/security/account_closure.jsx
```diff
@@ -0,0 +1,132 @@
+import React from 'react';
+import Loading from '../../../_common/components/loading.jsx';
+import { SeparatorLine } from '../../../_common/components/separator_line.jsx';
+import { FormRow, SubmitButton } from '../../../_common/components/forms.jsx';
```
negarn: alignment of `from`

negarn: FormRow and SubmitButton should be in two lines

src/templates/app/user/security/account_closure.jsx
```diff
@@ -0,0 +1,132 @@
...
+
+const AccountClosure = () => (
+    <React.Fragment>
+        <div id='msg_main' className='center-text gr-gutter gr-padding-10 invisible'>
+            <h1>{it.L('Account closure confirmed')}</h1>
+            <p className='notice-msg'>{it.L('Accounts closed successfully. A confirmation email will be sent to your email.[_1]This page will redirect to the [_2] homepage after 10 seconds.', '<br />', it.website_name)}</p>
```
negarn: better use <br/> and split the string into two instead of putting it as a placeholder in the middle of translations.

src/templates/app/user/security/account_closure.jsx
```diff
@@ -0,0 +1,132 @@
...
+        </div>
+        <div id='closure-container'>
```
negarn: ID should be `_` separated, class names are `-` separated

src/templates/app/user/security/account_closure.jsx
```diff
@@ -0,0 +1,132 @@
...
+            <div id='main-header' className='gr-padding-30'>
```
negarn: same here

src/templates/app/user/security/account_closure.jsx
```diff
@@ -0,0 +1,132 @@
...
+                <h1 id='heading'>{it.L('Account Closure')}</h1>
+                <p>{it.L('Closing your [_1] accounts involves closing all open positions in your accounts, and withdrawing your funds, and deactivating your accounts with [_1]', it.website_name)}</p>
```
negarn: do we need an `and` after the first comma? since there are three items IMO and should only come after the second comma, please double check this with copywriters

src/templates/app/user/security/account_closure.jsx
```diff
@@ -0,0 +1,132 @@
...
+                            list_items={[
+                                it.L('Bitcoin (BTC)'),
+                                it.L('Ether (ETH)'),
+                                it.L('Litecoin (LTC)'),
+                                it.L('Tether (UST)'),
+                            ]}
+                        />
+                        <ClosureDescription
+                            title={it.L('Change my affiliate')}
+                            subtitle={it.L('Contact [_1]affiliates@binary.com[_2] for more info on changing your affiliate.', '<a href="mailto:affiliates@binary.com">', '</a>')}
```
negarn: better not send the email address through translations. put it in the placeholder please and just use `[_1]` for the whole thing.

negarn: do we want to use `info` or `information`? sounds very informal :\

src/templates/app/user/security/account_closure.jsx
```diff
@@ -0,0 +1,132 @@
...
+                        />
+                        <ClosureDescription
+                            title={it.L('Change my account limits')}
+                            subtitle={it.L('You may set limits in your account to help prevent unwanted losses.[_1]Go to [_2]self-exclusion page[_3] to manage your account limits.', '<br />', `<a href=${it.url_for('user/security/self_exclusionws')}>`, '</a>')}
```
negarn: same here, please separate into two strings and don't use `<br/>` in the translation string. for example, change `subtitle` to accept an array of items and split them by `<br/>` in the component.

src/templates/app/user/security/account_closure.jsx
negarn: comment still not removed?

src/templates/app/user/security/account_closure.jsx
```diff
@@ -0,0 +1,132 @@
...
+                    subtitle={it.L('Remember to withdraw your funds from [_1]all[_2] your accounts', '<strong>', '</strong>')}
+                    list_items={[
+                        it.L('Go to [_1]Cashier[_2] to withdraw.', `<a href="${it.url_for('cashier')}">`, '</a>'),
+                        it.L('Go to [_1]MT5 dashboard[_2] to withdraw from your [_3] MT5 account.', `<a href="${it.url_for('user/metatrader')}">`, '</a>', it.website_name),
+                    ]}
+                />
+                <SeparatorLine className='gr-padding-10' />
+            
+                <h2 className='primary-color'>{it.L('Reason for closure')}</h2>
+                <p>{it.L('Why do you want to close your account?: (Please select one)')}</p>
```
negarn: is it correct to be using `?:`? please double check with copywriter

src/templates/app/user/security/account_closure.jsx
```diff
@@ -0,0 +1,132 @@
...
+                            value: 'not interested',
+                        },
+                        {
+                            label: it.L('Prefer another trading website'),
+                            value: 'another website',
+                        },
+                        {
+                            label       : it.L('Others (please specify)'),
+                            value       : 'other',
+                            textfield_id: 'other-reason',
```
negarn: reminder that id should be `_` separated, also above

src/javascript/app/pages/user/account/settings/account_closure.js
Khalidxx: Okay, I'll use `$txt_other_reason`

src/templates/app/user/security/account_closure.jsx
Khalidxx: Yeah, this seems more reasonable but the design shows it with the extra `and`. I'll ask them again

src/javascript/app/pages/user/account/settings/account_closure.js
Khalidxx: `setVisibility` does not work after `fadeOut`. I think it's because `fadeOut` directly changes the css of the element (it also overwrites `!important`) whereas `setVisibility` adds/removes a class. That's maybe why @cakasuma used `.css` I guess.

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,120 @@
...
+        $error_msg         = $('#msg_form');
+        $fiat_options      = $('#change-fiat');
+        $crypto_options    = $('#crypto');
+
+        BinarySocket.wait('authorize')
+            .then((response) => {
+                const accounts = response.authorize.account_list;
+                let only_virtual, fiat, crypto;
+                accounts.forEach((account) => {
+                    if (account.currency === 'ETH' || account.currency === 'BTC' || account.currency === 'UST' || account.currency === 'LTC') {
```
cakasuma: is it possible to use `isCryptocurrency` function here from `country_base` code

src/templates/app/user/security.jsx
```diff
@@ -56,6 +56,8 @@ const Security = () => (
                 <Column url='two_factor_authentication' image='2fa' header={it.L('Two-Factor Authentication')} text={it.L('Enable two-factor authentication for an extra layer of security.')} />
 
                 <Column url='vpn_app' image='vpn' header={it.L('VPN App')} text={it.L('Establish a secure Internet connection and protect your privacy.')} />
+
+                <Column url='closure' image='detail' header={it.L('Account Closure')} text={it.L('Close your account entirely.')} />
```
cakasuma: is this approved by copywriter, and the image too?

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,93 @@
+const BinarySocket         = require('../../../../base/socket');
+const localize             = require('../../../../../_common/localize').localize;
+const Url                  = require('../../../../../_common/url');
```
negarn: extra space before `=`

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,93 @@
...
+        $error_msg;
+
+    const onLoad = () => {
+        $txt_other_reason  = $('#other_reason');
+        $closure_loading   = $('#closure_loading');
+        $closure_container = $('#closure_container');
+        $success_msg       = $('#msg_main');
+        $error_msg         = $('#msg_form');
+
+        $(form_selector).on('submit', (event) => {
```
negarn: why not also store `$form` since it's also used on line 29? then line 29 can change to `$form.find('#btn_submit')`

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,93 @@
...
+            });
+        } else {
+            setTimeout(() => { $btn_submit.removeAttr('disabled'); }, 1000);
+        }
+    };
+
+    const showFormMessage = (localized_msg) => {
+        $.scrollTo($('#reason'), 500, { offset: -20 });
+        $error_msg
+            .attr('class', 'errorfield')
```
negarn: we are still using both `error-field` and `errorfield` in this PR, are they supposed to be the same or different?

src/templates/app/user/security/account_closure.jsx
```diff
@@ -0,0 +1,150 @@
...
+            </div>
+
+            <div className='gr-no-gutter' id='closure_description'>
+                <h2 className='primary-color'>{it.L('What would you like to do?')}</h2>
+                <fieldset>
+                    <div className='gr-padding-20 gr-gutter-left gr-gutter-right'>
+                        <ClosureDescription
+                            id='change-fiat'
+                            title={it.L('Change my account currency')}
+                            subtitle={it.L('[_1]Change your fiat currency[_2] to any of the following:', `<a href="${it.url_for('user/accounts')}">`, '</a>')}
```
negarn: did we decide to leave it as is with no changes?

Khalidxx: Yeah, they were okay with leaving all the messages without any checks

src/javascript/app/pages/user/account/settings/account_closure.js
Khalidxx: They're slightly different. `error-field` has a `border-color: red` attribute, which we don't need, we just want a `color: red`

src/javascript/app/pages/user/account/settings/account_closure.js
Khalidxx: I always seem to forget this 😅

src/templates/app/user/security/account_closure.jsx
cakasuma: in addition, as per discussion with lucas and jeya. and we will redirect the link to manage account if they want to change currency. the message of they don't have or they have the account will be shown there already in the manage account page.

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,139 @@
...
+        $closure_loading   = $('#closure_loading');
+        $closure_container = $('#closure_container');
+        $success_msg       = $('#msg_main');
+        $error_msg         = $('#msg_form');
+        $crypto            = $('#crypto');
+        $fiat              = $('#change-fiat');
+        $form              = $(form_selector);
+
+        BinarySocket.wait('landing_company').then((response) => {
+            const company = response.landing_company.gaming_company || response.landing_company.financial_company;
```
cakasuma: should we add handler to check if the response is exists?

Khalidxx: What's a good way to handle that?

negarn: better use getPropertyValue()

negarn: you shouldn't fallback, check if you should use gaming or financial depending on what the current landing company of the client is (landing_company_shortcode should match)

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,141 @@
...
+                    if (isCryptocurrency(currency)) {
+                        $crypto.find('ul').append(`<li>${getCurrencyFullName(currency)}</li>`);
+                    } else {
+                        $fiat.find('ul').append(`<li>${getCurrencyFullName(currency)}</li>`);
+                    }
+                });
+                if (!$crypto.find('li').length) $crypto.setVisibility(0);
+                if (!$fiat.find('li').length) $fiat.setVisibility(0);
+            }
+            const hasMT5 = response.landing_company.mt_gaming_company ||
```
negarn: please use the function `MetaTrader.isEligible()`

negarn: it doesn't tell you whether client has mt5 or not, just that they are allowed to have mt5 or not

Khalidxx: Yeah that's all we need. Thanks!

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,140 @@
...
+        }).catch((error) => {
+            showFormMessage(error.message);
+        });
+        $closure_container.setVisibility(1);
+
+        $(form_selector).on('submit', (event) => {
+            event.preventDefault();
+            submitForm();
+        });
+        $closure_container.setVisibility(1);
```
cakasuma: duplicate line of code here, can remove line 58

src/templates/app/user/security/account_closure.jsx
```diff
@@ -0,0 +1,143 @@
...
+    title,
+}) => (
+    <div id={id} className='gr-padding-10'>
+        <h3 className='secondary-color'>{title}</h3>
+        <p>{subtitle}</p>
+        { list_items &&
+            <ul className='bullet'>
+                { list_items.map((item, idx) => (
+                    <li id={item.id} key={idx}>{item.text}</li>
+                ))
```
cakasuma: nit, indentation

Khalidxx: The `eslint` is forcing the indentation, it's a little moody today 😂

cakasuma: 😂

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,140 @@
+const MetaTrader          = require('../../../user/metatrader/metatrader');
+const getCurrencyFullName = require('../../../../common/currency').getCurrencyFullName;
+const Client              = require('../../../../base/client');
+const BinarySocket        = require('../../../../base/socket');
```
negarn: `base` paths should be above `common` path

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,140 @@
...
+        $success_msg       = $('#msg_main');
+        $error_msg         = $('#msg_form');
+        $crypto            = $('#crypto');
+        $fiat              = $('#change-fiat');
+        $form              = $(form_selector);
+
+        BinarySocket.wait('landing_company').then(() => {
+            const currencies = Client.currentLandingCompany().legal_allowed_currencies;
+            const current_currency = Client.get('currency');
+            const show_currencies = currencies.filter((val) => val !== current_currency);
```
negarn: better also filter out other currencies that the client has since we have a full list of that, but up to UI/UX. you can look at the function `getCurrencies`

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -0,0 +1,140 @@
...
+            submitForm();
+        });
+
+        $txt_other_reason.on('keyup', () => {
+            const input = $txt_other_reason.val();
+            if (input && (input.length < 5 || input.length > 250)) {
+                $txt_other_reason.addClass('error-field');
+                $error_msg
+                    .addClass('error-field')
+                    .html('The reason should be between 5 and 250 characters')
```
negarn: this should be `.text(` instead, there are no html tags in the string to need a `.html(`. also you need a `localize(` wrapper around the text. so it should be `.text(localize('The reason should be between 5 and 250 characters'))`

Khalidxx: Oh alright 👍

PR #5537, opened 2019-05-29
Use date_expiry instead of exit_tick_time 

easteregg:


src/javascript/app/pages/trade/charts/highchart.js
```diff
@@ -602,7 +602,7 @@ const Highchart = (() => {
     const endContract = () => {
         if (chart && !stop_streaming) {
             drawLineX({
-                value     : (is_sold_before_expiry ? sell_time : end_time),
+                value     : (is_sold_before_expiry && exit_tick_time >= end_time ? sell_time : end_time),
```
negarn: please put brackets around `exit_tick_time >= end_time`

negarn: is `is_sold_before_expiry` not reliable? we are using it in other places too, should we fix that instead?

easteregg: From what I understand this is a new logic, not relevant to what `is_sold_before_expiry` implies. We are looking for contracts that their `date_expiry` field are `odd`, and happen 1 second after `exit_tick_time`. I am not sure if changing the `is_sold_before_expiry` wouldn't affect other contract types.

negarn: I'm confused, shouldn't we always use date_expiry for contracts that expire at date_expiry instead of before? so end_time here should be date_expiry, not exit_tick_time, if it is, then you are essentially comparing exit_tick_time to exit_tick_time in the `>=` condition

PR #5556, opened 2019-06-07
Khalid/updated package

Khalidxx:
Packages updated: 
- ws -> `v7.0.0`
- js-yaml -> `v3.13.1`
- fstream -> `v1.0.12`


package.json
```diff
@@ -105,7 +105,7 @@
     "webpack-dev-server": "3.1.14",
     "webpack-manifest-plugin": "2.0.4",
     "webpack-merge": "4.1.4",
-    "ws": "1.1.5"
+    "ws": "^7.0.0"
```
negarn: please use the exact version, without `^`

Khalidxx: Okay

PR #5562, opened 2019-06-10
sash/remove daily report block

sashcy:


src/templates/static/home.jsx
```diff
@@ -353,20 +357,15 @@ const Home = () => {
                 </div>
             </div>
 
-            <div className='container gr-padding-30 gr-child'>
+            <div className='container gr-padding-30 gr-child academy-section'>
```
negarn: is this class used anywhere?

sashcy: ah, we had some design changes and I forgot to remove this class

PR #5573, opened 2019-06-12
ahkuma/utm_email_api

ahkuma:


src/javascript/app/pages/new_account.js
```diff
@@ -25,10 +27,25 @@ const NewAccount = (() => {
...
+
+            const url_parameters_req = {
+                request_field: 'url_parameters',
+                value        : { utm_source: TrafficSource.getSource(utm_data) },
+            };
+
+            if (utm_data.utm_campaign) url_parameters_req.value.utm_medium = utm_data.utm_medium;
+            if (utm_data.utm_campaign) url_parameters_req.value.utm_campaign = utm_data.utm_campaign;
+            const affiliate_token = Cookies.getJSON('affiliate_tracking');
+            if (affiliate_token) url_parameters_req.value.affiliate_token = affiliate_token.t;
```
negarn: is it better to write it all together? i.e.

```
            const affiliate_token = Cookies.getJSON('affiliate_tracking');
            const url_parameters_req = {
                request_field: 'url_parameters',
                value        : { 
                        utm_source: TrafficSource.getSource(utm_data),
                        ...( utm_data.utm_campaign && { 
                                utm_medium   : utm_data.utm_medium,
                                utm_campaign: utm_data.utm_campaign,
                        }),
                        ...( affiliate_token && { affiliate_token: affiliate_token.t }),
                },
            };
```

src/javascript/static/pages/home.js
```diff
@@ -24,11 +26,26 @@ const Home = (() => {
...
+                request_field: 'url_parameters',
+                value        : { utm_source: TrafficSource.getSource(utm_data) },
+            };
+
+            if (utm_data.utm_campaign) url_parameters_req.value.utm_medium = utm_data.utm_medium;
+            if (utm_data.utm_campaign) url_parameters_req.value.utm_campaign = utm_data.utm_campaign;
+            const affiliate_token = Cookies.getJSON('affiliate_tracking');
+            if (affiliate_token) url_parameters_req.value.affiliate_token = affiliate_token.t;
+
+            req.push(url_parameters_req);
```
negarn: same here. better yet, create a common function that gives back an object of url_parameters_req so we don't have to repeat it.

ahkuma: Where do you think I should put that function?

negarn: IMO in traffic_source would be fine

ahkuma: Also since the entire `req` is similar, we could just use that and name it something like `getVerifyEmailFormRequest`

ahkuma: `const getVerifyEmailFormRequest = () => {
    const utm_data = TrafficSource.getData();
    const affiliate_token = Cookies.getJSON('affiliate_tracking');

    return [
        { selector: '#email', validations: ['req', 'email'], request_field: 'verify_email' },
        { request_field: 'type', value: 'account_opening' },
        {
            request_field: 'url_parameters',
            value        : {
                utm_source: TrafficSource.getSource(utm_data),
                ...(utm_data.utm_campaign && {
                    utm_medium  : utm_data.utm_medium,
                    utm_campaign: utm_data.utm_campaign,
                }),
                ...(affiliate_token && { affiliate_token: affiliate_token.t }),
            },
        },
    ];
};`

I have this, I don't think we could put it there since it isn't fully related to traffic source but rather verify_email

PR #5576, opened 2019-06-13
ahkuma/less_req_demo_mt5

ahkuma:


src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -119,11 +119,14 @@ const MetaTraderConfig = (() => {
...
             $new_account_financial_authenticate_msg.setVisibility(0);
             const is_virtual = Client.get('is_virtual');
+            const is_demo = accounts_info[acc_type].is_demo;
 
             if (!Client.get('currency')) {
                 resolve($messages.find('#msg_set_currency').html());
-            } else if (is_virtual && !accounts_info[acc_type].is_demo) { // virtual clients can only open demo MT accounts
+            } else if (is_virtual && !is_demo) { // virtual clients can only open demo MT accounts
                 resolve(needsRealMessage());
+            } else if (is_demo) {
```
negarn: why not move it above so we don't have to check `!is_demo`? this can be moved to line 125

ahkuma: changed!

src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -119,10 +119,21 @@ const MetaTraderConfig = (() => {
...
+                BinarySocket.wait('get_settings').then(() => {
+                    const response_get_settings = State.getResponse('get_settings');
+                    if (response_get_settings.residence === 'United Kingdom' && response_get_settings.date_of_birth === null) {
+                        $messages.find('#msg_age_verification').setVisibility(1);
+                        resolve($messages.find('#msg_age_verification').html());
+                    }
+                });
+
+                resolve($messages.find('#msg_age_verification').html());
```
negarn: @ahkuma won't this resolve without waiting for get_settings? how will the get settings resolve get called then? doesn't a promise only resolve once?

src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -119,10 +119,21 @@ const MetaTraderConfig = (() => {
...
             const is_virtual = Client.get('is_virtual');
+            const is_demo = accounts_info[acc_type].is_demo;
 
             if (!Client.get('currency')) {
                 resolve($messages.find('#msg_set_currency').html());
-            } else if (is_virtual && !accounts_info[acc_type].is_demo) { // virtual clients can only open demo MT accounts
+            } else if (is_demo) {
+                BinarySocket.wait('get_settings').then(() => {
+                    const response_get_settings = State.getResponse('get_settings');
+                    if (response_get_settings.residence === 'United Kingdom' && response_get_settings.date_of_birth === null) {
```
negarn: please never check the full display name. you should check `client.get('residence') === 'gb'` instead, then no need to wait for get_settings either

src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -119,10 +119,21 @@ const MetaTraderConfig = (() => {
...
+                        $messages.find('#msg_age_verification').setVisibility(1);
```
negarn: no, to check age verification you need to wait for `get_account_status`, it's a check in the API, not necessarily for empty DOB, also unverified DOB. check `/age_verification/` flag exists in `get_accout_status.status` or not

src/templates/app/user/metatrader.jsx
```diff
@@ -587,6 +587,7 @@ const Metatrader = () => (
                         <div id='msg_switch'>{it.L('To perform this action, please switch to your [_1] Real Account.', it.website_name)}</div>
                         <div id='msg_switch_financial'>{it.L('To perform this action, please switch to your [_1] Financial account', it.website_name)}</div>
                         <div id='msg_upgrade'>{it.L('To perform this action, please [_1]upgrade to [_2] Real Account[_3].', `<a href="${it.url_for('new_account/realws')}">` , it.website_name, '</a>')}</div>
+                        <div id='msg_age_verification'>{it.L('Please provide your [_1]date of birth[_2].', `<a href="${it.url_for('new_account/realws')}">`, '</a>')}</div>
```
negarn: it's not about providing date of birth, it's about submitting documents to verify your age

src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -119,10 +119,23 @@ const MetaTraderConfig = (() => {
...
+            const is_demo = accounts_info[acc_type].is_demo;
 
             if (!Client.get('currency')) {
                 resolve($messages.find('#msg_set_currency').html());
-            } else if (is_virtual && !accounts_info[acc_type].is_demo) { // virtual clients can only open demo MT accounts
+            } else if (is_demo) {
+                if (Client.get('residence') === 'gb') {
+                    resolve($messages.find('#msg_upgrade').html());
+
+                    BinarySocket.wait('get_account_status').then((response) => {
```
negarn: if the code gets resolved above does it reach this line?

ahkuma: No it doesn't, but we need users to upgrade their account first, since ProveID uses details entered during real account upgrade to automatically check if user is old enough. Otherwise, we will need users to manually send us documents to verify. That's why we don't show the verify document message until the account is upgraded.

src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -209,8 +224,6 @@ const MetaTraderConfig = (() => {
                             if (is_ok) resolve();
                             else resolveWithMessage();
                         });
-                    } else {
-                        resolve();
```
negarn: why was this removed?

ahkuma: @negarn Since we covered rest of the account types, isn't it checking for demo accounts, which we added so this would be a duplicate.

src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -153,12 +168,19 @@ const MetaTraderConfig = (() => {
...
                         $message.find('.maltainvest').setVisibility(1);
-                        resolveWithMessage();
+                        
+                        BinarySocket.wait('get_account_status').then((response) => {
+                            if (Client.get('residence') === 'gb' && !/age_verification/.test(response.get_account_status.status)) {
+                                $message.find('.authenticate').setVisibility(1);
+                            }
+
+                            resolveWithMessage();
+                        });
```
negarn: why don't we wrap `Client.get('residence') === 'gb'` around this too so we don't wait for get account status if it's not needed?

src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -205,12 +231,14 @@ const MetaTraderConfig = (() => {
                                 showCitizenshipMessage();
                                 is_ok = false;
                             }
+                            if (Client.get('residence') === 'gb' && !/age_verification/.test(response_get_account_status.status)) {
```
negarn: since we're using `Client.get('residence') === 'gb' && !/age_verification/.test(response_get_account_status.status)` multiple times should we add it as a function in this file and call it?

ahkuma: Personally, I think it is simple enough that we don't need to do it.

src/javascript/app/pages/user/metatrader/metatrader.config.js
ahkuma: fixed

PR #5598, opened 2019-06-19
Khalid/Tick trade overlap issue

Khalidxx:
Fixed an issue where the popup view in Profit Table has overlapping components for tick trades


src/sass/app/sell_popup.scss
```diff
@@ -95,6 +95,9 @@
                         }
                     }
                 }
+                #table_digits_container {
+                    margin-bottom: 20px;
```
cakasuma: is it always 20px?

Khalidxx: Yeah, and it looks okay on both phone and desktop. Should I use `vh` instead?

PR #5608, opened 2019-06-24
Khalid/remove authentication banner

Khalidxx:
Authentication banner does not show after the authentication documents are successfully uploaded


src/javascript/app/base/header.js
```diff
@@ -353,7 +352,7 @@ const Header = (() => {
                 const notified = check_statuses.some((check_type) => {
                     if (validations[check_type]()) {
                         // show MF retail message on Trading pages only
-                        if (check_type === 'mf_retail' && !(State.get('is_trading') || State.get('is_mb_trading'))) {
+                        if (hasStatus('document_under_review') || (check_type === 'mf_retail' && !(State.get('is_trading') || State.get('is_mb_trading')))) {
```
negarn: better to add it in a different if with a comment to explain why we're returning false

PR #5616, opened 2019-06-26
Khalid/MT5 Norway

Khalidxx:
Added a check to get `gaming_company` if a country does not have a `financial_company` for MT5 sign up (so far, only applies to Norway)


src/javascript/app/base/header.js
```diff
@@ -254,13 +254,13 @@ const Header = (() => {
         BinarySocket.wait('authorize', 'landing_company').then(() => {
             let get_account_status,
                 status;
-            const is_svg = Client.get('landing_company_shortcode') === 'svg';
+            const is_svg       = Client.get('landing_company_shortcode') === 'svg';
+            const requirements = (State.getResponse('landing_company.financial_company') || State.getResponse('landing_company.gaming_company')).requirements;
```
negarn: better see which landing company the current client is and use that, instead of financial by default and falling back to gaming. unless there's a reason we need to check general instead of client-specific?

Khalidxx: I thought of using `getLandingCompanyValue`, which works well even though it's verbose. Would that be better?

negarn: yes, let's try that

PR #5623, opened 2019-06-28
Khalid/UST decimal points

Khalidxx:
On money transfer, UST currency will show 2 decimal places

### DIFF
--- a/src/javascript/_common/base/currency_base.js
+++ b/src/javascript/_common/base/currency_base.js
@@ -4,7 +4,7 @@ const getPropertyValue = require('../utility').getPropertyValue;
 
 let currencies_config = {};
 
-const getTextFormat = (number, currency) => `${currency} ${addComma(number, getDecimalPlaces(currency), isCryptocurrency(currency))}`;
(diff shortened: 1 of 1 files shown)


src/javascript/_common/base/currency_base.js
```diff
@@ -4,7 +4,7 @@ const getPropertyValue = require('../utility').getPropertyValue;
 
 let currencies_config = {};
 
-const getTextFormat = (number, currency) => `${currency} ${addComma(number, getDecimalPlaces(currency), isCryptocurrency(currency))}`;
+const getTextFormat = (number, currency) => `${currency} ${addComma(number, getDecimalPlaces(currency), isCryptocurrency(currency) && currency !== 'UST')}`;
```
easteregg: If 2 decimal places is true for stable coins, maybe we should move this check inside `getDecimalPlaces` so other users of this function may benefit as well.

negarn: we shouldn't hardcode this, if it's based on decimal places for that crypto currency we should check that from the currency config object

PR #5626, opened 2019-07-01
Amam khalid/ust decimals

cakasuma:
fix Decimal points in `getTextFormat` function by mishandled order in `addComma` function


src/javascript/_common/base/currency_base.js
```diff
@@ -30,12 +30,12 @@ const formatCurrency = currency => `<span class="symbols ${(currency || '').toLo
 
 const addComma = (num, decimal_points, is_crypto) => {
     let number = String(num || 0).replace(/,/g, '');
-    if (typeof decimal_points !== 'undefined') {
-        number = (+number).toFixed(decimal_points);
-    }
     if (is_crypto) {
         number = parseFloat(+number);
```
negarn: why are we doing parsefloat here? if we move it up, it's like we're negating parsefloat

cakasuma: imo, this parsefloat is unnecessary. it was there since before, and I am not really sure why are we doing parsefloat for crypto. The problem was because, after the number get `toFixed` and then get `parseFloat`, `parseFloat` make decimal digits gone. that is why i just put that on top, maybe it was used for certain cases

PR #5630, opened 2019-07-08
kelly/hide_DE

kellybinary:


src/javascript/_common/language.js
```diff
@@ -9,7 +9,7 @@ const Language = (() => {
     const all_languages = {
         ACH  : 'Translations',
         EN   : 'English',
-        DE   : 'Deutsch',
+        // DE   : 'Deutsch',
```
negarn: can you add a todo comment to uncomment both of these when we want to enable german again?

PR #5631, opened 2019-07-08
amam/remove_bo_ban

cakasuma:
** Change **
- Remove MFSA component
- Remove MFSA Import
- Remove EU country check for MFSA
- Remove bo ban on professional not requested when purchase


src/javascript/app/pages/trade/purchase.js
```diff
@@ -122,10 +122,9 @@ const Purchase = (() => {
 
                                 message = prepareConfirmationErrorCta(message_text, button_text, true);
                             } else {
-                                const message_text = localize('In the EU, financial binary options are only available to professional investors.');
                                 const button_text  = localize('Apply now as a professional investor');
 
-                                message = prepareConfirmationErrorCta(message_text, button_text);
+                                message = prepareConfirmationErrorCta('', button_text);
```
negarn: I don't think we need to remove this message from purchase error. @ashkanx ?

ashkanx: no, keep it. only homepage error

PR #5658, opened 2019-07-22
amam/add_banner_grid

mustofa-binary:
- Add banner redirected to the binary grid page
- conditions:
-- only for non-EU clients both logged in and not logged in
-- show to every device


src/javascript/app/pages/trade/tradepage.js
```diff
@@ -77,6 +78,17 @@ const TradePage = (() => {
...
 
         ViewPopup.viewButtonOnClick('#contract_confirmation_container');
+
+        displayBanner();
+    };
+
+    const displayBanner = () => {
+        BinarySocket.wait('website_status', 'authorize', 'landing_company').then(() => {
+            if (!isEuCountry()) {
+                const $binary_grid_banner = $('#binary_grid_banner');
```
negarn: better use javascript since we're just setting visibility. so `getElementById('binary_grid_banner').setVisibility(1);` (feel free to look up the `getElementById` function versus the `document.getElementById`)

PR #5691, opened 2019-08-06
Amam/onfido

mustofa-binary:
** Changes **

- added onfido
- handle token
- handle the response and show messages
- onfido handle complete
- added tab selector
- added mock response


package.json
```diff
@@ -83,7 +83,7 @@
     "grunt-webpack": "3.1.3",
     "jsdom": "9.12.0",
     "jsdom-global": "2.1.1",
-    "load-grunt-config": "1.0.2",
+    "load-grunt-config": "^3.0.0",
```
mahdipourziaei-binary: Why was this package updated?

mustofa-binary: there was a high vulnerability and my hand was itchy to update it 😂 . will revert it for now

src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -8,49 +10,83 @@ const isImageType         = require('../../../../_common/image_utility').isImage
...
-                    const language_based_link = ['ID', 'RU', 'PT'].includes(language) ? `_${language}` : '';
-                    const $not_authenticated  = $('#not_authenticated');
-                    $not_authenticated.setVisibility(1);
-                    let link = Url.urlForCurrentDomain(`https://marketing.binary.com/authentication/Authentication_Process${language_based_link}.pdf`);
-                    if (Client.isAccountOfType('financial')) {
-                        $('#not_authenticated_financial').setVisibility(1);
-                        link = Url.urlForCurrentDomain('https://marketing.binary.com/authentication/MF_Authentication_Process.pdf');
-                    }
-                    $not_authenticated.find('.learn_more').setVisibility(1).find('a').attr('href', link);
+    const simulateCallForApiKey = () => new Promise((resolve, reject) => {
```
mahdipourziaei-binary: Are we still simulating call?

mustofa-binary: will change the name

package.json
mahdipourziaei-binary: No worries. No need to revert, just trace it to the places it's used and just make sure no build process is negatively affected. If it's too much to trace and identify then you can revert. But I believe @brandon-binary is working on updating packages.

src/templates/app/user/metatrader.jsx
```diff
@@ -303,7 +303,7 @@ const Metatrader = () => (
                                         </TypeGroup>
                                     </div>
                                     <p id='new_account_msg' className='notice-msg center-text invisible' />
-                                    <p id='new_account_financial_authenticate_msg' className='invisible notice-msg hint'>{it.L('You may proceed but please [_1]authenticate[_2] your account within 10 days to continue trading, otherwise your account will be disabled and any open positions will be closed.', `<a href="${it.url_for('user/authenticate')}">`, '</a>')}</p>
+                                    <p id='new_account_financial_authenticate_msg' className='invisible notice-msg hint'>{it.L('You may proceed but please authenticate your account within 10 days to continue trading, otherwise your account will be disabled and any open positions will be closed.')}</p>
```
mahdipourziaei-binary: Why was the link removed?

src/javascript/app/base/client.js
```diff
@@ -76,7 +76,7 @@ const Client = (() => {
     const doLogout = (response) => {
         if (response.logout !== 1) return;
         removeCookies('login', 'loginid', 'loginid_list', 'email', 'residence', 'settings'); // backward compatibility
-        removeCookies('reality_check', 'affiliate_token', 'affiliate_tracking');
+        removeCookies('reality_check', 'affiliate_token', 'affiliate_tracking', 'onfido_token', 'is_onfido_unsupported');
```
mahdipourziaei-binary: `is_onfido_unsupported` is not used anywhere. Can we remove this?

src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -116,6 +140,36 @@ const Authenticate = (() => {
         enableDisableSubmit();
     };
 
+    const onFileSelectedUns = (event) => {
```
mahdipourziaei-binary: does `onFileSelectedUns` mean `onFileSelectedUnsupported`? If yes, it's okay to have long names if necessary, can you please rename that so it's clearer?

Edit: I see that it's used a lot in the file. I guess we can keep it. Perhaps you can leave one line of comment after the `require()` lines to let people know that it's shortened to `Uns`?

src/templates/app/user/metatrader.jsx
mustofa-binary: this message shown before mt5 signup, ergo, authenticate page is not available

PR #5694, opened 2019-08-08
update binary-style and fix 2 vulnerabilities

brandon-binary:


package.json
```diff
@@ -83,7 +83,7 @@
     "grunt-webpack": "3.1.3",
     "jsdom": "9.12.0",
     "jsdom-global": "2.1.1",
-    "load-grunt-config": "1.0.2",
+    "load-grunt-config": "^3.0.0",
```
negarn: why is this updated?

PR #5696, opened 2019-08-09
enable View available payment methods for crypto cashier

brandon-binary:


src/templates/app/cashier/index.jsx
```diff
@@ -110,7 +110,7 @@ const Cashier = () => (
                     <span className='invisible normal_currency client_logged_out'>{it.L('Deposit or withdraw to your account via bank wire, credit card, or e-wallet.')}</span>
                     <span className='invisible crypto_currency'>{it.L('Manage the funds in your cryptocurrency account.')}</span>
                     &nbsp;
-                    <a className='invisible normal_currency client_logged_out' href={it.url_for('cashier/payment_methods')} id='view_payment_methods'>
+                    <a className='invisible normal_currency crypto_currency client_logged_out' href={it.url_for('cashier/payment_methods')} id='view_payment_methods'>
```
negarn: then should we just make it visible and remove all these classes?

brandon-binary: I've readded the class names so it doesn't show while the cashier is loading, without it, it shows when other things are not ready and looks odd.

PR #5724, opened 2019-08-21
sash/update job ad description

sash-binary:
Career Page Update: Compliance and Payments Job Ad text updated

### DIFF
--- a/src/templates/static/about/job_details.jsx
+++ b/src/templates/static/about/job_details.jsx
@@ -947,7 +947,7 @@ const JobDetails = () => (
                                 text={it.L('Minimum qualifications:')}
                                 className='bullet'
                                 items={[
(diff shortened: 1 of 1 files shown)


src/templates/static/about/job_details.jsx
```diff
@@ -923,36 +923,42 @@ const JobDetails = () => (
...
+                                    { text: it.L('Research new payment systems that are available in the market') },
                                 ]}
                             />
 
                             <UlText
-                                text={it.L('Preferences:')}
+                                text={it.L('Minimum qualifications:')}
                                 className='bullet'
                                 items={[
                                     { text: it.L('A bachelor\'s degree in commerce or any business-related field, IT-related field or criminology') },
```
nazanin-binary: there is a `,` missing, `IT-related field, or criminology`

PR #5725, opened 2019-08-22
negar/hide_mt5_for_iom

negarn:
remove exceptions for iom and gb


ashkanx:
@negarn please resolve the conflict and we can merge

PR #5747, opened 2019-09-12
MF Tweak

brandon-binary:


src/javascript/_common/base/client_base.js
```diff
@@ -312,6 +312,8 @@ const ClientBase = (() => {
         return (is_from_crypto ? !is_to_crypto : is_to_crypto);
     };
 
+    const isMFAccount = (loginid) => /^MF/.test(loginid || current_loginid);
```
negarn: please don't test the login id. we don't hardcode these. we should always rely on landing company. you can instead use isAccountOfType('financial')

src/javascript/app/base/logged_in.js
```diff
@@ -43,7 +43,10 @@ const LoggedInHandler = (() => {
             if (set_default) {
                 const lang_cookie = urlLang(redirect_url) || Cookies.get('language');
                 const language    = getLanguage();
-                redirect_url      = Client.defaultRedirectUrl();
+                // const current_loginid = getPropertyValue(response, ['authorize', 'loginid']);
```
negarn: why is this here?

src/javascript/app/pages/trade/tradepage.js
```diff
@@ -26,57 +26,57 @@ const TradePage = (() => {
...
-            Header.displayAccountStatus();
-            if (Client.get('is_virtual')) {
-                Header.upgradeMessageVisibility(); // To handle the upgrade buttons visibility
-                // if not loaded by pjax, balance update function calls TopUpVirtualPopup
-                if (State.get('is_loaded_by_pjax')) {
-                    BinarySocket.wait('balance').then(() => {
-                        TopUpVirtualPopup.init(State.getResponse('balance.balance'));
-                    });
-                }
+        if (!Client.isMFAccount()) {
```
negarn: please use a return if client is financial instead so the indentation of the below lines don't change

negarn:
@brandon-binary please resolve the conflict

brandon-binary:
@negarn Fixed

src/javascript/app/base/binary_loader.js
```diff
@@ -151,7 +154,7 @@ const BinaryLoader = (() => {
             return;
         }
 
-        const div_container = createElement('div', { class: 'logged_out_title_container', html: content.getElementsByTagName('h1')[0] || '' });
+        const div_container = createElement('div', { class: 'logged_out_title_container', html: Client.isAccountOfType('financial') ? '' : content.getElementsByTagName('h1')[0] });
```
negarn: the `|| ''` for the second condition should remain for pages that don't have header.

PR #5762, opened 2019-09-26
amam/show_authentication

mustofa-binary:
- show authentication page in setting
- add new message when they are not required to authentication
- support mobile


src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -915,7 +914,7 @@ const Authenticate = (() => {
         const { identity, document, needs_verification } = authentication_status;
 
         if (identity.status === 'none' && document.status === 'none' && !needs_verification.length) {
```
oskar-binary:
```suggestion
        if (!is_high_risk_client) {
```

PR #5763, opened 2019-09-26
negar/mt5_vanuatu_to_svg

negarn:
change vanuatu to svg but handle old accounts being vanuatu


src/javascript/app/pages/user/account/settings/financial_assessment.js
```diff
@@ -49,7 +49,7 @@ const FinancialAssessment = (() => {
         }
 
         // display Trading Experience only for financial & MT5 financial accounts
-        const is_mt5_financial = /labuan_advanced|real_vanuatu_(standard|advanced|mamm_advanced)/.test(localStorage.getItem('financial_assessment_redirect'));
+        const is_mt5_financial = /labuan_advanced|real_svg_standard|real_vanuatu_advanced|real_vanuatu_mamm_advanced/.test(localStorage.getItem('financial_assessment_redirect'));
```
negarn: @brandon-binary then we should also be able to remove it from here since FA is also not needed for svg

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -340,7 +340,7 @@ const PersonalDetails = (() => {
                 }
                 const get_settings        = data.get_settings;
                 const is_tax_req          = +State.getResponse('landing_company.config.tax_details_required') === 1;
-                const is_for_mt_financial = /real_vanuatu_(standard|advanced)/.test(redirect_url);
+                const is_for_mt_financial = /real_svg_standard/.test(redirect_url);
```
negarn: and I think this should change to `real_vanuatu_advanced`, we still have that group for financial, right?

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -340,7 +340,7 @@ const PersonalDetails = (() => {
                 }
                 const get_settings        = data.get_settings;
                 const is_tax_req          = +State.getResponse('landing_company.config.tax_details_required') === 1;
-                const is_for_mt_financial = /real_vanuatu_(standard|advanced)/.test(redirect_url);
+                const is_for_mt_financial = /real_svg_standard|labuan_advanced/.test(redirect_url);
```
negarn: I think we don't need `real_svg_standard` here, since they have no validation, right?

src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -374,7 +379,11 @@ const MetaTraderConfig = (() => {
                     resolve(needsRealMessage());
                 } else if (accounts_info[acc_type].account_type === 'financial') {
                     BinarySocket.send({ get_account_status: 1 }).then(() => {
-                        resolve(!isAuthenticated() ? $messages.find('#msg_authenticate').html() : '');
+                        if (getMTFinancialAccountType(acc_type) === 'advanced' && isAuthenticationPromptNeeded()) {
```
negarn: can we change this back to `!isAuthenticated`? prompt to authenticate will only show for high risk

src/javascript/app/pages/user/metatrader/metatrader.ui.js
```diff
@@ -566,8 +566,8 @@ const MetaTraderUI = (() => {
     };
 
     const showHideFinancialAuthenticate = (acc_type) => {
-        if (MetaTraderConfig.hasAccount(acc_type) && accounts_info[acc_type].account_type === 'financial') {
-            $('#financial_authenticate_msg').setVisibility(!MetaTraderConfig.isAuthenticated());
+        if (MetaTraderConfig.hasAccount(acc_type)) {
+            $('#financial_authenticate_msg').setVisibility(MetaTraderConfig.isAuthenticationPromptNeeded());
```
negarn: same here, can we check that mt account type is advanced and revert to is authenticated check?

src/javascript/app/pages/user/metatrader/metatrader.ui.js
```diff
@@ -572,6 +572,10 @@ const MetaTraderUI = (() => {
     };
 
     const setCounterpartyAndJurisdictionTooltip = ($el, acc_type) => {
+        if (/real_vanuatu_standard/.test(acc_type)) {
```
negarn: @brandon-binary can you add a comment to the code as to why we're doing this?

PR #5767, opened 2019-09-26
Update deposit confirmation message title

mahdiaryayi-fs:
Done


src/javascript/app/pages/cashier/deposit_withdraw.js
```diff
@@ -219,7 +219,7 @@ const DepositWithdraw = (() => {
                     id                : 'deposit_currency_change_popup_container',
                     ok_text           : localize('Yes I\'m sure'),
                     cancel_text       : localize('Cancel'),
-                    localized_title   : localize('Are you sure?'),
+                    localized_title   : localize('Are you sure you want to deposit in [_1]?', [Client.get('currency')]),
```
oskar-binary: Might as well put it in a `const` since we are using the client currency below as well (line 231)?

```suggestion
                    localized_title   : localize('Are you sure you want to deposit in [_1]?', [client_currency]),
```

mahdiaryayi-fs: @oskar-binary  Changes applied

PR #5776, opened 2019-09-30
mahdi.aryayi/Added binary grid link to trade menu and platforms page

mahdiaryayi-fs:
Added binary grid link


src/templates/app/_layout/header.jsx
```diff
@@ -103,6 +103,7 @@ const Header = () => (
                                                 { text: it.L('Binary Bot'),   href: 'https://bot.binary.com',       target: '_blank' },
                                                 { text: it.L('MetaTrader 5'), href: it.url_for('user/metatrader'),  className: 'invisible mt_visibility' },
                                                 { text: it.L('Deriv'),        href: 'https://deriv.app/?utm_source=binary&utm_medium=referral', target: '_blank', dataShow: '-eucountry' },
+                                                { text: it.L('Binary Grid'),  href: it.url_for('binary-grid'),      target: '_blank',  dataShow: '-eucountry' },
```
negarn: @mahdiaryayi-fs we have two header files, please make the same change in both

mahdiaryayi-fs: @negarn Changes are applied

PR #5790, opened 2019-10-13
Fix DE/TH languages bug

mahdiaryayi-fs:
Redirection their infinite loop 404 pages to EN home page

### DIFF
--- a/src/root_files/_common/404.html
+++ b/src/root_files/_common/404.html
@@ -40,8 +40,9 @@
     function redirect(lang) {
         var base = /^(.+?).github.io$/.test(window.location.hostname) ? window.location.origin + '/binary-static' : window.location.origin;
 
-        // TODO: REMOVE/CHANGE this if statement after addition of DE or TH languages
-        if (lang.toLowerCase() === 'de' || lang.toLowerCase() === 'th') {
+        // TODO: REMOVE/CHANGE this after addition of DE or TH languages
+        var unsupported_languages = ['de', 'th']
+        if (unsupported_languages.indexOf(lang.toLowerCase()) >= 0) {
             lang = 'en'
             window.location.href = base + '/' + lang.toLowerCase() + '/' + 'home.html';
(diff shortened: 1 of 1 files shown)


src/root_files/_common/404.html
```diff
@@ -40,6 +40,13 @@
     function redirect(lang) {
         var base = /^(.+?).github.io$/.test(window.location.hostname) ? window.location.origin + '/binary-static' : window.location.origin;
 
+        // TODO: REMOVE/CHANGE this if statement after addition of DE or TH languages
+        if (lang.toLowerCase() === 'de' || lang.toLowerCase() === 'th') {
```
negarn: maybe instead of checking unsupported languages we can check supported languages?

negarn: or if we want to keep this as checking unsupported ones, it could be an array and check indexof instead of individually checking each of them.

mahdiaryayi-fs: @negarn As we talked before somewhere else the default language is coming from backend and I have to check for unsupported ones. I will make changes for array checking

src/root_files/_common/404.html
```diff
@@ -40,6 +40,14 @@
     function redirect(lang) {
         var base = /^(.+?).github.io$/.test(window.location.hostname) ? window.location.origin + '/binary-static' : window.location.origin;
 
+        // TODO: REMOVE/CHANGE this after addition of DE or TH languages
+        var unsupported_languages = ['de', 'th']
+        if (unsupported_languages.indexOf(lang.toLowerCase()) >= 0) {
+            lang = 'en'
+            window.location.href = base + '/' + lang.toLowerCase() + '/' + 'home.html';
+            return
```
negarn: please use `;` :) line 44, 46, 48

mahdiaryayi-fs: @negarn done!

PR #5796, opened 2019-10-18
brandon/remove_limit

brandon-binary:


src/javascript/app/base/page.js
```diff
@@ -102,7 +102,10 @@ const Page = (() => {
         }
         if (Client.isLoggedIn()) {
             BinarySocket.wait('authorize', 'website_status', 'get_account_status').then(() => {
-                RealityCheck.onLoad();
+                if (!Client.isAccountOfType('financial')) {
+                    RealityCheck.onLoad();
```
negarn: @brandon-binary we should ask API to remove `has_reality_check` from MF accounts

brandon-binary: @negarn Removed hard coded reality check filter.

PR #5800, opened 2019-10-22
brandon/crypto_payment_methods

brandon-binary:


src/javascript/app/pages/cashier/cashier.js
```diff
@@ -113,6 +113,27 @@ const Cashier = (() => {
         el_acc_currency.setVisibility(show_current_currency);
     };
 
+    const setCryptoMinimumWithdrawal = () => {
+        BinarySocket.wait('website_status').then((response) => {
+            $('#cryptocurrency tbody tr').each(function () {
+                const row = $(this);
```
negarn: please use `$row` naming so we know that it's a jQuery element

src/javascript/app/pages/cashier/cashier.js
```diff
@@ -113,6 +113,27 @@ const Cashier = (() => {
...
+                const columns = row.find('td:nth-child(2) div:nth-child(2)');
```
negarn: same here, `$columns`

src/javascript/app/pages/cashier/cashier.js
```diff
@@ -113,6 +113,27 @@ const Cashier = (() => {
...
+
+                const shortname = columns.find('p:nth-child(1)').text();
+                const el_crypto_min_withdrawal = columns.find('p:nth-child(3)');
```
negarn: same here, should be `$crypto_min_withdrawal`

src/templates/app/cashier/payment_methods.jsx
```diff
@@ -563,8 +564,8 @@ const PaymentMethods = () => {
                                         { td: it.L('Tether is a blockchain-based cryptocurrency whose cryptocoins in circulation are backed by an equivalent amount of traditional fiat currencies. For more information, please visit [_1].', `${createLink('https://www.tether.to')}`) },
                                         { td_list: [
                                             { text: 'UST' },
-                                            { text: '1' },
-                                            { text: '1' },
+                                            { text: '-' },
+                                            { text: '-' },
```
negarn: instead of `-` let's leave the withdrawals empty? since they will be populated after response comes, so that user doesn't think it's not available

brandon-binary:
@negarn Mostly updated

PR #5811, opened 2019-11-04
Bala/Not supported versions - text & style change

balakrishna-binary:


src/sass/_common/common.scss
```diff
@@ -609,6 +609,7 @@ ul.bullet.gr-row > li {
 .foot-note {
     font-size: $FONT_SIZE_XS;
     color: $COLOR_GRAY_SHADE;
+    display: inline-block;
```
negarn: this class is also used in trading times page, have you checked to make sure adding this doesn't break the styling there?

balakrishna-binary: Yeah, I've checked there and it didn't break anything

PR #5829, opened 2019-12-04
mamali/check-for-mt5-api-error

reza-fs:


src/javascript/_common/base/client_base.js
```diff
@@ -316,7 +316,7 @@ const ClientBase = (() => {
 
     const canChangeCurrency = (statement, mt5_login_list, is_current = true) => {
         const currency             = get('currency');
-        const has_no_mt5           = mt5_login_list.length === 0;
+        const has_no_mt5           = mt5_login_list ? mt5_login_list.length === 0 : true;
```
negarn: can we change this to:
```suggestion
        const has_no_mt5           = !mt5_login_list || !mt5_login_list.length;
```
I think it's less confusing

src/javascript/app/pages/user/metatrader/metatrader.js
```diff
@@ -22,7 +22,12 @@ const MetaTrader = (() => {
         BinarySocket.wait('landing_company', 'get_account_status', 'statement').then(async () => {
             if (isEligible()) {
                 if (Client.get('is_virtual')) {
-                    await addAllAccounts();
+                    try {
+                        await addAllAccounts();
+                    } catch (error) {
+                        // eslint-disable-next-line no-console
+                        console.warn(error.message);
```
negarn: why are we just warning in console?

reza-fs: I wanted to ignore it, but I thought I left a trace for further debugging. I can just do nothing in the error block, what do you suggest?

src/javascript/_common/base/client_base.js
reza-fs: Sure. Will do.

src/javascript/app/pages/user/metatrader/metatrader.js
```diff
@@ -68,8 +66,12 @@ const MetaTrader = (() => {
     };
 
     const addAllAccounts = () => (
-        new Promise((resolve) => {
+        new Promise((resolve, reject) => {
             BinarySocket.wait('mt5_login_list').then((response) => {
+                if (response.error) {
+                    reject(response.error);
```
negarn: I think we can't proceed if there's an error; user can't really use the dashboard unless we know what their MT5 accounts are. I suggest using the error function of metatrader to display that error message and ask user to try again.

reza-fs: Yeah, that is a great idea, will work on it.

src/javascript/app/pages/user/metatrader/metatrader.js
```diff
@@ -22,7 +22,11 @@ const MetaTrader = (() => {
         BinarySocket.wait('landing_company', 'get_account_status', 'statement').then(async () => {
             if (isEligible()) {
                 if (Client.get('is_virtual')) {
-                    await addAllAccounts();
+                    try {
+                        await addAllAccounts();
```
negarn: how about the second usage of this on line 33?

reza-fs: Done.

negarn: what is done? 😅 I see no change committed

PR #5830, opened 2019-12-04
mamali/show-error-for-pa-withdrawal

reza-fs:


src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -225,6 +225,11 @@ const PaymentAgentWithdraw = (() => {
                 }
                 break;
         }
+        if (response.error) {
```
negarn: seems like we were handling error in `default` case above. should we move that code to an error function so we handle them both the same way? or are they different?

reza-fs: Yeah, I've changed it. Thanks for mentioning that, I completely missed it.

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -216,6 +216,9 @@ const PaymentAgentWithdraw = (() => {
                 break;
             default: // error
                 if (response.echo_req.dry_run === 1) {
+                    const $error = $('#form-error');
+                    $error.text(response.error.message);
+                    $error.removeClass('invisible');
                     setActiveView(view_ids.form);
                     $(field_ids.frm_msg).setVisibility(1).html(response.error.message);
```
negarn: why doesn't this already work?

reza-fs: Yep, there was a missing lookup value, I've changed it to use that. ended up removing these 3 lines.

src/javascript/app/pages/cashier/payment_agent_withdraw.js
```diff
@@ -26,6 +26,7 @@ const PaymentAgentWithdraw = (() => {
         ddl_agents: '#ddlAgents',
         txt_agents: '#txtAgents',
         txt_amount: '#txtAmount',
+        frm_msg   : '#form-error',
```
negarn: there's a `txt_desc` also needing to use `frm_msg` but `txt_desc` also doesn't exist. should we remove that block?

reza-fs: By removing `txt_desc` block, do you mean this block? 
```js
$(field_ids.txt_desc).off().on('keyup', () => {
                if (isVisible(getElementById('withdrawFormMessage'))) {
                    $(field_ids.frm_msg).setVisibility(0);
                }
            });
```
If so, I'm not sure whether or not we want to remove this. from the commit message, I only can know this is necessary. What do you suggest?

reza-fs: Ok, I've checked it. looks like it was a duplicate of what I've done here. I removed it.

negarn:
please resolve the test failing

PR #5833, opened 2019-12-11
Regulatory update

sash-binary:
- Add PDF files
- Change external links to PDF


src/images/pages/regulation/map.svg
```diff
@@ -1,3565 +1 @@
-<?xml version="1.0" encoding="utf-8"?>
```
negarn: wait, if we're deleting the map.svg, how will we display the regulation map image?

sash-binary: map svg should be updated, not deleted

PR #5838, opened 2019-12-13
br_dp2p_port

mahdipourziaei-binary:


src/javascript/app/pages/cashier/dp2p.js
```diff
@@ -0,0 +1,95 @@
...
+        const is_svg = Client.get('landing_company_shortcode') === 'svg';
+        if (is_svg) {
+            import('@deriv/p2p').then((module) => {
+                const el_dp2p_container = document.getElementById('binary-dp2p');
+                const shadowed_el_dp2p = el_dp2p_container.attachShadow({ mode: 'closed' });
+
+                const el_main_css = document.createElement('style');
+                el_main_css.innerHTML = `
+                @import url(${urlForStatic('css/p2p.min.css')});
+                :host {
```
mustofa-binary: can we put it inside scss instead?

src/sass/static/pages.scss
```diff
@@ -1609,3 +1609,65 @@ body #not_authenticated_financial ul.checked > li {
...
+    //font-family: $FONT_STACK;
+    //text-rendering: optimizeLegibility;
+    //-webkit-font-smoothing: antialiased;
+    //-moz-osx-font-smoothing: grayscale;
+    //-webkit-text-size-adjust: 100%;
+    //-ms-text-size-adjust: 100%;
+    //box-sizing: border-box;
+    position: relative;
+    //
+    //h1, h2, h3, h4, h5, h6, p {
```
mustofa-binary: we can remove the comments if its not being used

src/templates/app/cashier/dp2p.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+// eslint-disable-next-line arrow-body-style
+const DP2P = () => {
+
```
mustofa-binary: nit: can remoce line here

src/javascript/app/pages/cashier/dp2p.js
mahdipourziaei-binary: @mustofa-binary We cannot 👍 The styles should be placed inside the shadow DOM which we access through JS, and we need to get the path of the CSS file with `urlForStatic` function.

mustofa-binary: can you make a comment on it then, since its an unusual flow on BS

mahdipourziaei-binary: Sure I'll add it @mustofa-binary

PR #5842, opened 2019-12-17
Mahdi Aryayi / Added alert for svg accout authentication

mahdiaryayi-fs:


src/javascript/app/pages/user/account/settings.js
```diff
@@ -14,6 +15,24 @@ const Settings = (() => {
                 $('#change_password').setVisibility(1);
             }
 
+            // Disabling Authentication button for SVG accounts
+            const is_authenticated                 = State.getResponse('get_account_status.status').includes('authenticated');
+            const is_client_prompt_to_authenticate = State.getResponse('get_account_status.prompt_client_to_authenticate');
+            const is_high_risk                     = Client.getRiskAssessment();
```
negarn: are you sure we need is_high_risk? I think is_client_prompt_to_authenticate will be true if client is high risk, so we can remove this

src/javascript/app/pages/user/account/settings.js
```diff
@@ -14,6 +15,24 @@ const Settings = (() => {
                 $('#change_password').setVisibility(1);
             }
 
+            // Disabling Authentication button for SVG accounts
+            const is_authenticated                 = State.getResponse('get_account_status.status').includes('authenticated');
```
negarn: we already have `status` declared, so we can use that instead of `State.getResponse('get_account_status.status')`

src/javascript/app/pages/user/account/settings.js
mahdiaryayi-fs: This is what I saw in the card. I am not sure about relationship between their logic. If they are relevant, I can remove is_high_risk

negarn: could you remove it and ask them to try?

mahdiaryayi-fs: sure

PR #5859, opened 2019-12-25
mamali/independent-mf-mlt

reza-fs:


src/javascript/app/pages/user/accounts.js
```diff
@@ -86,27 +86,44 @@ const Accounts = (() => {
     const getCompanyCountry = account => Client.getLandingCompanyValue(account, landing_company, 'country');
 
     const populateNewAccounts = (upgrade_info) => {
+        // TODO Populate multiple account opening.
```
negarn, line 89: what is this comment?

reza-fs: Ah :D I forgot to remove it after I was done with the task.

PR #5861, opened 2019-12-30
Mahdi Aryayi / Updated payment method notes

mahdiaryayi-fs:


src/templates/app/cashier/payment_methods.jsx
```diff
@@ -600,8 +600,12 @@ const PaymentMethods = () => {
             </div>
 
             <div className='gr-padding-10'>
-                <p className='hint' data-show='-eucountry'>{it.L('Note : The minimum amount for withdrawal will vary depending on the latest exchange rates.')}</p>
-                <p className='hint'>* {it.L('All your deposits and withdrawals are processed by [_1] within 24 hours. However, there may be additional processing time required by your bank or money transfer service.', it.website_name)}</p>
+                <p className='hint'>Note:</p>
```
negarn: should be `it.L('Note:')`

PR #5873, opened 2020-01-08
amam/personal_details_error

mustofa-binary:
we shouldnt let client to udload poi when place of birth is not completed


src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -896,7 +896,8 @@ const Authenticate = (() => {
                 service      : 'onfido',
             }).then((response) => {
                 if (response.error) {
-                    resolve();
+                    resolve(response.error.code);
+                    return;
```
negarn: why do we need another return? doesn't resolve already return?

src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -928,6 +929,7 @@ const Authenticate = (() => {
         }
         
         const onfido_token = await getOnfidoServiceToken();
+        const is_onfido_pob_error = onfido_token === 'MissingPersonalDetails';
```
negarn: can we rename the variable? it's not POB error, it's a generic personal details error code

src/templates/app/user/authenticate.jsx
```diff
@@ -94,6 +97,17 @@ const Authenticate = () => (
                             <img className='gr-padding-20' src={it.url_for('images/pages/authenticate/valid.svg')} />
                             <h1 className='gr-padding-10'>{it.L('Your proof of identity has been verified successfully')}</h1>
                         </div>
+
+                        <div id='personal_details_error' className='center-text gr-padding-20 invisible'>
+                            <h2 className='gr-padding-10'>{it.L('Update your personal details')}</h2>
+                            <p>{it.L('We can\'t validate your personal details because there is some information missing.\nPlease update your telephone to continue.')}</p>
```
negarn: please don't use `\n` in the message. use two paragraph tags instead.

negarn: and we are not hardcoding `telephone`. you can take that from the API error details field and concatenate them with a comma here. look at how cashier is doing it.

src/javascript/app/pages/user/account/authenticate.js
mustofa-binary: rejecting or resolving won't stop the execution of the rest of the function. The function may contain code that will create confusing results

negarn: I see, I always thought it did :D

src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -36,6 +36,12 @@ const Authenticate = (() => {
         $submit_status_uns,
         $submit_table_uns;
 
+    const personal_details_map = {
+        place_of_birth: localize('Place of birth'),
+        residence     : localize('Country of residence'),
+        phone         : localize('Phone number'),
+    };
```
negarn: this is not the full list of errors. you can find it here https://github.com/binary-com/deriv-app/blob/dev/packages/core/src/Modules/Cashier/Components/Error/error.jsx#L48-L56

perhaps we can refactor it so it can be declared once and used in both places

src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -919,15 +926,30 @@ const Authenticate = (() => {
...
+        
+        if (
+            service_token_response.error &&
+            service_token_response.error.code === 'MissingPersonalDetails'
+        ) {
+            has_personal_details_error = true;
+            let missing_personal_fields = '';
+            Object.keys(service_token_response.required_fields).forEach(field => {
+                missing_personal_fields += !missing_personal_fields ? personal_details_map[field] : `, ${personal_details_map[field]}`;
+            });
```
negarn: @mustofa-binary please have a look at how we do it in the existing files https://github.com/binary-com/deriv-app/blob/dev/packages/core/src/Modules/Cashier/Components/Error/error.jsx#L121

better to stick to one way of doing it

src/templates/app/user/authenticate.jsx
```diff
@@ -1,11 +1,14 @@
...
     TabContainer,
     TabContent,
     TabContentContainer,
-    TabsSubtabs }                                  from '../../_common/components/tabs.jsx';
-import { AuthenticateMessage, UnsupportedMessage } from '../_includes/authenticate_message.jsx';
+    TabsSubtabs }        from '../../_common/components/tabs.jsx';
+import { Button }        from '../../_common/components/elements.jsx';
+import {
+    AuthenticateMessage,
+    UnsupportedMessage } from '../_includes/authenticate_message.jsx';
```
negarn: ../ should be before ../../

PR #5877, opened 2020-01-09
salar / bug/international-targeting

salar-fs:


salar-fs:
I was using "/" in branch name which leaded to troubles for gh-pages deployment.
**DELETED** in favor of #5877

PR #5879, opened 2020-01-12
salar / lower TnC inactivity limits

salar-fs:


src/translations/ach_UG.po
```diff
@@ -2612,7 +2612,7 @@ msgstr "crwdns38045:0crwdne38045:0"
 msgid "Client withdrawals prior to this year amounted to almost USD 500 million. More than USD 140 million worth of client withdrawals is expected this year."
 msgstr "crwdns43616:0crwdne43616:0"
 
-msgid "Client's demo accounts will be deleted after 365 days of inactivity."
+msgid "Client's demo accounts will be deleted after 30 days of inactivity."
```
negarn: @salar-fs please don't update the autogenerated translations files directly. they will be auto updated once we run the update translations script after merging your PR. @reza-fs @mahdiaryayi-fs please assist if there's any confusion

yarn.lock
```diff
@@ -0,0 +1,11303 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
negarn: I don't understand why we need yarn here, also if you are making a change in one PR please don't make the same change in another PR.

.gitignore
```diff
@@ -9,3 +9,4 @@ scripts/CNAME
 .DS_Store
 *.swp
 npm-debug.log
+.nvmrc
```
negarn: same here, if you changed this in one PR, don't submit the same change in another PR

yarn.lock
salar-fs: There is no **need**. It's often optional which package manager one is using and I'm pushing this as it's a best practice to track **lock** files in git.
Please have a look [here](https://stackoverflow.com/questions/44552348/should-i-commit-yarn-lock-and-package-lock-json-files/44904494)

negarn: we already have package-lock.json. please remove yarn as it's not used in this repo.

PR #5889, opened 2020-01-20
amam/mt5_dry_run

mustofa-binary:
use dry run call from BE for mt5 account creation intention of labuan advanced


src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -1012,7 +1010,11 @@ const Authenticate = (() => {
         if (is_required || has_svg_account){
             initTab();
             initAuthentication();
-            if (is_from_mt5) {
+
+            const { identity, document } = authentication_status;
+            const is_not_fully_authenticated = identity.status !== 'verified' && document.status !== 'verified';
+            const is_not_high_risk = !/high/.test(State.getResponse('get_account_status.risk_classification'));
+            if (is_not_fully_authenticated && has_svg_account && is_not_high_risk) {
```
mustofa-binary: this validations are from previous commits.  whether to add the banner or not is still under discussion

src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -222,19 +222,22 @@ const MetaTraderConfig = (() => {
...
+            leverage        : 100,
+            mainPassword    : 'Test1234',
+            mt5_account_type: 'advanced',
+            mt5_new_account : 1,
+            name            : 'test real labuan advanced',
+        };
+        BinarySocket.send(req);
+
+        // update account status authentication info
+        BinarySocket.send({ get_account_status: 1 }, { forced: true });
```
negarn: should we wait for the response of mt5_new_account before sending this?

src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -233,10 +233,12 @@ const MetaTraderConfig = (() => {
             mt5_new_account : 1,
             name            : 'test real labuan advanced',
         };
-        BinarySocket.send(req);
-
-        // update account status authentication info
-        BinarySocket.send({ get_account_status: 1 }, { forced: true });
+        BinarySocket.send(req).then((response) => {
+            if (!response.error) {
```
negarn: but the point is that after it gives us error, we should update account status, right?

mustofa-binary: omg yes youre right 😅

PR #5893, opened 2020-01-21
Mamali/Update Crypto Document

reza-fs:


src/templates/static/legal/regulation.jsx
```diff
@@ -155,7 +155,7 @@ const Regulation = () => {
                         <p>{it.L('These documents provide you with key information about our investment products. This information is required by law to help you understand the nature, risks, costs, potential gains and losses of these products and to help you compare them with other products.')}</p>
                         <div className='gr-row'>
                             <FillBox id='crypto_fillbox' padding='4' center border='border-dark-gray' href={it.url_for(`/download/key_information_document/${lang}/Crypto.pdf`)} target='_blank' text={it.L('Cryptocurrencies')} />
```
negarn: we are using Crypto.pdf here, why is the file removed in this PR?

PR #5901, opened 2020-01-28
QuotaExceededError fallback

reza-fs:


src/javascript/_common/storage.js
```diff
@@ -3,13 +3,67 @@ const getPropertyValue       = require('./utility').getPropertyValue;
...
+/**
+ * Removing all but user details
+ */
+const keepUserAndClean = () => {
+    const client_accounts = localStorage.getItem('client.accounts');
+    const active_loginid  = localStorage.getItem('active_loginid');
+    localStorage.clear();
+    localStorage.setItem('client.accounts', client_accounts);
+    localStorage.setItem('active_loginid', active_loginid);
+    location.reload();
```
negarn: not sure if we should reload the page without telling them it's going to happen. let's collect data first and see why this is happening before we decide on how to handle it

reza-fs: Sure thing. Will remove the reload part.

PR #5904, opened 2020-01-29
amam/p2p improvements binary static

mustofa-binary:
- fix styling on binary static
- fix issues on binary static


src/javascript/app/pages/cashier/dp2p.js
```diff
@@ -117,6 +169,9 @@ const DP2P = (() => {
                 local_currency_config: Client.get('local_currency_config'),
                 residence            : Client.get('residence'),
             },
+            custom_strings: {
+                email_domain: 'binary.com',
```
negarn: didn't you make this change in your other PR?

mustofa-binary: removing it 🙏

PR #5914, opened 2020-02-10
salar / Add download mt5 btn

salar-fs:


src/templates/static/platforms.jsx
```diff
@@ -15,29 +15,38 @@ const Platforms = ({
...
-            <strong>{description}</strong>
-            <p>{text}</p>
-            <div className='gr-row'>
-                <div className='gr-12'>
-                    { url &&
-                        <a className='button-secondary' download={download || undefined} href={url} target={target || undefined} rel={/http/.test(url) ? 'noopener noreferrer' : undefined}><span>{button_text}</span></a>
-                    }
+    buttons,
+}) => {
+    const btnStyle = { display: 'inline-block', marginRight: '5px' };
```
negarn:
```suggestion
    const btn_style = { display: 'inline-block', marginRight: '5px' };
```

negarn: it is only used once. why are we declaring it outside? it's changing the indentation of the whole block

src/templates/static/platforms.jsx
```diff
@@ -15,29 +15,38 @@ const Platforms = ({
...
+                        {buttons && buttons.map(button =>
+                            <a
+                                key={button.text}
+                                className='button-secondary'
+                                style={btnStyle}
+                                download={button.download || undefined}
+                                href={button.url}
+                                target={button.target || undefined}
+                                rel={/http/.test(button.url) ? 'noopener noreferrer' : undefined}
+                            ><span>{button.text}</span>
```
negarn:
```suggestion
                            >
                                <span>{button.text}</span>
```

src/templates/static/platforms.jsx
```diff
@@ -218,26 +228,28 @@ const Platform = () => (
                                 header={it.L('MetaTrader 5')}
                                 description={it.L('Advanced multi-asset trading platform')}
                                 text={it.L('Trade Forex, CFDs, and binary options with a powerful platform recognised as the global standard.')}
-                                url={it.url_for('user/metatrader')}
-                                button_text={it.L('Access MT5 dashboard')}
+                                buttons={[
+                                    { text: it.L('Access MT5 dashboard'), url: it.url_for('user/metatrader') },
+                                    { text: it.L('Download MT5'), url: 'https://www.binary.com/en/metatrader/download.html' },
```
negarn: please don't hardcode this link. use the same `it.url_for` format.

src/templates/static/platforms.jsx
salar-fs: about indentation, any suggestion?

negarn: I mean if we don't make a variable for it, we won't need to add `return (` so the indentations won't change

salar-fs: ah OK :) I thought you're talking about the styles and their effect on the buttons.

PR #5924, opened 2020-02-19
salar / remove visa,mastercard PDF tut links

salar-fs:


src/templates/app/cashier/payment_methods.jsx
```diff
@@ -413,9 +414,8 @@ const PaymentMethods = () => {
                                 ],
                             },
                             {
-                                id      : 'jeton',
-                                dataShow: '-eucountry',
-                                row     : [
+                                id : 'jeton',
```
negarn: there is another file where jeton has the flag -eucountry, can you remove it from there too?

salar-fs: Sure

PR #5962, opened 2020-03-31
negarn/hide_conversion_note_for_usd

negarn:
hide limit varies note if current account is in USD currency


src/javascript/app/pages/cashier/account_transfer.js
```diff
@@ -108,6 +108,11 @@ const AccountTransfer = (() => {
         if (/iom|malta/.test(Client.get('landing_company_shortcode'))) {
             el_transfer_fee.setVisibility(0);
         }
+
+        // transfer limit is calculated by USD, so for USD the amount will not vary based on exchange rates
+        if (client_currency !== 'USD') {
+            document.getElementsByClassName('exchange-rate-vary')[0].setVisibility(1);
```
mustofa-binary: if its unique, can just use id, unless we want to select multiple element

PR #5971, opened 2020-04-02
negarn/fix_buttons_mobile

negarn:
fix button styling in mobile

### DIFF
--- a/src/templates/static/metatrader/download.jsx
+++ b/src/templates/static/metatrader/download.jsx
@@ -198,6 +198,7 @@ const MT5WebPlatform = ({ is_first_child }) => (
                 <span>{it.L('Trade with Real account')}</span>
             </a>
             <a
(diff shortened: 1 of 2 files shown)


src/templates/static/metatrader/download.jsx
```diff
@@ -198,6 +198,7 @@ const MT5WebPlatform = ({ is_first_child }) => (
                 <span>{it.L('Trade with Real account')}</span>
             </a>
             <a
+                id='btn_trade_with_demo'
```
mustofa-binary, line 201: is this intentional?

negarn: @mustofa-binary https://github.com/binary-com/binary-static/pull/5971/files#diff-a34c6a1ccc85212a993f3010dafeb4f3R158

PR #5974, opened 2020-04-02
negarn/rename_mt5_accounts

negarn:
rename mt5 accounts


src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -19,29 +19,29 @@ const MetaTraderConfig = (() => {
             const standard_config = {
                 account_type: 'standard',
                 leverage    : 1000,
-                short_title : localize('Standard'),
+                short_title : localize('Financial'),
```
mustofa-binary, line 22: are we changing this in deriv as well?

negarn: @mustofa-binary yes, the card is in deriv for that

mustofa-binary: deriv.com also need to change then, let me create a card for it as well

negarn: @mustofa-binary thanks, but please add a don't merge label as we want to make sure everyone is onboard and aware before we merge any of them

src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -19,29 +19,29 @@ const MetaTraderConfig = (() => {
...
-                short_title : localize('Synthetic Indices'),
+                short_title : localize('Synthetic'),
             };
 
             return ({
                 gaming: {
-                    demo_volatility: { mt5_account_type: volatility_config.account_type, max_leverage: volatility_config.leverage, title: localize('Demo Synthetic Indices'), short_title: volatility_config.short_title },
-                    real_volatility: { mt5_account_type: volatility_config.account_type, max_leverage: volatility_config.leverage, title: localize('Real Synthetic Indices'), short_title: volatility_config.short_title },
+                    demo_volatility: { mt5_account_type: volatility_config.account_type, max_leverage: volatility_config.leverage, title: localize('Demo Synthetic'), short_title: volatility_config.short_title },
+                    real_volatility: { mt5_account_type: volatility_config.account_type, max_leverage: volatility_config.leverage, title: localize('Real Synthetic'), short_title: volatility_config.short_title },
```
mustofa-binary, line 38: @negarn sorry, `volatility_config`|`standard_config`|`advanced_config`. either in this or next pr, we need to rename all this to not make more confusion 🙏

negarn: @mustofa-binary I think we should keep the variable naming, because it's what everyone knows, and if we name it financial it will cause confusion with MF accounts. so just updating the display names should be enough IMO.

PR #5980, opened 2020-04-06
negarn/show_conversion_non_usd_limit

negarn:
convert non-USD currencies to show USD equivalent


src/templates/app/user/security/limits.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import { Table } from '../../../_common/components/elements.jsx';
+import Loading from '../../../_common/components/loading.jsx';
```
mustofa-binary, line 3: are we still using rules from spacing in binary static 😅

negarn: @mustofa-binary we align `=`, not `from` as far as I remember :)

src/javascript/app/pages/user/account/settings/limits/limits.init.js
```diff
@@ -1,60 +1,94 @@
...
+            }
+
+            const el_withdraw_limit_agg = getElementById('withdrawal-limit-aggregate');
+            const remainder = formatMoney(currency, limits.remainder, 1);
+            const remainder_converted = should_convert ? formatMoney(base_currency, limits.remainder / exchange_rate, 1) : '';
+
+            elementTextContent(el_withdraw_limit_agg,
+                localize('Hence, your withdrawable balance is only up to [_1][_2], subject to your account’s available funds.', [
+                    `${remainder} ${currency}`,
+                    should_convert ? ` (${remainder_converted} ${base_currency})` : '',
```
mustofa-binary, line 83: is it possible to make this into a function ?

PR #5988, opened 2020-04-10
negar/remove_crypto_materials

negarn:
hide inaccurate materials until new ones are ready


src/templates/app/cashier/payment_methods.jsx
```diff
@@ -524,10 +524,10 @@ const PaymentMethods = () => {
                                         { td: it.L('Bitcoin is the world\'s first decentralised cryptocurrency, created in 2009. For more information, please visit [_1].', `${createLink('https://bitcoin.org')}`) },
                                         { td_list: [
                                             { text: 'BTC' },
-                                            { text: '-' },
+                                            { text: '—' },
                                             { text: '' },
                                             { text: <TableValues value={[it.L(`${deposit}${blockchain_confirmations}`, 3), it.L(`${withdrawal}${working_day}`, 1)]} /> },
-                                            { text: <ReferenceLinks pdf_file='Binary.com_Bitcoin.pdf' video_link='https://youtu.be/StIW7CviBTw' /> },
+                                            { text: <ReferenceLinks /* pdf_file='Binary.com_Bitcoin.pdf' video_link='https://youtu.be/StIW7CviBTw' */ /> },
```
mustofa-binary, line 530: since there will be a new material, these files will be deprecated. we can remove the pdf files
```suggestion
                                            { text: <ReferenceLinks /> },
```

negarn: I think we can remove them once we get the new ones, in case some are decided to be accurate and kept

PR #6001, opened 2020-04-29
Arash/ update deriv banner on home page and smart trader

arash-fs:


src/templates/static/home.jsx
```diff
@@ -164,6 +164,18 @@ const Home = () => {
...
         <React.Fragment>
+            <div className='container gr-parent gr-padding-30'>
+                <div id='deriv_banner_container' data-show='-eucountry'>
+                    <a target='_blank' rel='noopener noreferrer' href='https://deriv.com/interim/deriv/?utm_source=binary&utm_medium=referral&utm_campaign=deriv-launch&utm_content=page-trading-banner'>
+                        <img className='deriv_banner_background' src={it.url_for('images/deriv/deriv-banner-bg.png')} />
+                        <img className='deriv_banner_background deriv_banner_background_mobile' src={it.url_for('images/deriv/deriv-banner-bg-mobile.png')} />
+                        <img className='deriv_banner_chevron' src={it.url_for('images/deriv/chevron_right.svg')} />
+                        <h3 className='deriv_banner_standards'>{it.L('[_1] is rebranding to Deriv.com', it.website_name)}</h3>
+                        <img className='deriv_banner_platform' src={it.url_for('images/deriv/deriv-platform.png')} />
+                        <h3 className='deriv_banner_explore'>{it.L('Discover what’s new')}</h3>
```
negarn, line 175: if the two pieces of code for home and trade are exactly the same, please extract them to a new file and import it here instead

arash-fs: Done. would you check to see the new file is located in the right place?

src/templates/_common/components/deriv_banner.jsx
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
+import PropTypes from 'prop-types';
 
-const DerivBanner = () => (
+const DerivBanner = ({ dynamic_link }) => (
```
negarn, line 4: can we use a better name such as content_name etc?

arash-fs: Sure :+1:

PR #6019, opened 2020-05-10
behnaz/Update-dependencies

behnaz-fs:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "binary-static",
-  "version": "3.8.0",
+  "version": "3.8.1",
```
mustofa-binary, line 3: we dont have npm module on npm for binary static, i dont think we need to bump new version on binary static now

behnaz-fs: thanks will change to 3.8.0

PR #6028, opened 2020-05-19
Sam/add p2p tnc

sam-binary:


src/sass/static/p2p.scss
```diff
@@ -0,0 +1,59 @@
+#p2p-tnc ol {
```
negarn, line 1: please don't put css on the element, use classes instead

src/templates/static/legal/p2p/terms_and_conditions.jsx
```diff
@@ -0,0 +1,96 @@
...
+        <p><i>{it.L('Last updated 22/05/2020')}</i></p>
+
+        <h2 data-anchor='p2p-cashier'>{it.L('P2P Cashier services')}</h2>
+        <p>{it.L('This section of our <a href="[_1]">standard terms of use</a> (“the Terms”) applies to your use of P2P Cashier, our platform that provides an alternative way for buying and selling Deriv credits. You need to read this in conjunction with the <a href="[_1]">General terms of use</a> and <a href="[_1]">Funds and transfers</a>.', 'https://www.binary.com/en/terms-and-conditions.html#legal-binary')}</p>
+
+        <ol>
+            <li>{it.L('Ads')}
+                <ol>
+                    <li>{it.L('You are only authorised to post ads if we have pre-approved you according to our <a href="https://www.binary.com/en/terms-and-conditions.html#legal-binary">Know Your Customer policy</a>.')}</li>
+                    <li>{it.L('You can post one or more ads on our platform when you wish to offer to buy or sell Deriv credits. We have the right to restrict the number of your ads or remove your ads for security reasons and reasons of suspected misconduct or fraudulent activity.')}</li>
```
negarn, line 16: please use `website_name` instead of `Deriv` throughout

src/sass/static.scss
```diff
@@ -2,6 +2,7 @@
 
 @import 'static/affiliates_ib';
 @import 'static/metatrader';
+@import 'static/p2p';
```
negarn, line 5: we should rename the file if it's only about tnc p2p

PR #6031, opened 2020-05-19
behnaz/maxDateDateTimeInit

behnaz-fs:


src/javascript/app/common/attach_dom/birth_date_picker.js
```diff
@@ -10,7 +10,7 @@ const generateBirthDate = () => {
         DatePicker.init({
             selector : date_of_birth,
             minDate  : -100 * 365,
-            maxDate  : (-18 * 365) - 4,
+            maxDate  : (-18 * 365) - 5,
```
negarn, line 13: probably this is an exception that happens on leap year? why do we need to - 4 or - 5?

behnaz-fs: yes for leap years

PR #6032, opened 2020-05-20
behnaz/rename-Binary(BVI)-Ltd-Deriv

behnaz-fs:


src/templates/static/about/group_history.jsx
```diff
@@ -67,7 +67,7 @@ const GroupHistory = () => (
 
                     { className: 'year center-text',                                                                                               text: '2014' },
                     { className: 'event right featured', header: it.L('[_1] reaches $2 billion turnover since starting in 1999', it.website_name), text: it.L('[_1] has over 800,000 clients in its database and reaches the US$ 2 billion milestone in our 15th year, making us one of the leading binary options providers.', it.website_name) },
-                    { className: 'event featured',       header: it.L('Regent Markets Group Ltd. renames to Binary Ltd.'),                         text: it.L('Regent Markets Group Ltd. renames to Binary Ltd., harmonising company and website branding throughout the organisation.') },
+                    { className: 'event featured',       header: it.L('Regent Markets Group Ltd. renames to Deriv Ltd.'),                         text: it.L('Regent Markets Group Ltd. renames to Deriv Ltd., harmonising company and website branding throughout the organisation.') },
```
salar-fs, line 70: These are part of the group history so they are facts and do not need to change.

PR #6065, opened 2020-06-08
behnaz/update-licencetoDeriv

behnaz-fs:


negarn:
@behnaz-fs any mention of BVI should be changed to DVI as well, in the file name, link, and text `Binary (BVI)`

behnaz-fs:
no need to change to DVI, as the licence changed from Binary BVI to Deriv BVI, can check the pdf name, so BVI still remain

PR #6090, opened 2020-06-15
negar/update_mt5_names

negarn:
update all MT5 names to the new names


src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -89,8 +89,8 @@ const MetaTraderConfig = (() => {
         };
     })();
 
-    // we need to check if the account type is standard or advanced account before returning landing_company shortcode
-    const getMTFinancialAccountType = acc_type => `${/_advanced$/.test(acc_type) ? 'advanced' : 'standard'}`;
+    // we need to check if the account type is financial or financial_stp account before returning landing_company shortcode
```
ali-mirzaei-fs, line 92: @negar-binary Can we make these regexes backward compatible? `acc_type` should be equal to financial if it matches both `financial` or `standard`.
That's because maybe it takes us some time to move accounts from old groups (advanced & standard) to new ones (financial_stp & financial).

PR #6091, opened 2020-06-15
negar/dynamically_populate_digit_stat_symbols

negarn:
make drop-down population dynamic


src/javascript/app/pages/trade/charts/digit_info.js
```diff
@@ -95,21 +99,23 @@ const DigitInfo = (() => {
...
-        underlyings = underlyings.sort();
-        underlyings.splice(2, 0, '1HZ10V');  // add Volatility 10 (1s)
-        underlyings.splice(4, 0, '1HZ100V'); // add Volatility 100 (1s)
-        let elem    = '';
-        for (let i = 0; i < underlyings.length; i++) {
-            elem += `<option value="${underlyings[i]}">${symbols[underlyings[i]]}</option>`;
-        }
+        const domain  = document.domain.split('.').slice(-2).join('.');
+        const market  = Defaults.get('market');
+        const symbols = ActiveSymbols.getSymbols(State.getResponse('active_symbols'));
```
mustofa-binary, line 104: maybe we can have a helper function for `getSymbolsForMarket`, currently we only have `getSymbolsForSubmarket`

src/javascript/app/pages/trade/markets.jsx
```diff
@@ -89,7 +89,7 @@ const submarket_order = {
     random_nightly : 21,
 };
 
-const submarketSort = (a, b) => {
+export const submarketSort = (a, b) => {
```
mustofa-binary, line 92: we can put this in active symbols, so we have one source for active symbols related function?, and import it here

src/javascript/app/pages/trade/charts/digit_info.js
negarn: @mustofa-binary that's what I thought at first, but `getSymbolsForSubmarket` isn't usable from the outside, it's for populating symbols internally in that function. if you see the line `submarket.symbols = {};` in that function it makes it clear that it's for internal use. so we can keep this outside as it's just a filtering on symbols by market, not something very long or complicated.

src/javascript/app/pages/trade/markets.jsx
negarn: @mustofa-binary I see what you mean, so you want the sorting to be part of the function `getSymbolsForMarket`, okay, so let me create a function for that

PR #6100, opened 2020-06-18
behnaz/Logotrust4

behnaz-fs:


src/templates/static/_layout/head.jsx
```diff
@@ -28,6 +28,7 @@ const Head = () => (
 
         <link rel='manifest' href={`${it.root_url}manifest.json`} />
         <script type='text/javascript' src={`${it.root_url}pushwoosh-web-notifications.js`} async />
+        <script type='text/javascript' src='//widget.trustpilot.com/bootstrap/v5/tp.widget.bootstrap.min.js' async />
```
negarn, line 31: is this the correct link? why is it starting with `//`?

src/templates/static/home.jsx
```diff
@@ -403,6 +403,12 @@ const Home = () => {
                             </div>
                         </div>
                     </div>
+                    <SeparatorLine className='gr-bottom-30' show_mobile />
+                </div>
+            </div>
+            <div >
+                <div className='trustpilot-widget .gr-centered-m' data-locale='en-US' data-template-id='53aa8912dec7e10d38f59f36' data-businessunit-id='58dc92220000ff00059f7dc4' data-style-height='150px' data-style-width='100%' data-theme='light' data-stars='4,5' data-review-languages='en'>
```
negarn, line 410: typo? `.gr-centered-m`

behnaz-fs: checking

behnaz-fs: used to grantee centered margin, removed as being in center handled by other way

src/templates/static/_layout/head.jsx
behnaz-fs: yes, this is a "protocol-relative" link. It uses http or https depending on what was used to load the current page, https://stackoverflow.com/questions/550038/is-it-valid-to-replace-http-with-in-a-script-src-http

negarn: we should enforce https, no?

src/templates/static/home.jsx
negarn: @behnaz-fs that's not what I meant, in beginning of class names we don't use `.` in html, only in css

src/sass/_common/footer.scss
```diff
@@ -108,8 +107,8 @@
         }
     }
     .trustlogo2 {
-        float: right;
-        width: 175px;
+        width: 180px;
+        margin-right: -15px;
```
negarn, line 111: please use center alignment class rather than hardcoded pixel of margin

src/templates/app/_layout/footer.jsx
```diff
@@ -140,6 +140,11 @@ const Footer = () => (
                                     { media: 'reddit',      href: 'https://www.reddit.com/user/Deriv_official/' },
                                 ]}
                             />
+                            <div className='trustlogo2'>
+                                <div className='trustpilot-widget' data-locale='en-US' data-template-id='5419b757fa0340045cd0c938' data-businessunit-id='58dc92220000ff00059f7dc4' data-style-height='25px' data-style-width='100%'>
+                                    <a href='https://www.trustpilot.com/review/binary.com' target='_blank' rel='noopener noreferrer'>Trustpilot</a>
```
negarn, line 145: do we need to send `Trustpilot` to translations?

src/templates/static/_layout/head.jsx
behnaz-fs: added https

src/sass/_common/footer.scss
behnaz-fs: it has an internal margin out of our control, know it is not good idea but cannot find other way with this third-party widget to force be right align with next line in footer.

src/templates/static/home.jsx
behnaz-fs: ohh you are right, dot was extra, no need to this class anymore

src/templates/app/_layout/footer.jsx
behnaz-fs: it is third-party logo with dynamic text so think cannot translate, pls share your advice

negarn: true

src/sass/_common/external/grid.scss
```diff
@@ -61,6 +61,10 @@
     padding-bottom: 30px;
 }
 
+.gr-bottom-30 {
+    padding-bottom: 30px;
```
negarn, line 65: can you try using `gr-parent gr-padding-30` in conjunction? I think it should remove the padding top but keep the padding bottom 30 as you want

src/templates/app/_layout/head.jsx
```diff
@@ -28,6 +28,7 @@ const Head = () => (
 
         <link rel='manifest' href={`${it.root_url}manifest.json`} />
         <script type='text/javascript' src={`${it.root_url}pushwoosh-web-notifications.js`} async />
+        <script type='text/javascript' src='//widget.trustpilot.com/bootstrap/v5/tp.widget.bootstrap.min.js' async />
```
negarn, line 31: https should be here too

PR #6118, opened 2020-06-30
mamali/upgradeable eu to maltainvest

reza-fs:


src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -498,10 +498,8 @@ const PersonalDetails = (() => {
             get_settings_data = State.getResponse('get_settings');
             is_fully_authenticated = checkStatus(account_status , 'authenticated');
 
-            if (!residence) {
+            if (!residence || is_virtual) {
```
negarn, line 501: why?

reza-fs: Thanks for finding this. Sorry I missed your comment. it was there for a test and I forgot to remove it.

PR #6146, opened 2020-07-14
chaks/nginx_error_403

shudarshon-binary:


binary.com.conf
```diff
@@ -5,6 +5,8 @@ server {
     add_header Cache-Control "public, max-age=7200, s-maxage=600, must-revalidate";
     charset UTF-8;
 
+    root /usr/share/nginx/html;
+    index index.html index.htm;
```
mustofa-binary, line 9: is this assuming inside en, there is index.html?, or redirecting it to en/home?

shudarshon-binary: this is not inside `en` instead this is inside `dist` which the root directory for static contents.

PR #6154, opened 2020-07-26
behnaz/accountclosure-dialog

behnaz-fs:


src/images/pages/account_closure/ic-emergency.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="96" height="96"><g fill="none" fill-rule="evenodd"><path fill="#CC2E3D" fill-rule="nonzero" d="M42.187 3.595c2.388-4.794 9.234-4.794 11.622 0L74.005 43.98l21.3 42.612c2.16 4.32-.984 9.408-5.814 9.408H6.511c-4.836 0-7.974-5.088-5.814-9.408l21.3-42.612 20.19-40.386zM48 66a4.286 4.286 0 004.274-3.99l1.68-24.568C54.396 33.492 51.61 30 48 30s-6.396 3.492-5.954 7.442l1.68 24.567A4.286 4.286 0 0048 66zm6 12.07v-.14c0-3.35-2.545-5.93-6.003-5.93C44.551 72 42 74.584 42 77.93v.14c0 3.346 2.55 5.93 5.997 5.93C51.455 84 54 81.41 54 78.07z"/><path d="M0 0h96v96H0z"/></g></svg>
```
salar-fs, line 1: Did you run `svgo` for this?

src/sass/static/pages.scss
```diff
@@ -1612,6 +1612,38 @@ body #not_authenticated_financial ul.checked > li {
...
+            background-color: $COLOR_WHITE;
+        }
+        #ic-emergency {
+            width: 96px;
+            height: 96px;
+            object-fit: contain;
+            margin: 0.5em 0em;
+        }
+        .warning-msg {
+            line-height: 0.7;
```
salar-fs, line 1632: Can we use `.7em` here?

src/templates/app/user/security/account_closure.jsx
```diff
@@ -32,7 +32,29 @@ const ClosureDescription = ({
...
+            <div className='center-text gr-padding-10'>
+                <img
+                    id='ic-emergency'
+                    className='responsive'
+                    src={it.url_for('images/pages/account_closure/ic-emergency.svg')}
+                />
+                <h1 className='gr-padding-10'>{it.L('Warning!')}</h1>
+                <p className='warning-msg'>{it.L('If you deactivate:')}</p>
+                <p className='warning-msg'>{it.L('You’ll be logged out automatically.')}</p>
+                <p className='warning-msg'>{it.L('You will ')}<span id='red-msg'>{it.L('NOT')}</span>{it.L(' be able to log in again.')}</p>
```
salar-fs, line 48: Better to use `placeholders` here and not to split the sentence.
You can use:
`it.L('You will [_1] be able to log in again.', `&lt;span id='red-msg'&gt;${it.L('NOT')}&lt;/span&gt;`)`

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -302,10 +302,16 @@ const AccountClosure = (() => {
             collapsible: true,
             active     : true,
         });
+        $('#back').on('click', () => $('#account-closure-dialog').css('display', 'none'));
```
negarn, line 305: let's store `$('#account-closure-dialog')` in a global variable so we don't have to keep fetching it every time on clicking back, deactivate, and form submission

negarn: can we use `.setVisibility(0)` instead of setting css display to none?

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -302,10 +302,16 @@ const AccountClosure = (() => {
...
+
+        $('#deativate').on('click', () => {
+            $('#account-closure-dialog').css('display', 'none');
+            submitForm();
+        });
 
         $(form_selector).on('submit', (event) => {
             event.preventDefault();
-            submitForm();
+            $('#account-closure-dialog').css('display', 'flex');
```
negarn, line 314: then this could change to `.setVisibility(1)`

src/sass/static/pages.scss
```diff
@@ -1612,6 +1612,38 @@ body #not_authenticated_financial ul.checked > li {
     & .ui-accordion .ui-accordion-icons {
         padding-right: 1.2em;
     }
+    & #account-closure-dialog {
```
negarn, line 1615: why is `&` needed here?

src/sass/static/pages.scss
```diff
@@ -1612,6 +1612,38 @@ body #not_authenticated_financial ul.checked > li {
...
+        display: none;
+        left: 0;
+
+        #account-closure-dialog-content {
+            width: 496px;
```
negarn, line 1620: it's not a good idea to hardcode pixel width and height since responsiveness will be broken (mobile view etc). if you need to hardcode values, it's better to use either rem or percentage values.

negarn: please test this entire dialog in different browser sizes and update any css values to be responsive that need it accordingly

src/templates/app/user/security/account_closure.jsx
negarn: no, this should be `it.L('You will [_1]NOT[_2] be able to log in again.', '<span id="red-msg">', '</span>')` so that translators can see the whole string and structure the grammar however they need to

salar-fs: Right :+1:

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -302,10 +302,18 @@ const AccountClosure = (() => {
             collapsible: true,
             active     : true,
         });
+        const accountclose_dialog = $('#account-closure-dialog');
```
negarn, line 305: please use our jquery naming convention,
```suggestion
        const $account_closure_dialog = $('#account-closure-dialog');
```

PR #6195, opened 2020-09-06
Amir / Fix: enable create account button after selecting a currency

amirghadimi-fs:
Expected behavior:
After we choose currency the Create account button should go back to enable.


package-lock.json
```diff
@@ -6372,14 +6355,12 @@
...
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "yallist": {
           "version": "3.0.3",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
```
negarn, line 6363: why are there changes to package-lock in this PR?

src/javascript/app/pages/user/set_currency.js
```diff
@@ -152,6 +152,10 @@ const SetCurrency = (() => {
             const $clicked_currency = $(this);
             $currency_list.find('> div').removeClass('selected');
             $clicked_currency.addClass('selected');
+            
+            if ($submit) {
+                $submit.removeClass('button-disabled');
```
negarn, line 157: we shouldn't be doing this here, this is on line 275 right before showing the error, so we should figure out why that line isn't working and fix that instead

package-lock.json
amirghadimi-fs: It seams I used a different version of "npm". I'll fix it.

src/javascript/app/pages/user/set_currency.js
```diff
@@ -276,13 +276,27 @@ const SetCurrency = (() => {
...
     };
 
+    /**
+     * Remove error text if $error is defined
+     * Enable confirm button if isBtnEnable is true
+     *
+     * @param {object} $error // error text jquery element
+     * @param {boolean} isBtnEnable // Enable button
+     */
+    const removeError = ($error, isBtnEnable)=>{
```
negarn, line 291: for variables, we separate them with underscores:
```suggestion
    const removeError = ($error, is_btn_enabled) => {
```

negarn (approved):

src/javascript/app/pages/user/set_currency.js
```diff
@@ -276,13 +276,27 @@ const SetCurrency = (() => {
...
     };
 
+    /**
+     * Remove error text if $error is defined
+     * Enable confirm button if is_btn_enabled is true
+     *
+     * @param {object} $error // error text jquery element
+     * @param {boolean} is_btn_enabled // Enable button
+     */
+    const removeError = ($error, is_btn_enabled)=>{
```
negarn, line 291: please add the same spacing that we generally do
```suggestion
    const removeError = ($error, is_btn_enabled) => {
```

negarn (approved):

PR #6206, opened 2020-09-22
Amir / fix: remove multiplier from asset index

amirghadimi-fs:
In "Asset index" page hide multiplier.


salar-fs (dismissed):

src/javascript/app/pages/resources/asset_index/asset_index.js
```diff
@@ -51,7 +51,7 @@ const AssetIndex = (() => {
                     };
                 }
 
-                const asset_cells = asset_item[idx.cells];
+                const asset_cells = asset_item[idx.cells].filter(cell => cell[idx.cell_props.cell_name] !== 'multiplier'); // Multiplier should be hidden from binary static
```
negarn, line 54: we should not add it to asset items in the first place. please look at the `setActiveSymbols` function, we can filter out all the multipliers data before storing it in active_symbols object

negarn (approved):

PR #6215, opened 2020-09-29
Amir / fix: handle selecting unsupported underlying on trading page

amirghadimi-fs:
On trading page when you select an underlying that is not supported for your country and then you reload page, it remains on loading. When you’re logged in and you select an underlying that is not supported by your country of your IP address and then logout page will remains on loading.

My solution:
I remove underlying and market from session storage and url params when you're selecting an unsupported underlying or you are logging out of your account.
I used a function called "remove" which is in this file:
src/javascript/app/pages/trade/duration.js
and purpose of it is removing trader parameters from session storage and url params.


src/javascript/app/base/client.js
```diff
@@ -8,6 +8,7 @@ const removeCookies      = require('../../_common/storage').removeCookies;
 const urlFor             = require('../../_common/url').urlFor;
 const applyToAllElements = require('../../_common/utility').applyToAllElements;
 const getPropertyValue   = require('../../_common/utility').getPropertyValue;
+const Defaults           = require('../pages/trade/defaults');
```
negarn, line 11: should be on line 2 since we sort these alphabetically by pathname

src/javascript/app/base/client.js
```diff
@@ -80,6 +81,8 @@ const Client = (() => {
         // clear elev.io session storage
         sessionStorage.removeItem('_elevaddon-6app');
         sessionStorage.removeItem('_elevaddon-6create');
+        // clear trading session
+        Defaults.remove('underlying','market');
```
negarn, line 85:
```suggestion
        Defaults.remove('underlying', 'market');
```

src/javascript/app/pages/trade/process.js
```diff
@@ -109,9 +109,10 @@ const Process = (() => {
             getElementById('contract_confirmation_container').style.display      = 'block';
             getElementById('contracts_list').style.display = 'none';
             getElementById('confirmation_message').hide();
-
+
```
negarn, line 112:
```suggestion

```

src/javascript/app/pages/trade/process.js
```diff
@@ -109,9 +109,10 @@ const Process = (() => {
             getElementById('contract_confirmation_container').style.display      = 'block';
             getElementById('contracts_list').style.display = 'none';
             getElementById('confirmation_message').hide();
-
+            
             const confirmation_error = getElementById('confirmation_error');
-            confirmation_error.show();
+            confirmation_error.classList.remove('invisible');
```
negarn, line 114:
```suggestion
            confirmation_error.setVisibility(1);
```

negarn (approved):

PR #6216, opened 2020-09-29
zaki / handle_poi_poa_be_changes

zaki-binary:
**Changelog**
- refactor: refactor authentication for POI and POA to comply with `account_status` API updates


mustofa-binary (commented):
do we still relying on allow_document_upload flag to show the page as we do now, or they have changed it to only rely on needs_verification flag of each status. if no change, then its all good 👍

zaki-binary:
> do we still relying on allow_document_upload flag to show the page as we do now, or they have changed it to only rely on needs_verification flag of each status. if no change, then its all good 👍

It seems like apart from `further_resubmission_allowed`, everything else works (as per my own testing with the API changes) and I looked for mentions of other deprecations in the API changes and its only the `further_resubmissions_allowed` that's removed as we'll use `needs_verification` instead.

mustofa-binary (dismissed):

negarn (approved):

PR #6223, opened 2020-10-01
Amir / Reword responsible trading info on GAMSTOP

amirghadimi-fs:


src/templates/static/responsible_trading.jsx
```diff
@@ -41,7 +41,7 @@ const ResponsibleTrading = () => (
 
             <div data-show='eucountry'>
```
negarn, line 42: we should also change this to iom instead of eucountry

negarn: for gamstop only

negarn (approved):

PR #6239, opened 2020-10-12
Akmal / Added min/max fields validation for self-exclusion form based on account limits

akmal-binary:


src/javascript/app/pages/user/account/settings/self_exclusion.js
```diff
@@ -165,8 +170,7 @@ const SelfExclusion = (() => {
             if (!/session_duration_limit|max_open_bets/.test(id)) {
                 options.type     = 'float';
                 options.decimals = decimal_places;
-            }
-            if (/max_open_bets/.test(id)) {
+            } else if (/max_open_bets/.test(id)){
```
negarn, line 173: if we move this if first, we don't need to check for max open bets twice

akmal-binary: Alright, changed

negarn (approved):

PR #6240, opened 2020-10-12
Akmal / Fixed new cryptocurrencies issues

akmal-binary:


src/javascript/app/base/client.js
```diff
@@ -13,7 +13,7 @@ const getPropertyValue   = require('../../_common/utility').getPropertyValue;
 const Client = (() => {
     const processNewAccount = (options) => {
         if (ClientBase.setNewAccount(options)) {
-            window.location.href = options.redirect_url || defaultRedirectUrl(); // need to redirect not using pjax
+            setTimeout(() => {window.location.replace(options.redirect_url || defaultRedirectUrl());},250); // need to redirect not using pjax
```
salar-fs, line 16: Why do we need a timeout here?

src/javascript/app/pages/user/set_currency.js
```diff
@@ -109,11 +109,15 @@ const SetCurrency = (() => {
                     text: Currency.getCurrencyName(c) || c,
                     ...(/^UST$/.test(c) && {
                         'data-balloon'       : localize('Tether Omni (USDT) is a version of Tether that\'s pegged to USD and is built on the Bitcoin blockchain.'),
-                        'data-balloon-length': 'large',
+                        'data-balloon-length': 'medium',
+                        'data-balloon-pos'   : 'top',
+                        'class'              : 'show_mobile',
```
salar-fs, line 114: I think we use `-` as a separator for multi-word class names in binary-static.

src/javascript/app/pages/user/set_currency.js
```diff
@@ -109,11 +109,15 @@ const SetCurrency = (() => {
...
                     }),
                     ...(/^eUSDT/.test(c) && {
                         'data-balloon'       : localize('Tether ERC20 (eUSDT) is a version of Tether that\'s pegged to USD and is hosted on the Ethereum platform.'),
-                        'data-balloon-length': 'large',
+                        'data-balloon-length': 'medium',
+                        'data-balloon-pos'   : 'top',
+                        'class'              : 'show_mobile',
```
salar-fs, line 120: Same here.

src/javascript/app/base/client.js
akmal-binary: Because without a timeout, in some cases on IOS and safari, a new page is opened before the location is replaced, so it opens the previous page instead of a successful message. That's the solution proposed on this thread: https://stackoverflow.com/questions/31223216/why-isnt-window-location-href-not-forwarding-to-page-using-safari

src/javascript/app/pages/user/set_currency.js
akmal-binary: I utilized the existing class created 3 years ago actually, do I change it all over the code? It has 26 occurrences  across 13 files.

salar-fs: Ah no need to do so.

negarn (dismissed):

negarn (approved):

src/javascript/app/base/client.js
```diff
@@ -13,7 +13,7 @@ const getPropertyValue   = require('../../_common/utility').getPropertyValue;
 const Client = (() => {
     const processNewAccount = (options) => {
         if (ClientBase.setNewAccount(options)) {
-            window.location.href = options.redirect_url || defaultRedirectUrl(); // need to redirect not using pjax
+            setTimeout(() => {window.location.replace(options.redirect_url || defaultRedirectUrl());},500); // need to redirect not using pjax
```
negarn, line 16:
```suggestion
            setTimeout(() => { window.location.replace(options.redirect_url || defaultRedirectUrl()); }, 500); // need to redirect not using pjax
```

src/sass/_common/common.scss
```diff
@@ -728,14 +728,11 @@ label + input {
 }
 
 @media (max-width: 768px) {
-    .show_mobile {
```
negarn, line 731: we are using this class in other files, if you want to update it we should update it everywhere

negarn (approved):

PR #6241, opened 2020-10-12
Ali / fix: table header size

ali-fs:


src/javascript/app/common/attach_dom/table.js
```diff
@@ -71,13 +71,13 @@ const Table = (() => {
         }
 
         const is_data = (opt === 'data');
-
+
```
salar-fs, line 74: Please remove the extra space here.

ali-fs: White spaces removed

salar-fs (approved):

PR #6247, opened 2020-10-14
Amir / fix:  Client must be able to choose currency after signup

amirghadimi-fs:


salar-fs (approved):

src/javascript/app/pages/user/accounts.js
```diff
@@ -38,7 +38,8 @@ const Accounts = (() => {
             // ask client to set residence first since cannot wait landing_company otherwise
             BinaryPjax.load(urlFor('user/settings/detailsws'));
         }
-        SetCurrency.cleanupPopup();
+        clearPopup();
+        window.addEventListener('beforeunload', clearPopup);
```
negarn, line 42: why do we need before unload as well as on unload?

amirghadimi-fs: When user changes url directly in browser url bar, we face the error on this card. I figured out onUnload function doesn’t work in this situation so I used  "beforeunload" event. Also "beforeunload" doesn’t work when user clicks back button of the browser (I've tested on Chrome).
So I need to use both of them to cover the error.

negarn: @amirghadimi-fs then I suggest we do this in the common page loader script to always set the onuload script (if any) of the current page to the beforeunload event, because if it happens here it can happen on other pages too

src/javascript/app/base/binary_loader.js
```diff
@@ -44,6 +44,7 @@ const BinaryLoader = (() => {
 
         container = getElementById('content-holder');
         container.addEventListener('binarypjax:before', beforeContentChange);
+        window.addEventListener('beforeunload', beforeContentChange);
```
negarn, line 47: we don't need to remove the event later?

amirghadimi-fs: I've didn’t find anywhere suitable for this but the 'beforeunload' itself. Do we need to remove it there?

negarn: as long as loading different pages by clicking on menu items doesn't duplicate this event then it's fine, can you test it?

amirghadimi-fs: Yes sure, I'll test.
Somehow I’m sure that it can’t be duplicated, because it’s on init function of loader. I'll inform you when I tested.

amirghadimi-fs: Done, no duplication by clicking on menu items.

negarn (approved):

PR #6255, opened 2020-10-21
Akmal / Updated self exclusion form logic to allow passing null values to backend to clear the limits

akmal-binary:


src/javascript/app/pages/user/account/settings/self_exclusion.js
```diff
@@ -180,7 +178,7 @@ const SelfExclusion = (() => {
             validations.push({
                 selector        : `#${id}`,
                 validations     : checks,
-                exclude_if_empty: 1,
+                exclude_if_empty: 0,
```
negarn, line 181: can we just remove the `exclude_if_empty` param instead of setting it to false? I believe by default it's false

src/javascript/app/pages/user/account/settings/self_exclusion.js
```diff
@@ -289,14 +287,19 @@ const SelfExclusion = (() => {
     const additionalCheck = data => (
         new Promise((resolve) => {
             const is_changed = Object.keys(data).some(key => ( // using != in next line since response types is inconsistent
-                key !== 'set_self_exclusion' && (!(key in self_exclusion_data) || self_exclusion_data[key] != data[key]) // eslint-disable-line eqeqeq
+                key !== 'set_self_exclusion' && (self_exclusion_data[key] != data[key] && data[key] !== '') || (self_exclusion_data[key] !== undefined && data[key] === '')// eslint-disable-line eqeqeq
```
negarn, line 290: why not do typeof undefined so we don't hit the eslint error?

src/javascript/app/pages/user/account/settings/self_exclusion.js
```diff
@@ -289,14 +287,19 @@ const SelfExclusion = (() => {
...
             ));
 
             if (!is_changed) {
                 showFormMessage(localize('You did not change anything.'), false);
                 resolve(false);
             }
 
+            // use for in loop instead of Object.entries to avoid unnecessary convertion of the object into an array, which later needs to be stored, processed and converted back into an object
```
negarn, line 298: conversion? :D

negarn: please break the comment into multiline for readability

src/javascript/app/pages/user/account/settings/self_exclusion.js
```diff
@@ -289,14 +287,19 @@ const SelfExclusion = (() => {
...
+            for (let key in data) {// eslint-disable-line
```
negarn, line 299: please disable the specific rule if possible

src/javascript/app/pages/user/account/settings/self_exclusion.js
```diff
@@ -289,14 +287,19 @@ const SelfExclusion = (() => {
...
+                data[key] = data[key] === '' ? null : data[key];// eslint-disable-line
```
negarn, line 300: same here

src/javascript/app/pages/user/account/settings/self_exclusion.js
akmal-binary: Ok, it's removed

src/javascript/app/pages/user/account/settings/self_exclusion.js
akmal-binary: I changed as suggested, however, it wasn't the reason causing eslint error. This happens because of `!=` operator as mentioned.

negarn (approved):

src/javascript/app/pages/user/account/settings/self_exclusion.js
```diff
@@ -338,6 +338,7 @@ const SelfExclusion = (() => {
             }
             return;
         }
+        getData();
```
negarn, line 341: we already have the response of self exclusion in this function, can we pass it in instead of requesting it again?

akmal-binary: Alright, I changed the logic to reuse API response instead of making an additional call

negarn (dismissed):

negarn (approved):

PR #6268, opened 2020-10-28
Matin/Fix: Block Rwanda create account

matin-binary:
* If landing company has legal_allowed_currencies then user is able to create account


src/javascript/_common/base/__tests__/client_base.js
```diff
@@ -171,7 +171,7 @@ describe('ClientBase', () => {
             expect(upgrade_info.can_upgrade).to.eq(false);
             expect(upgrade_info.can_upgrade_to.length).to.eq(0);
             expect(upgrade_info.type).to.eq(undefined);
-            expect(upgrade_info.can_open_multi).to.eq(true);
+            expect(upgrade_info.can_open_multi).to.eq(undefined);
```
negarn, line 174: why?

src/javascript/_common/base/client_base.js
```diff
@@ -262,13 +262,15 @@ const ClientBase = (() => {
...
+        const landing_company_obj = State.getResponse('landing_company');
 
         let can_open_multi = false;
         let can_upgrade_to = [];
         let type;
         if ((upgradeable_landing_companies || []).length) {
             const current_landing_company = get('landing_company_shortcode');
-            can_open_multi = upgradeable_landing_companies.indexOf(current_landing_company) !== -1;
+            can_open_multi = upgradeable_landing_companies.indexOf(current_landing_company) !== -1 &&
+                             (landing_company_obj && landing_company_obj.legal_allowed_currencies);
```
negarn, line 273: can you add a comment here about why this check works?

matin-binary: Sure

src/javascript/_common/base/__tests__/client_base.js
matin-binary: This is to bypass the test

src/javascript/_common/base/__tests__/client_base.js
```diff
@@ -166,13 +168,24 @@ describe('ClientBase', () => {
...
             Client.set('landing_company_shortcode', 'svg');
             const upgrade_info = Client.getBasicUpgradeInfo();
             expect(upgrade_info.can_upgrade).to.eq(false);
             expect(upgrade_info.can_upgrade_to.length).to.eq(0);
             expect(upgrade_info.type).to.eq(undefined);
             expect(upgrade_info.can_open_multi).to.eq(true);
         });
+        it('returns as expected for multi account opening without legal currencies', () => {
+            State.set(['response', 'authorize', 'authorize', 'upgradeable_landing_companies'], [ 'svg' ]);
+            State.set(['response', 'landing_company', 'landing_company', 'legal_allowed_currencies'], false);
```
negarn, line 181: I think we should set a real value for it, to mimic what API will send.

matin-binary: In the function itself, it is checking if it exists or not, therefore a boolean value, I have tried using 'USD' as we discussed but the test expects boolean

negarn: @matin-binary then it won't work in real life because API sends a value, right? that means the function is not written correctly

negarn: the test and real life should match, and the function should work the same way.

negarn (dismissed):

negarn (approved):

PR #6278, opened 2020-11-11
Ali / Remove unavailable underlings

ali-fs:


src/javascript/app/common/active_symbols.js
```diff
@@ -171,6 +172,23 @@ const ActiveSymbols = (() => {
         return trade_markets_list;
     };
 
+    const getAvailableUnderlyings = (markets_list) => {
+        const unavailable_underlyings = ['BOOM500', 'BOOM1000', 'CRASH500', 'CRASH1000', 'stpRNG'];
```
negarn, line 176: can you add a comment that these are only offered on deriv?

src/javascript/app/common/active_symbols.js
```diff
@@ -170,6 +171,24 @@ const ActiveSymbols = (() => {
         extend(trade_markets_list, getSubmarkets(active_symbols));
         return trade_markets_list;
     };
+    
+    // The unavailable underlyings are only offered on deriv.com.
+    const getAvailableUnderlyings = (markets_list) => {
+        const unavailable_underlyings = ['BOOM500', 'BOOM1000', 'CRASH500', 'CRASH1000', 'stpRNG'];
+        const cloned = clone(markets_list);
```
salar-fs, line 178:
```suggestion
        const markets_list_clone = clone(markets_list);
```

negarn: I think we can have this variable defined outside of this function so we don't declare it every time, since the value is static

src/javascript/app/common/active_symbols.js
```diff
@@ -170,6 +171,24 @@ const ActiveSymbols = (() => {
         extend(trade_markets_list, getSubmarkets(active_symbols));
         return trade_markets_list;
     };
+    
+    // The unavailable underlyings are only offered on deriv.com.
+    const getAvailableUnderlyings = (markets_list) => {
+        const unavailable_underlyings = ['BOOM500', 'BOOM1000', 'CRASH500', 'CRASH1000', 'stpRNG'];
```
negarn, line 177: I think we can have this variable defined outside of this function so we don't declare it every time, since the value is static

negarn (approved):

PR #6279, opened 2020-11-12
Amir / Fix: Don't show set currency error on creating multi account

amirghadimi-fs:


src/javascript/app/pages/user/set_currency.js
```diff
@@ -206,7 +206,9 @@ const SetCurrency = (() => {
                     }
                 } else {
                     const previous_currency = Client.get('currency');
-                    Client.set('currency', selected_currency);
+                    // If it is creation of a new account, pick it's id up instead of the current loginid
+                    const new_account_loginid = popup_action === 'multi_account' ? response_c.new_account_real.client_id : undefined;
+                    Client.set('currency', selected_currency, new_account_loginid);
```
salar-fs, line 211: Is sending `undefined` safe? Could you have a look at `Client.set` inner working and make sure this wouldn't cause side effects?

amirghadimi-fs: Yes I it’s safe. As you can see in link below we can ignore an argument by set it’s value to `undefined`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

`Client.set` has a default value for third arrangement:
`const set = (key, value, loginid = current_loginid) => {`
So I must pass `undefined` to have it’s default value in some cases.

negarn:
LGTM, will wait for the application of this https://github.com/binary-com/binary-static/pull/6279#discussion_r522082076

salar-fs (approved):

negarn (approved):

PR #6295, opened 2020-11-26
amam/open_livechat_url

mustofa-binary:
- add handler to open livechat via URL


src/javascript/app/base/page.js
```diff
@@ -80,14 +80,24 @@ const Page = (() => {
...
                     // TODO: uncomment below to enable interview popup dialog
                     // InterviewPopup.onLoad();
                 }
-                if (window.location.href.indexOf('?data-elevio-article=') > 0) {
+                if (url_query_strings.get('data-elevio-article')) {
                     Elevio.injectElevio();
                 }
+
+                // Handle opening livechat via URL
+                const is_livechat_open = url_query_strings.get('is_livechat_open');
```
negarn, line 94: I would suggest turning this to a boolean instead of string comparison

negarn (approved):

PR #6305, opened 2020-12-07
suthesh/livechat-fallback

suthesan-deriv:
added fallback for livechat (in the event elevio is not available)


src/javascript/_common/base/livechat.js
```diff
@@ -1,10 +1,27 @@
...
+
 const LiveChat = (() => {
+    const elevio_account_id = '5bbc2de0b7365';
+
+    const livechat_id = 'gtm-deriv-livechat';
+    let livechat_shell;
     const initial_session_variables = { loginid: '', landing_company_shortcode: '', currency: '', residence: '', email: '' };
     const init = () => {
         if (window.LiveChatWidget) {
+            const httpresponse = httpGet(`https://cdn.elev.io/sdk/bootloader/v4/elevio-bootloader.js?cid=${elevio_account_id}`);
```
negarn, line 19: are we doing this on each pjax load as well? is it needed?

src/javascript/app/base/page.js
```diff
@@ -72,6 +72,31 @@ const Page = (() => {
         });
     };
 
+    // Fallback LiveChat icon when Elevio is not available
+    const livechatFallbackCheck = () => {
```
negarn, line 76: can we put this function in the livechat file itself?

negarn (approved):

PR #6309, opened 2020-12-09
Suthesh/livechat-routing

suthesan-deriv:
Livechat group rerouting on each page load (livechat reloads on each page load if there is no ongoing chat)


src/javascript/app/base/binary_loader.js
```diff
@@ -61,9 +61,23 @@ const BinaryLoader = (() => {
             }
             active_script = null;
         }
+        if (window.LiveChatWidget){
```
mustofa-binary, line 64: any reason why do we put it on binary loader instead of page load?

src/javascript/app/base/binary_loader.js
```diff
@@ -61,9 +61,23 @@ const BinaryLoader = (() => {
...
+            window.LiveChatWidget.on('ready', () => {
+                if (window.LiveChatWidget.get('customer_data').status !== 'chatting') {
+                    window.LiveChatWidget.call('destroy');
```
mustofa-binary, line 67: can we reinitialize chat after destroy, is it intentional to separate them?

src/javascript/app/base/binary_loader.js
```diff
@@ -78,8 +92,13 @@ const BinaryLoader = (() => {
         ContentVisibility.init().then(() => {
             BinarySocket.wait('authorize', 'website_status', 'landing_company').then(() => {
                 GTM.pushDataLayer({ event: 'page_load' }); // we need website_status.clients_country
-                LiveChat.init();
-
+                if (!window.LiveChatWidget){
```
mustofa-binary, line 95: i think we can initialize after destroy, and add check in the livechat file itself for this rather than here for better specificity

mustofa-binary (changes requested):

mustofa-binary (approved):

PR #6317, opened 2020-12-21
amam/change_poi_message

mustofa-binary:
- change upload complete status from wait 1-3 days to chat with us
- open livechat window on click chat with us


src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -57,6 +57,10 @@ const Authenticate = (() => {
 
         $('#not_authenticated .file-picker').on('change', onFileSelected);
 
+        $('#chat_with_us').on('click', () => {
```
negarn, line 60: should we off click before on click to handle pjax load not piling on the listener?

negarn (approved):

PR #6319, opened 2020-12-22
Akmal / Fixed redirect from risk disclaimer when changing language

akmal-binary:


src/javascript/app/pages/user/new_account/financial_acc_opening.js
```diff
@@ -21,6 +21,22 @@ const FinancialAccOpening = (() => {
             return;
         }
 
+        if (sessionStorage.getItem('is_risk_disclaimer')){
+            handleResponse(JSON.parse(sessionStorage.getItem('client_form_response')));
+        }
+
+        const setClientFormResponse = () => {
+            const client_form_response = sessionStorage.getItem('client_form_response') ? JSON.parse(sessionStorage.getItem('client_form_response')).echo_req : false;
```
negarn, line 29:
```suggestion
            const client_form_response = sessionStorage.getItem('client_form_response') ? JSON.parse(sessionStorage.getItem('client_form_response')).echo_req : {};
```

src/javascript/app/pages/user/new_account/financial_acc_opening.js
```diff
@@ -21,6 +21,22 @@ const FinancialAccOpening = (() => {
...
+            if (!isEmptyObject(client_form_response)) {
+                const keys = Object.keys(client_form_response);
```
negarn, line 31: do we need this declaration? it's only used once

src/javascript/app/pages/user/new_account/financial_acc_opening.js
```diff
@@ -127,6 +149,7 @@ const FinancialAccOpening = (() => {
                 fnc_response_handler: handleResponse,
             });
         } else {
+            sessionStorage.removeItem('is_risk_disclaimer');
```
negarn, line 152: let's also remove it onunload of this page?

src/templates/app/new_account/financial.jsx
```diff
@@ -79,7 +79,7 @@ const Financial = () => (
                     <p>{it.L('By clicking <strong>Accept</strong> below and proceeding with the Account Opening you should note that you may be exposing yourself to risks (which may be significant, including the risk of loss of the entire sum invested) that you may not have the knowledge and experience to properly assess or mitigate.')}</p>
                     <p className='center-text'>
                         <button className='button' type='submit'>{it.L('Accept')}</button>
-                        <a className='button' href={it.url_for('trading')}><span>{it.L('Decline')}</span></a>
+                        <a id='financial-risk-decline' className='button' href={it.url_for('trading')}><span>{it.L('Decline')}</span></a>
```
negarn, line 82: in binary for id we use underscore separators

src/javascript/app/pages/user/new_account/financial_acc_opening.js
akmal-binary: True, I previously called it twice, therefore created a function, but we don't need it anymore

negarn (approved):

src/javascript/app/pages/user/new_account/financial_acc_opening.js
```diff
@@ -55,6 +64,16 @@ const FinancialAccOpening = (() => {
         });
 
         Promise.all([req_settings, req_financial_assessment]).then(() => {
+            const client_form_response = sessionStorage.getItem('client_form_response') ? JSON.parse(sessionStorage.getItem('client_form_response')).echo_req : {};
```
negarn, line 67: let's store this `sessionStorage.getItem('client_form_response')` instead of fetching it twice

src/templates/app/new_account/financial.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import FinancialForm from '../_includes/financial_form.jsx';
+import Loading from '../../_common/components/loading.jsx';
```
negarn, line 3: should be on line 28 since we sort alphabetically by path in binary

src/javascript/app/pages/user/new_account/financial_acc_opening.js
akmal-binary: Thanks, this is indeed a better way 😃

src/javascript/app/pages/user/new_account/financial_acc_opening.js
```diff
@@ -11,16 +11,26 @@ const State          = require('../../../../_common/storage').State;
 
 const FinancialAccOpening = (() => {
     const form_id = '#financial-form';
+    const client_details = sessionStorage.getItem('client_form_response');
```
negarn, line 14: I think this needs to be in the onload function since it needs to happen when this page loads, right?

akmal-binary: Actually, it would be the same since we just need to get this details once before it's required, so I moved it onLoad to encapsulate the variable scope.

negarn (approved):

PR #6323, opened 2020-12-28
suthesh/refactor livechat

suthesan-deriv:
Refactor livechat to have almost all livechat related codes in livechat.js file and remove console error while travelling between pages quickly


mustofa-binary (approved):

src/javascript/_common/base/livechat.js
```diff
@@ -66,9 +79,75 @@ const LiveChat = (() => {
...
+    const livechatDeletion = () => new Promise ((resolve) => {
+        if (window.LiveChatWidget){
+            window.LiveChatWidget.on('ready', () => {
+                try {
+                    if (window.LiveChatWidget.get('customer_data').status !== 'chatting') {
+                        window.LiveChatWidget.call('destroy');
+                        resolve();
+                    }
+                } catch (e) {
+                    resolve();
```
mustofa-binary, line 119: reject(e)?

suthesan-deriv: I don't think I'll have to reject it, just resolve will do

suthesan-deriv: We need the resolve just to make sure no errors are logged on the console. These errors pop up when user opens certain links that causes the url to change more than once or when user travels between pages too quickly (on either occasions we can ignore the error since the rerouting function will be called again)

PR #6350, opened 2021-01-15
amam/onfido_reach_limit

mustofa-binary:
- add exclusion message for onfido reach submission limit


src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -972,6 +973,12 @@ const Authenticate = (() => {
 
         if (has_personal_details_error) {
             $('#personal_details_error').setVisibility(1);
+        } else if (has_submission_attempts) {
+            $('#limited_livechat').off('click').on('click', (e) => {
+                e.preventDefault();
+                window.LC_API.LC_API.open_chat_window();
+            });
+            $('#limited_poi').setVisibility(1);
```
negarn, line 981: we probably need to set it to invisible again if client loads the page with pjax next time and has_submission_attempts is false, right? (html loads with the last state, i.e. once you set `limited_poi` to visible until you refresh the page, if you land on it by moving around the website links it will not be set to invisible again)

mustofa-binary: actually there isn't any direct link going to authenticate page from pjax, so this case might not happen, however, i add cleanElementVIsibility function to ensure it will never happen as suggested

negarn (dismissed):

negarn (approved):

PR #6361, opened 2021-01-25
Abdalla/Update: Payment processing period

AbdallaMourad:
This PR is to update payment processing time
- Change processing time from `working days` to `days`


src/templates/app/cashier/payment_methods.jsx
```diff
@@ -91,9 +91,9 @@ const PaymentMethods = () => {
 
     const deposit                  = 'Deposit: ';
     const withdrawal               = 'Withdrawal: ';
-    const working_day              = '[_1] working day';
+    const period                = '[_1] day';
```
akmal-binary, line 94: Could you keep the equal sign alignment? 😆 It is spaced out on purpose, so that all equal signs are on the same column.

src/templates/app/cashier/payment_methods.jsx
```diff
@@ -91,9 +91,9 @@ const PaymentMethods = () => {
...
     const instant                  = 'Instant';
-    const working_day_range        = '[_1] to [_2] working days';
+    const period_range        = '[_1] to [_2] days';
```
akmal-binary, line 96: Same here

src/templates/app/cashier/payment_methods.jsx
AbdallaMourad: I thought there is smth to do it on precommit 😄

akmal-binary (approved):

mustofa-binary (approved):

PR #6364, opened 2021-01-26
Amir / update:  Add document upload for Nigerian clients

amirghadimi-fs:


src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -103,6 +115,20 @@ const Authenticate = (() => {
...
         }
+
+        if (is_nigeria) {
+            const $nigeria_file_selector = document.querySelector('[data-type="nigeria"]');
+            document.querySelector('[data-type="poi"]').setVisibility(0);
+            $show_onfido_btns = $nigeria_file_selector.querySelectorAll('h3[data-show-onfido]');
+            $show_onfido_btns.forEach($el => {
+                const data_show_onfido = $el.getAttribute('data-show-onfido');
+                if (data_show_onfido) {
+                    $el.addEventListener('click', showOnfido);
```
mustofa-binary, line 126: why do we want to intialize onfido at this point?, afaik, nigeria client will fallback to unsupported poi right?

src/templates/app/_includes/authenticate_message.jsx
```diff
@@ -45,7 +45,7 @@ const FileSelector = ({
                                 const j = i + 1;
                                 return (
                                     <React.Fragment key={i}>
-                                        <h3>{document.name}</h3>
+                                        <h3 data-show-onfido={(type === 'nigeria' && !document.is_nimc) ? document.value : ''}>{document.name}</h3>
```
mustofa-binary, line 48: we can combine the logic from authenticate.js instead of adding ternary in template

mustofa-binary (changes requested):

src/templates/app/_includes/authenticate_message.jsx
amirghadimi-fs: We have a selfie section for Nigeria, and there are many changes for poi part in the DOM as well. So IMO if we want to handle this completely from `authenticate.js`, we need to inject or change DOM using JS. This will add complexity to the code.

mustofa-binary (approved):

src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -727,11 +727,13 @@ const Authenticate = (() => {
...
     // Validate user input
     const validate = (file) => {
+        const is_nigeria = country_code === 'NGA';
         const required_docs = ['passport', 'national_identity_card', 'driving_licence'];
         const doc_name = {
             passport              : localize('Passport'),
-            national_identity_card: localize('Identity card'),
+            national_identity_card: is_nigeria ? localize('NIMC slip') : localize('Identity card'),
             driving_licence       : localize('Driving licence'),
+            other                 : is_nigeria ? localize('Age declaration document') : '',
```
negarn, line 736: nitpick: `...(is_nigeria && { other: localize('Age declaration document') }),`

amirghadimi-fs:
I'll continue this on here:
https://github.com/binary-com/deriv-app/pull/2762

matin-binary (approved):

src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -339,31 +374,54 @@ const Authenticate = (() => {
         processFiles(files);
     };
 
+    const showNimcDialog = () =>  new Promise(resolve => {
```
reza-fs, line 377:
```suggestion
    const showNIMCDialog = () =>  new Promise(resolve => {
```

reza-fs: No need to return the wrapped promise. Since you can just return `Dialog.confirm` function call.

src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -339,31 +374,54 @@ const Authenticate = (() => {
...
             return;
         }
+
+        if (is_nigeria && !has_accepted_nimc_notice){
+            await showNimcDialog().then(result => has_accepted_nimc_notice = result);
+
+            if (!has_accepted_nimc_notice){
+                return;
+            }
+        }
```
reza-fs, line 403:
```suggestion
        if (is_nigeria && !has_accepted_nimc_notice && !await showNIMCDialog()){
            return;
        }
```

amirghadimi-fs: We have to set `has_accepted_nimc_notice` inside the condition, because before we show the dialog we should check if user agreed it before or not.

src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -339,31 +374,52 @@ const Authenticate = (() => {
...
-    const submitFilesUns = ($files) => {
+    const submitFilesUns = async ($files) => {
+        const is_nigeria = country_code === 'NGA';
+
         if ($button_uns.length && $button_uns.find('.barspinner').length) { // it's still in submit process
             return;
         }
+
+        if (is_nigeria && !has_accepted_nimc_notice){
+            await showNIMCDialog().then(result => has_accepted_nimc_notice = result);
```
mustofa-binary, line 396:
```suggestion
            has_accepted_nimc_notice = await showNIMCDialog()
```

matin-binary (approved):

reza-fs (approved):

PR #6372, opened 2021-02-03
Daniil/Mt5 missing download option

Daniil-Deriv:


src/templates/static/metatrader/download.jsx
```diff
@@ -104,13 +104,15 @@ const DesktopDownloadBlock = ({
 );
 
 const MT5DesktopApp = ({ is_first_child, has_desktop_app }) => {
-    const txt_unsupported = it.L('The MT5 desktop app is not supported by macOS, Windows XP, Windows 2003, and Windows Vista.');
+    const txt_unsupported = it.L('The MT5 desktop app is not supported by Windows XP, Windows 2003, and Windows Vista.');
```
mustofa-binary, line 107: we do support macOs, any reason as of why do we remove it?

mustofa-binary (changes requested):

src/javascript/app/pages/resources/metatrader/download.ui.js
```diff
@@ -26,6 +26,8 @@ const toggleDownloadPage = target => {
         getElementById(`${target}-description`).setVisibility(1);
     }
     getElementById(`mt5_download_${target === 'mac' ? 'mac_' : ''}platforms`).setVisibility(1);
+    getElementById(`mt5_download_${target !== 'mac' ? 'mac_' : ''}platforms`).setVisibility(0);
```
Bahar-FS, line 29: Don't both of these lines mean the same?

src/templates/static/metatrader/download.jsx
Daniil-Deriv: @mustofa-binary don't know why it was ommited, but as I understand, download option for mac should be added to binary now

Bahar-FS (changes requested):

src/javascript/app/pages/resources/metatrader/download.ui.js
Daniil-Deriv: No, second string resolves the component duplication problem. 
image.   What happens without it

src/templates/static/metatrader/download.jsx
Bahar-FS: @mustofa-binary It seems that some months ago showing the download for mac removed, and now the QA asked for that.
--https://github.com/binary-com/binary-static/pull/5965/files

Bahar-FS (approved):

PR #6387, opened 2021-02-17
Hojjat / fix: setting fallback symbol

hojjat-fs:


src/javascript/app/pages/trade/symbols.js
```diff
@@ -51,13 +51,56 @@ const Symbols = (() => {
...
+
+            if (is_symbol_offered) return first_symbol;
+        }
+        
+        return undefined;
+    };
+
+    const getDefaultOpenSymbol = async active_symbols => {
+        const default_open_symbol =
+            (await findSymbol(active_symbols, '1HZ100V')) ||
```
negarn, line 87: what is this? why are we hardcoding symbols?

kevinw-binary (approved):

PR #6397, opened 2021-02-22
Daniil/Removed cashier references

Daniil-Deriv:


carol-binary (approved):

package-lock.json
```diff
@@ -3518,7 +3518,6 @@
           "resolved": "https://registry.npmjs.org/is-glob/-/is-glob-4.0.1.tgz",
           "integrity": "sha512-5G0tKtBTFImOqDnLB2hG6Bp2qcKEFduo4tZu9MT/H6NQv/ghhy30o55ufafxJ/LdH79LLs2Kfrn85TLKyA7BUg==",
           "dev": true,
-          "optional": true,
```
Bahar-FS, line 3521: Should we have this change here?

src/templates/app/cashier/payment_methods.jsx
```diff
@@ -25,30 +25,9 @@ const TableValues = ({ value }) => {
...
-        href={href}
-        target='_blank'
-        aria-disabled={!href}
-        title={title}
-        rel='noopener noreferrer'
-    />
-);
-
-const ReferenceLinks = ({ pdf_file, video_link }) => (
+const ReferenceLinks = () => (
```
Bahar-FS, line 28: Do we still need this empty method? Isn't it better to remove it if it's unnecessary?

Bahar-FS: And as I'm seeing the card, it asks for removing video links only, do we have any extra request for removing `PDF` as well?

Bahar-FS (changes requested):

src/templates/app/cashier/payment_methods.jsx
Daniil-Deriv: Yes, PDF should be removed as well, it's just not mentioned in the card

package-lock.json
Daniil-Deriv: No, we shouldn't. Second time I have this issue: haven't changed this file, haven't added it to staging, but it is in PR again. Will fix now

src/templates/app/cashier/payment_methods.jsx
Daniil-Deriv: About the method: I think we still need it so it would be really easy and fast to add new links to updated videos and PDFs, when they will be available. The main reason for deleting them was that all of them are out of date.

Bahar-FS: Alright, so isn't it better to just remove the props passed to the <ReferenceLinks /> (in the usages) and let these methods stand steal for future use?

Daniil-Deriv:
refactored the task by omitting the props, like someone did before me))

src/templates/app/cashier/payment_methods.jsx
Daniil-Deriv: done

Bahar-FS (approved):

PR #6428, opened 2021-03-12
suthesh/livechat-logout-bug

suthesan-deriv:
enable customer sdk on staging and production only

### DIFF
--- a/src/javascript/app/base/client.js
+++ b/src/javascript/app/base/client.js
@@ -4,7 +4,8 @@ const RealityCheckData   = require('../pages/user/reality_check/reality_check.da
 const ClientBase         = require('../../_common/base/client_base');
 const GTM                = require('../../_common/base/gtm');
 const SocketCache        = require('../../_common/base/socket_cache');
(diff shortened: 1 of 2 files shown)


src/javascript/app/base/client.js
```diff
@@ -89,14 +99,14 @@ const Client = (() => {
...
         RealityCheckData.clear();
-        LiveChat.endLiveChat().then(() => {
-            const redirect_to = getPropertyValue(response, ['echo_req', 'passthrough', 'redirect_to']);
-            if (redirect_to) {
-                window.location.href = redirect_to;
-            } else {
-                window.location.reload();
-            }
-        });
+        if (window.location.hostname === ('staging.binary.com' || 'binary.com')){
```
zaki-binary, line 102:
```suggestion
const url = new URL(window.location.href);
const domain = url.hostname.split('.').slice(-2).join('.');
        if (domain === 'binary.com') {
```
This way u only need to check for the `root` domain and no need to check explicitly for the subdomains

zaki-binary (approved):

PR #6442, opened 2021-03-24
Nuri / Replace Vanuatu License pdf file

njazuli-deriv:
Changes:
- Replace Vanuatu License pdf file


kevinw-binary:
According to @njazuli-deriv 
the previous pdf changes had been merged, but the changes are not reflected, so on this PR we'll keep the change for now

kevinw-binary (approved):

PR #6457, opened 2021-04-06
suthesh/livechat-engagement

suthesan-deriv:
added session variables for livechat engagement


src/javascript/_common/base/livechat.js
```diff
@@ -1,28 +1,35 @@
-const { init }     = require('@livechat/customer-sdk');
-const BinarySocket = require('./socket_base');
-const ClientBase = require('./client_base');
+const BinarySocket  = require('./socket_base');
+const ClientBase    = require('./client_base');
+const TrafficSource = require('../../app/common/traffic_source');
+const licenseID     = require('../utility').lc_licenseID;
```
negarn, line 4: this should be one line above since it's ../ and the one above is ../../

src/javascript/_common/base/livechat.js
```diff
@@ -1,28 +1,35 @@
...
 
 const LiveChat = (() => {
-    const licenseID = 12049137;
-    const clientID = '66aa088aad5a414484c1fd1fa8a5ace7';
-    let session_variables = { loginid: '', landing_company_shortcode: '', currency: '', residence: '', email: '' };
+    const utm_data = TrafficSource.getData();
+    const utm_source = TrafficSource.getSource(utm_data) || '';
+    const utm_campaign = utm_data.utm_campaign || '';
+    const utm_medium = utm_data.utm_medium || '';
+    let session_variables = { is_logged_in: false, loginid: '', landing_company_shortcode: '', currency: '', residence: '', email: '', utm_source: utm_source || '', utm_medium: utm_medium || '', utm_campaign: utm_campaign || '' };
```
negarn, line 11: utm source, campaign, and medium already fallback to '' in the declaration above. I think the repeated fallback '' for them is not needed here.

src/javascript/app/base/client.js
```diff
@@ -86,6 +88,33 @@ const Client = (() => {
         }
     };
 
+    // Called when logging out to end ongoing chats if there is any
+    const endLiveChat = () => new Promise ((resolve) => {
```
negarn, line 92: why did we decide to move this here?

suthesan-deriv: Cause importing from livechat.js into client.js causes a circular dependency...client -> livechat -> traffic source ->client. So I moved this here to break that chain

negarn (approved):

PR #6464, opened 2021-04-11
Hojjat / handle cashier locked statuses in deposit

hojjat-fs:


src/javascript/app/pages/cashier/cashier.js
```diff
@@ -182,7 +182,7 @@ const Cashier = (() => {
     const applyStateLockLogic = (status, deposit, withdraw) => {
         // statuses to check with their corresponding selectors
         const statuses_to_check = [
-            { lock: 'cashier_locked', selectors: [deposit, withdraw] },
+            { lock: 'cashier_locked', selectors: [withdraw] },
```
negarn, line 185: I think for non-uk we still want to disable it if deposit is locked, no?

negarn (dismissed):

PR #6467, opened 2021-04-12
Hojjat / fix: remove extra details columns

hojjat-fs:


src/javascript/app/pages/user/account/statement/statement.init.js
```diff
@@ -104,6 +104,11 @@ const StatementInit = (() => {
                 //     .on('click', () => { StatementUI.exportCSV(); });
             }
         }
+
+        if (filter === 'deposit' || filter === 'withdrawal'){
```
reza-fs, line 108:
```suggestion
        if (['deposit', 'withdrawal'].includes(filter)){
```

src/javascript/app/pages/user/account/statement/statement.init.js
```diff
@@ -104,6 +104,11 @@ const StatementInit = (() => {
                 //     .on('click', () => { StatementUI.exportCSV(); });
             }
         }
+
+        if (['deposit', 'withdrawal'].includes(filter)){
+            document.querySelectorAll('.details').forEach(item => item.remove());
```
negarn, line 109: why are we removing them? do they get added if the filter updates later to non deposit/withdrawal? would it better to not add the details in the first place instead of removing it? 

if we decide to keep this way, we should be more specific than removing all the detail classes on the page. for example, if statement table has a unique classname, we should use that before .details

negarn (approved):

PR #6468, opened 2021-04-13
Ali (Ako)/ Redirect career page to deriv

ali-hosseini-fs:
Career page link redirected to deriv.com/career

Also the old career page of binary deleted.


scripts/config/pages.js
```diff
@@ -80,7 +80,6 @@ module.exports = [
     ['about-us',                                'static/about/index',                          'full_width', 'About Us'],
     ['asuncion',                                'static/about/asuncion',                       'full_width', 'Careers - Asunción'],
     ['binary-in-numbers',                       'static/about/binary_in_numbers',              'default',    'Binary.com in Numbers'],
-    ['careers',                                 'static/about/careers',                        'full_width', 'Careers'],
```
negarn, line 83: we can't just delete the page. it could be used in marketing emails or other materials. let's keep the url and the page live, just change our links on the site to the deriv careers. we can consider asking devops to redirect binary.com careers page to deriv careers whenever someone lands on the url

negarn (approved):

PR #6472, opened 2021-04-15
Hubert/ Option receiving product emails

hubert-deriv:


src/javascript/_common/utility.js
```diff
@@ -40,6 +40,41 @@ const getHighestZIndex = (selector = 'div,p,area,nav,section,header,canvas,aside
...
+    'Hungary',
+    'Denmark',
+    'Sweden',
+    'Ireland',
+    'Isle of Man',
+    'Great Britain',
+    'Malta',
+];
+// check if client is from EU
+const isEuCountry = country => eu_countries.includes(country);
```
negarn, line 76: we already have a `isEuCountry` function which dynamically decides it based on the API. why are we repeating this by hardcoding the countries?

src/javascript/app/pages/user/new_account/virtual_acc_opening.js
```diff
@@ -73,6 +74,25 @@ const VirtualAccOpening = (() => {
                 },
             })
             .setVisibility(1);
+            
+        const get_selected_value = document.getElementById('select2-residence-container').getAttribute('title');
+        const residence_dropdown = document.getElementById('residence');
+
+        if (!isEuCountry(get_selected_value)) {
+            email_consent.classList.add('hide-checkbox');
+            consent_checkbox.classList.add('hide-checkbox');
```
negarn, line 83: why not just email_consent.setVisibility(0) and consent_checkbox.setVisibility(0), then we don't need to add a new class or change our form manager behaviour

src/templates/app/new_account/virtual.jsx
```diff
@@ -24,17 +24,15 @@ const Virtual = () => (
...
+            </Fieldset>
+            <div id='consent_checkbox'>
                 <FormRow
                     type='checkbox'
-                    checked
                     id='email_consent'
-                    row_class='invisible'
                     label_row_id='email_consent_label'
-                    label={it.L('Receive news and special offers')}
+                    label={it.L('I want to receive updates on binary.com products, services, and events.')}
```
negarn, line 33: binary.com should be a placeholder with it.website_name

src/javascript/_common/utility.js
hubert-deriv: The checkbox has to be coded in a way, that it appears by selecting a eu country from the selectbox. The current isEuCountry function gets the countrycode if the user is registered or logged in, otherwise, it uses ip address locating to determine the countrycode. Ashkan mentioned that he doesn't want the checkbox to appear due to VPN/ipaddress, hence, it's hard coded. A similar function is used in DerivApp, and I have copied it here. 

I will definitely refactor the code a bit though, and give this function a different name, since isEuCountry already exists.

src/javascript/app/pages/user/new_account/virtual_acc_opening.js
hubert-deriv: The reason I add this class is not only to hide the checkbox, but mainly to see specifically in the form manager when a checkbox has this specific class. When a hidden checkbox has this class, instead of sending 0 (since it's unchecked) it will send a 1 (even when it's unchecked). The reason behind this, is that Ashkan mentioned that the checkbox has to still send the value 1, when it's hidden after selecting a non-EU country.

src/templates/app/new_account/virtual.jsx
hubert-deriv: binary.com now replaced with the placeholder.

src/javascript/_common/utility.js
```diff
@@ -40,6 +40,41 @@ const getHighestZIndex = (selector = 'div,p,area,nav,section,header,canvas,aside
...
+    'hu',
+    'dk',
+    'se',
+    'ie',
+    'im',
+    'gb',
+    'mt',
+];
+// check if client is from EU
+const isEuCountryNoIp = country => eu_countries.includes(country);
```
negarn, line 76: I have a suggestion for the function name, please let me know if you agree, otherwise the changes are fine.

src/javascript/app/pages/user/new_account/virtual_acc_opening.js
```diff
@@ -13,6 +13,7 @@ const LocalStore               = require('../../../../_common/storage').LocalSto
 const State                    = require('../../../../_common/storage').State;
 const urlFor                   = require('../../../../_common/url').urlFor;
 const Utility                  = require('../../../../_common/utility');
+const isEuCountrySelected         = require('../../../../_common/utility').isEuCountrySelected;
```
negarn, line 16:
```suggestion
const isEuCountrySelected      = require('../../../../_common/utility').isEuCountrySelected;
```

negarn (dismissed):

PR #6474, opened 2021-04-16
fiona/ fix banner in mt5 dashboard on mobile

fiona-deriv:


src/javascript/app/base/socket_general.js
```diff
@@ -162,9 +161,6 @@ const BinarySocketGeneral = (() => {
                 }
                 break;
             }
-            case 'RateLimit':
```
mahdiaryayi-fs, line 165: Are you removing the whole functionality of the RateLimit alert or is it appearing somewhere else?
Don't we need this functionality anymore?

fiona-deriv: @mahdiaryayi-fs rate limit alert for mt5 seems to be appearing somewhere else

negarn: @fiona-deriv then we should remove that other one and keep this. this is used for our whole website.

fiona-deriv: okay thanks, noted

PR #6479, opened 2021-04-20
Amir / fix: use binary_direct for referrer if user doesn't have utm_source

amirghadimi-fs:
As a discussion with @mustofa-binary we removed referrer because it’s not mandatory.


src/javascript/app/common/traffic_source.js
```diff
@@ -47,13 +48,28 @@ const TrafficSource = (() => {
...
+        // When the user comes to the site with URL params
+        if (params.utm_source || params.utm_medium || params.utm_campaign) {
+
+            // if url is missing one of required fields, do nothing
+            for (let ctr = 0; ctr < param_keys.length; ctr++) {
+                if (!Object.hasOwnProperty.call(params, param_keys[ctr])) {
+                    is_need_change = false;
                 }
-            });
+            }
```
negarn, line 61: you can change this whole thing to:

```
// When the user comes to the site with URL params
if (params.utm_source || params.utm_medium || params.utm_campaign) {
    // if url is missing one of required fields, do nothing
    const is_need_change = param_keys.every((param) => param in params)
```

negarn (approved):

src/javascript/app/common/traffic_source.js
```diff
@@ -48,12 +49,23 @@ const TrafficSource = (() => {
...
+            if (is_need_change) {
+                param_keys.forEach((key) => {
+                    if (params[key]) {
+                        cookie.set(key, params[key], { sameSite: 'none', secure: true });
+                    }
+                });
+            }
+            
+        } else if (!Cookies.get('utm_data')) {
+            initCookie();
```
mustofa-binary, line 67: any reason why do we need initCookie() again here?, since have called initCookie on line 47

src/javascript/app/common/traffic_source.js
```diff
@@ -48,28 +48,28 @@ const TrafficSource = (() => {
...
-            param_keys.map((key) => {
-                if (params[key] && !current_values[key]) {
-                    cookie.set(key, params[key], { sameSite: 'none', secure: true });
-                }
-            });
+        // When the user comes to the site with URL params
+        if (params.utm_source || params.utm_medium || params.utm_campaign) {
+
+            // if url is missing one of required fields, do nothing
+            const is_need_change = param_keys.every((param) => param in params);
```
reza-fs, line 55:
```suggestion
            const has_all_params = param_keys.every((param) => param in params);
```

src/javascript/app/common/traffic_source.js
```diff
@@ -48,28 +48,28 @@ const TrafficSource = (() => {
...
+            
+            if (is_need_change) {
```
reza-fs, line 57:
```suggestion
            if (has_all_params) {
```

mustofa-binary (approved):

PR #6493, opened 2021-04-29
mahdiyeh/fix: alignment_issue_on_home_page

mahdiyeh-fs:
image
image


src/templates/static/home.jsx
```diff
@@ -100,7 +100,7 @@ const AccountsTabContent = ({
                         <img className='responsive' src={it.url_for(image)} />
                     </div>
                     <div className={`center-text gr-12 gr-hide gr-show-m ${mobile_class || ''}`}>
-                        <div className='gr-8 gr-centered'>
+                        <div className='gr-10 gr-centered margin-top-10'>
```
reza-fs, line 103: This'll affect desktop viewports as well, if it is specific to mobile, we can use `gr-10-m` and let `gr-8` be intact.

mahdiyeh-fs: the parent div just shown in the mobile and it is hidden in desktop mood, so I think no need for gr-10-m class

reza-fs (approved):

akmal-binary (approved):

akmal-binary (approved):

reza-fs (approved):

PR #6512, opened 2021-05-06
mahdiyeh/fix: visiblity_for_australia_clients

mahdiyeh-fs:
image

image

image


negarn (approved):

negarn (approved):

src/javascript/app/base/binary_loader.js
```diff
@@ -141,12 +142,22 @@ const BinaryLoader = (() => {
             loadActiveScript(config);
         }
         if (config.no_mf && Client.isLoggedIn() && Client.isAccountOfType('financial')) {
-            BinarySocket.wait('authorize').then(() => displayMessage(error_messages.no_mf()));
+            BinarySocket.wait('authorize').then(() => {
+                if (this_page === 'platforms') {
```
negarn, line 146: can we have a separate flag for residence blocked rather than checking the name of the page?

negarn (dismissed):

src/javascript/static/pages/platforms.js
```diff
@@ -19,8 +21,21 @@ const os_list = [
     // }
 ];
 
+const addLoader = () => {
+    const container = getElementById('content-holder');
+    const content = getElementById('content');
+    const loading_spinner = createElement('div', { html: '' });
+    container.appendChild(loading_spinner);
```
negarn, line 28: why do we need to make this so complicated? let's just add the loading element to the HTML code and then wait for authorize to hide it.

mahdiyeh-fs: sure, I will fix that

negarn (approved):

PR #6520, opened 2021-05-11
Amir / Extend form function to pass object structure to backend API

amirghadimi-fs:


src/javascript/app/common/form_manager.js
```diff
@@ -136,7 +136,11 @@ const FormManager = (() => {
             can_submit  = form.can_submit;
             if (!can_submit) return;
             if (Validation.validate(options.form_selector)) {
-                const req = $.extend({}, options.obj_request, getFormData(options.form_selector));
+                const req = $.extend({},
+                    options.obj_request,
+                    getFormData(options.form_selector),
+                    options.static_values
```
mustofa-binary, line 142: how about request_fields or other_request_fields?

mustofa-binary: and whats the difference between static_values and obj_request?

amirghadimi-fs: You’re right I'll remove `static_values` we can add a nested object in obj_request

PR #6521, opened 2021-05-11
Hubert / User should be logged out after deactivating account.

hubert-deriv:


src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -371,9 +372,25 @@ const AccountClosure = (() => {
...
+                $('#menu-top').removeClass('smaller-font top-nav-menu');
+                $('.client_logged_out').setVisibility(1);
+                $('topbar').addClass('primary-bg-color-dark');
+                $('topbar').removeClass('secondary-bg-color');
                 $.scrollTo(0, 500);
 
                 sessionStorage.setItem('closingAccount', 1);
+
+                // If the user clicks on a link while not having been logged out yet, log the user out.
+                $('a').on('click', () => {
```
amirghadimi-fs, line 385: Selecting  tag `a` is global, are you sure we have to log the user out on select any link on the page?

hubert-deriv: In this case I was thinking yes, since the user is still logged in technically until he/she clicks something or gets logged out automatically after 10 seconds. This selector is only relevant when the user is already on the deactivated account page.

amirghadimi-fs: (optional) I think if we need to show a `message` or any other information about that the account is closed, we can handle it using a promise instead of timeout , or we can logout user and redirect them to a page to show them the message (product team agreement is required). In all cases we don’t need to add a event to `a` tag globally.

amirghadimi-fs: *by promise I mean a modal with a CTA, closing modal is equal to CTA

hubert-deriv: I created a modal after the user deactivates his account, where the user clicks on a button to close it. After that the user gets redirected automatically to the home page.

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -370,15 +370,19 @@ const AccountClosure = (() => {
             } else {
                 $submit_loading.setVisibility(0);
                 $closure_container.setVisibility(0);
+                $('.client_logged_in, #client-logged-in, #topbar-msg').setVisibility(0);
                 $success_msg.setVisibility(1);
+                $('.client_logged_out').setVisibility(1);
+                $('#topbar').addClass('primary-bg-color-dark');
+                $('#menu-top').removeClass('smaller-font top-nav-menu');
+                $('#topbar').removeClass('secondary-bg-color');
```
negarn, line 378: why are we faking the logout? 🤔

src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -370,15 +370,19 @@ const AccountClosure = (() => {
...
 
                 sessionStorage.setItem('closingAccount', 1);
-                setTimeout(() => {
-                    // we need to clear all stored client data by performing a logout action and then redirect to home
-                    // otherwise it will think that client is still logged in and redirect to trading page
+
+                $('.close_main_modal').on('click', () => {
                     Client.sendLogoutRequest(false, Url.urlFor('home'));
-                }, 10000);
+                });
```
negarn, line 385: this should be sufficient, we don't need to update the style above to fake log them out. if we change our styles later it will break the above hardcodes.

src/sass/static/pages.scss
```diff
@@ -1624,6 +1624,35 @@ body #not_authenticated_financial ul.checked > li {
     }
 }
 
+#msg_main.close-account {
```
negarn, line 1627: it would be better to have a more unique identifier that's specific to the element rather than two generic identifiers grouped together

src/sass/static/pages.scss
```diff
@@ -1624,6 +1624,35 @@ body #not_authenticated_financial ul.checked > li {
...
+        left: 0;
+        right: 0;
+        top: 0;
+        bottom: 0;
+        background-color: rgba(0, 0, 0, 0.5);
+        z-index: 5;
+
+        .notice-msg {
+            top: 25%;
+            width: 22.5em;
```
negarn, line 1643: generally we use rem and not em for width

src/sass/static/pages.scss
```diff
@@ -1624,6 +1624,35 @@ body #not_authenticated_financial ul.checked > li {
...
+            height: auto;
+            padding: 1em;
```
negarn, line 1645: same here, padding should be either fixed (px) or rem

src/javascript/app/pages/user/account/settings/account_closure.js
hubert-deriv: Initially I didn't. know another way since the webpages are hardwired. Thus, if we would actually log the user out to show the logged out menu, this user wont see the message anymore, since that page is only available for logged in users.

What I did now is using the already available closingAccount session to display the logged out menu when the user closes his/her account.

src/sass/static/pages.scss
hubert-deriv: The modal has now a unique class.

src/sass/static/pages.scss
hubert-deriv: fixed.

negarn (dismissed):

src/javascript/static/pages/deactivated_account.js
```diff
@@ -0,0 +1,22 @@
+const urlFor = require('../../_common/url').urlFor;
+const Client = require('../../app/base/client');
+
+const DeactivatedAccount = (() => {
+    const onLoad = () => {
+        const redirect_home = urlFor('home');
+        const redirect_trading = urlFor('trading');
+        setTimeout(() => {
+            if (Client.isLoggedIn()) {
+                window.location.href = redirect_trading;
```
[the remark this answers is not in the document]
hubert-deriv: Updated the code.

negarn (approved):

PR #6530, opened 2021-05-17
Hubert / Hiding checkbox when page is loading

hubert-deriv:


src/templates/app/new_account/virtual.jsx
```diff
@@ -25,10 +25,11 @@ const Virtual = () => (
                     </p>
                 </div>
             </Fieldset>
-            <div id='consent_checkbox'>
+            <div id='consent_checkbox' className='hide-product-checkbox'>
                 <FormRow
                     type='checkbox'
                     id='email_consent'
+                    className='hide-product-checkbox'
```
akmal-binary, line 32: Hey, I'm wondering, why exactly do we need to give the same classname to both parent and child elements? Can we remove one of these or use different selector for child element?

src/javascript/app/pages/user/new_account/virtual_acc_opening.js
```diff
@@ -75,22 +75,22 @@ const VirtualAccOpening = (() => {
...
             if (eu_country) {
-                email_consent.classList.remove('hide-product-checkbox');
-                consent_checkbox.classList.remove('hide-product-checkbox');
+                email_consent.removeClass('hide-product-checkbox');
+                consent_checkbox.removeClass('hide-product-checkbox');
             } else {
-                email_consent.classList.add('hide-product-checkbox');
-                consent_checkbox.classList.add('hide-product-checkbox');
+                email_consent.addClass('hide-product-checkbox');
+                consent_checkbox.addClass('hide-product-checkbox');
```
negarn, line 91: what's the reason for changing js to jquery?

src/templates/app/new_account/virtual.jsx
hubert-deriv: Hello akmal, the classname is to display and hide the checkbox, but also for validation on the input. I changed the name of the class on the checkbox now to validate-hidden-product-checkbox. I also noticed that I can't directly parse a class into the FormRow component as I did here. It doesn't display in the DOM. I resolved this now through javascript.

akmal-binary: Could you give to one of these classes a more targeted name? Currently, both of them have `checkbox`, and I don't really see from the code whether we have two different checkboxes, or one of them refers to something else rather than the checkbox.

hubert-deriv: I changed the class names. Let me know if they are better this way or not.

akmal-binary (approved):

negarn (approved):

PR #6531, opened 2021-05-17
mahdiyeh/ fix: remove_outdated_banners

mahdiyeh-fs:
image
image


src/templates/app/user/metatrader.jsx
```diff
@@ -70,7 +69,6 @@ const CashierDesc = ({ title, desc, arrow_direction }) => (
 
 const Metatrader = () => (
     <React.Fragment>
-        <MT5Banner />
```
negarn, line 73: should we delete the element as well since it's not being used anywhere else?

src/javascript/app/common/deriv_banner.js
```diff
@@ -35,10 +35,11 @@ const DerivBanner = (() => {
     const showBanner = () => {
         if (deriv_banner_type === banner_types.rebranding) {
             el_banner_to_show = el_rebranding_banner_container;
```
negarn, line 37: why not delete the rebranding banner instead of hiding it?

negarn (dismissed):

src/templates/static/home.jsx
```diff
@@ -166,9 +165,7 @@ const Home = () => {
 
     return (
         <React.Fragment>
-            <div className='container'>
-                <DerivBanner has_margin />
-            </div>
+            <div className='container' />
```
negarn, line 168: do we still need this container?

mahdiyeh-fs: you're right, I remove it

negarn (approved):

PR #6540, opened 2021-05-21
mitchell/Read more redirection in signup

mitchell-deriv:


src/sass/_common/common.scss
```diff
@@ -588,6 +588,10 @@ ul.bullet.gr-row > li {
     display: list-item;
 }
 
+.toggle-arrow#tax_information_note_toggle {
+    font-weight: bold;
+}
+
 .toggle-arrow {
```
amirghadimi-fs, line 595:
```suggestion
.toggle-arrow {
    &#tax_information_note_toggle {
        font-weight: bold;
    }
```

negarn: do we need the toggle-arrow class mentioned at all? usually we prefer single identifiers, not grouping them together

mitchell-deriv: solved

akmal-binary (approved):

negarn (approved):

PR #6561, opened 2021-05-27
Mahdiyeh/fix: message_container_width

mahdiyeh-fs:
image
image


src/javascript/app/base/binary_loader.js
```diff
@@ -182,7 +182,7 @@ const BinaryLoader = (() => {
             return;
         }
 
-        const div_container = createElement('div', { class: 'logged_out_title_container', html: Client.isAccountOfType('financial') || Client.isOptionsBlocked() ? '' : content.getElementsByTagName('h1')[0] || '' });
+        const div_container = createElement('div', { class: 'container logged_out_title_container', html: Client.isAccountOfType('financial') || Client.isOptionsBlocked() ? '' : content.getElementsByTagName('h1')[0] || '' });
```
amirghadimi-fs, line 185: By adding `container` we overwrite all `logged_out_title_container` styles, could you please check if need this class or we can remove it?
I see there is another usage of it on Line 205. can you please check if we can replace it with container and remove `logged_out_title_container` and styles of it from css as well?

mahdiyeh-fs: as consistency of code I keep it and add container class to the parent

negarn (approved):

PR #6574, opened 2021-06-02
yashim/fix: downgrade onfido version

yashim-deriv:


package.json
```diff
@@ -127,7 +127,7 @@
     "js-cookie": "2.2.0",
     "kinetic": "5.2.0",
     "moment": "2.19.4",
-    "onfido-sdk-ui": "^6.7.2",
+    "onfido-sdk-ui": "^6.3.0",
```
mustofa-binary, line 130: why don't we only import onfido-sdk-ui in account package?

src/sass/static/pages.scss
```diff
@@ -973,28 +970,13 @@ body #not_authenticated_financial ul.checked > li {
             color: $COLOR_BLACK;
             border-radius: 0;
         }
-        .onfido-sdk-ui-Theme-textButton:hover {
-            color: $COLOR_BLACK;
-        }
-        .onfido-sdk-ui-CustomFileInput-container {
```
mustofa-binary, line 979: you can comment this with some TODO message

yashim-deriv: Done 😃

PR #6586, opened 2021-06-08
mamali/Hide currency when no currency is set

reza-fs:


src/javascript/app/pages/user/account/settings/self_exclusion.js
```diff
@@ -52,7 +52,7 @@ const SelfExclusion = (() => {
 
         currency = Client.get('currency');
 
-        $('.append_currency').after(Currency.formatCurrency(currency));
+        $('.append_currency').after(currency ? Currency.formatCurrency(currency) : '');
```
mahdiaryayi-fs, line 55: IMO, it is better to add this ternary in the `formatCurrency` itself in `src/javascript/_common/base/currency_base.js` (instead of adding a ternary here) to prevent this bug in all other similar places in the future:

changing 

```js
const formatCurrency = currency => `<span class="symbols">&nbsp;${getCurrencyDisplayCode(currency)}</span>`; 
```

to

```js
const formatCurrency = currency => currency ? `<span class="symbols">&nbsp;${getCurrencyDisplayCode(currency)}</span>` : '';
```

mahdiaryayi-fs (dismissed):

src/javascript/app/pages/user/account/settings/self_exclusion.js
reza-fs: Thanks for the tip, yeah will see to add this.

mahdiaryayi-fs (approved):

PR #6595, opened 2021-06-09
Nikita/Add funds redirect button in MT5 manage funds not redirecting to cashier page in one click.

nikita-deriv:
Link "add funds" opens on first click


src/javascript/index.js
```diff
@@ -24,3 +24,4 @@ $(window).on('pageshow', (e) => { // Safari doesn't fire load event when using b
         BinaryLoader.init();
     }
 });
+
```
akmal-binary, line 27: There is an extra empty line in the end, can we remove it?

src/javascript/app/common/guide.js
```diff
@@ -43,12 +44,8 @@ const Guide = (() => {
             $(opt.guideBtnID).remove();
             return;
         }
-
         makeButton();
     };
-    /*
```
akmal-binary, line 49: Any reason we are removing this comment? Seems useful and consistent with other functions. Maybe we can discard changes on this file if it's not intentional.

src/javascript/index.js
nikita-deriv: Looks like was added unnecessary changes, removed all, thanks.

src/javascript/app/common/guide.js
nikita-deriv: Returned.

akmal-binary (approved):

PR #6600, opened 2021-06-09
Mahdiyeh/fix: mlt_and_uk_market

mahdiyeh-fs:


src/javascript/_common/base/client_base.js
```diff
@@ -403,6 +403,13 @@ const ClientBase = (() => {
         return options_blocked_countries.includes(country);
     };
 
+    const isMalta = () => {
+        const malta_countries = ['at', 'be', 'bg', 'cy', 'cz', 'dk', 'ee', 'fi', 'hr', 'hu', 'ie', 'lt', 'lv', 'nl', 'pl', 'pt', 'ro', 'se', 'si', 'sk'];
```
akmal-binary, line 407: Would it be possible to use `landing_company_shortcode` to check if it's `malta` instead?
I think you can access it like that `const current_landing_company = get('landing_company_shortcode');`. Or perhaps we can check it directly in `markets.jsx` if it's accessible there.

mahdiyeh-fs: the landing_company_shortcode is 'virtual' when we using virtual account in any landing_company and we can not understand it is the virtual account of Malta client

src/javascript/app/pages/trade/markets.jsx
```diff
@@ -80,9 +84,19 @@ class Markets extends React.Component {
...
-        this.markets_all = non_crypto_markets_arr.slice();
+        final_markets_arr = market_arr.filter(market => is_not_crypto(market));
+        final_market_obj = Object.fromEntries(final_markets_arr);
+        if ((Client.isMalta() || is_uk) && Client.getAccountOfType('virtual')) {
+            final_markets_arr = market_arr.filter(market => is_synthetic(market));
+            final_market_obj = Object.fromEntries(final_markets_arr);
+            market_symbol = Object.keys(final_market_obj)[0];
+            const submarket = Object.keys(final_market_obj[market_symbol].submarkets).sort(sortSubmarket)[0];
+            underlying_symbol = Object.keys(final_market_obj[market_symbol].submarkets[submarket].symbols).sort()[0];
+        }
```
akmal-binary, line 98: How about we move lines 90, 91 into else block here to avoid extra `market_arr filter` when the client is from malta?

akmal-binary:
```suggestion
        } else {
           final_markets_arr = market_arr.filter(market => is_not_crypto(market));
           final_market_obj = Object.fromEntries(final_markets_arr);
        }
```

src/javascript/app/pages/trade/markets.jsx
```diff
@@ -80,9 +84,19 @@ class Markets extends React.Component {
             underlying_symbol = Object.keys(this.markets[market_symbol].submarkets[submarket].symbols).sort()[0];
         }
         const is_not_crypto = symbol => !/^(cry|JD)/i.test(symbol);
+        const is_synthetic = symbol => /^(synthetic)/i.test(symbol);
+        const is_uk = State.getResponse('authorize.country') === 'gb';
         const market_arr = Object.entries(this.markets).sort((a, b) => sortSubmarket(a[0], b[0]));
-        const non_crypto_markets_arr = market_arr.filter(market => is_not_crypto(market));
-        this.markets_all = non_crypto_markets_arr.slice();
+        final_markets_arr = market_arr.filter(market => is_not_crypto(market));
```
akmal-binary, line 90:
```suggestion
```

src/javascript/app/pages/trade/markets.jsx
```diff
@@ -80,9 +84,19 @@ class Markets extends React.Component {
...
+        final_market_obj = Object.fromEntries(final_markets_arr);
```
akmal-binary, line 91:
```suggestion
```

src/javascript/_common/base/client_base.js
akmal-binary: Oh yeah, this makes sense. We'll leave it like that then 👍🏻

src/javascript/app/pages/trade/markets.jsx
mahdiyeh-fs: yeah , you're right, will change it

akmal-binary (dismissed):

src/javascript/_common/base/client_base.js
reza-fs, line 407: in `landing_company` response, you can check for `gaming_company` and then see if the shorcode inside that object contains `shortcode: "malta"`.  -> `landing_company.gaming_company.shortcode` is the key here. 

This works for Virtual clients too.

reza-fs (changes requested):

src/javascript/_common/base/client_base.js
mahdiyeh-fs: sure, work on that

reza-fs (approved):

akmal-binary (approved):

PR #6606, opened 2021-06-13
Prince/ PA application form

prince-deriv:


reza-fs (approved):

scripts/render.js
```diff
@@ -250,7 +250,7 @@ async function compile(page) {
 
             affiliate_signup_url  : `https://login.binary.com/signup.php?lang=${affiliate_language_code}`,
             affiliate_password_url: `https://login.binary.com/password-reset.php?lang=${affiliate_language_code}`,
-            affiliate_email       : 'partners@binary.com',
+            affiliate_email       : 'paymentagents@binary.com',
```
ali-fs, line 253: I think we need to add a new key for `payment_agent_email`. The affiliate email is `partners@binary.com` and we need it in other pages

ali-fs (changes requested):

scripts/render.js
prince-deriv: I see. Let me add a new key

ali-fs (approved):

PR #6608, opened 2021-06-14
Sara / Added ids to onfido documents

sara-fs:


src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -911,11 +913,19 @@ const Authenticate = (() => {
         }
     };
 
-    const handleComplete = () => {
+    const handleComplete = (data) => {
+        const docIds = [];
```
mustofa-binary, line 917:
```suggestion
        const document_ids = [];
```

src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -911,11 +913,19 @@ const Authenticate = (() => {
         }
     };
 
-    const handleComplete = () => {
+    const handleComplete = (data) => {
+        const document_ids = [];
+        Object.keys(data).forEach(key => {
+            document_ids.push(data[key].id);
+        });
```
yashim-deriv, line 920:
```suggestion
        const document_ids = Object.keys(data).map(key => data[key].id);
```
i think it would be cleaner if we do it like this

mustofa-binary (approved):

yashim-deriv (approved):
LGTM~~

PR #6612, opened 2021-06-15
Hubert / Adding function that hides multiple error messages.

hubert-deriv:


mahdiaryayi-fs:
@hubert-deriv 

I found that the original issue is something else!
The issue is not that the user can see 2 error messages at the same time
The issue is the regex we are using for Email address validation
That regex is not working properly for detecting all wrong email addresses
just replace the regex for email validation in binary-static/src/javascript/app/common/form_validation.js line 168
from 

```javascript
const validEmail = value => /^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,63}$/.test(value);
```

to 

```javascript
const validEmail = value => /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/.test(value);
```

and the issue is fixed
That regex is from this StackOverflow post: https://stackoverflow.com/questions/46155/how-to-validate-an-email-address-in-javascript

So we don't need to hide one of the error messages anymore.

mahdiaryayi-fs (approved):

PR #6616, opened 2021-06-17
Ali / fix livechat error

ali-fs:


src/javascript/_common/base/livechat.js
```diff
@@ -44,8 +44,10 @@ const LiveChat = (() => {
...
 
-        setSessionVariables();
-        setNameEmail();
+        if (window.LiveChatWidget){
+            setSessionVariables();
+            setNameEmail();
+        }
 
         window.LC_API.on_chat_ended = () => {
             setNameEmail();
```
mustofa-binary, line 53: we also need to check it here i think

mustofa-binary (changes requested):

src/javascript/_common/base/livechat.js
ali-fs: Check added.

PR #6620, opened 2021-06-18
Kevin/Account closure peding withdrawal(s) error message

kevinw-binary:


src/javascript/app/pages/user/account/settings/account_closure.js
```diff
@@ -245,6 +245,13 @@ const AccountClosure = (() => {
                 addSection(account, txt_balance);
             });
         }
+        if (response.error.details.pending_withdrawals) {
+            Object.keys(response.error.details.pending_withdrawals).forEach((account) => {
+                const txt_pending_withdrawals = `${response.error.details.pending_withdrawals[account]} withdrawal(s)`;
```
akmal-binary, line 250: How do you think it will translate?

kevinw-binary: Seems like the other warning messages inside this file aren't translated either, going to check it

akmal-binary (approved):

PR #6621, opened 2021-06-18
Hubert / Hiding crash boom and step indices for MLT and MX accounts.

hubert-deriv:


src/javascript/app/pages/resources/trading_times/trading_times.ui.js
```diff
@@ -126,7 +127,7 @@ const TradingTimesUI = (() => {
 
     const createMarketTables = (market) => {
         const $market_tables = $('<div/>');
-
+        const gaming_account = ClientBase.getAccountType() === 'gaming';
```
mitchell-deriv: can you put this in init function or where you declared initially

mitchell-deriv (approved):

PR #6622, opened 2021-06-18
Nikita / Binary.com trade  page be accessible by client with meaning full message

nikita-deriv:
added deriv multiplier banner on MT5 page for MF and 'au' residents


src/javascript/_common/base/client_base.js
```diff
@@ -82,6 +82,8 @@ const ClientBase = (() => {
         return account_type;
     };
 
+    const getBrokerCode = loginid => loginid.match(/[A-Z]*/)[0];
+
```
hubert-deriv, line 86: I think we have a function called `getAccountType()` I think it kind of does the same as what you do here, but instead translates the broker codes to account types. Perhaps, you can use this function without writing a new one?

src/javascript/app/pages/user/metatrader/metatrader.js
```diff
@@ -2,6 +2,7 @@ const MetaTraderConfig   = require('./metatrader.config');
 const MetaTraderUI       = require('./metatrader.ui');
 const Client             = require('../../../base/client');
 const BinarySocket       = require('../../../base/socket');
+const DerivBanner        = require('../../../common/deriv_banner');
```
hubert-deriv, line 5: I would import the ClientBase here aswell then, to make use of the getAccountType function.

src/javascript/app/pages/user/metatrader/metatrader.js
```diff
@@ -17,6 +18,10 @@ const MetaTrader = (() => {
     const getAccountsInfo = MetaTraderConfig.getAccountsInfo;
 
     const onLoad = () => {
+        const current_account = Client.getBrokerCode(Client.getAllLoginids()[0]);
+        if (current_account === 'MF' || Client.isOptionsBlocked()){
+            DerivBanner.onLoad();
+        }
```
hubert-deriv, line 24: I think doing something like this, should also yield the same result you want to achieve.

```
const financial_account = ClientBase.getAccountType() === 'financial';   

if (financial_account || Client.isOptionsBlocked()) {            
   DerivBanner.onLoad();        
}
```

src/javascript/_common/base/client_base.js
```diff
@@ -411,6 +413,7 @@ const ClientBase = (() => {
         get,
         getAllLoginids,
         getAccountType,
+        getBrokerCode,
```
hubert-deriv, line 416:
```suggestion
```

src/javascript/_common/base/client_base.js
nikita-deriv: Thanks for suggestion, Hubert! It's more right way to do a check with `getAccountType()`. I don't understand meaning of account types in the first time. Removed.

src/javascript/app/pages/user/metatrader/metatrader.js
nikita-deriv: Replaced logic, based on your suggestion, thanks.

src/javascript/_common/base/client_base.js
nikita-deriv: Removed.

hubert-deriv (approved):

PR #6623, opened 2021-06-22
Ali / trackjs mt5 bugfix

ali-fs:


akmal-binary (dismissed):

src/javascript/app/pages/user/metatrader/metatrader.ui.js
```diff
@@ -89,7 +89,8 @@ const MetaTraderUI = (() => {
             used     : 0,
         };
 
-        State.getResponse('trading_servers').forEach(trading_server => {
+        const trading_servers = State.getResponse('trading_servers') || (await BinarySocket.send({ trading_servers: 1 })).trading_servers;
```
mustofa-binary, line 92: there would still be a possibility that `trading_servers` is undefined if websocket call has an error, and i think its better to try catch an async call

mustofa-binary (changes requested):

mustofa-binary (approved):

PR #6627, opened 2021-06-22
Prince/ Incorrect error message payment agent transfer

prince-deriv:


src/javascript/app/common/form_validation.js
```diff
@@ -231,13 +231,21 @@ const Validation = (() => {
             message = localize('Should be less than [_1]', addComma(options.max, options.format_money ? getDecimalPlaces(Client.get('currency')) : undefined));
         }
 
+        // Priority Validation
+        if ('balance' in options && isLessThanBalance(value, options)) {
+            is_ok   = false;
```
carol-binary, line 236: Can you change this variable to something clearer. Like is_**what**_ok?

prince-deriv: Hi @carol IMO we can leave it as it is otherwise we will be needing to change the variable name in the entire project. The whole validation helper have different functions inside that use the same `is_ok` variable.

carol-binary: Sure, we can leave it as it is now but we should consider better variables.

carol-binary (dismissed):

carol-binary (approved):

PR #6653, opened 2021-07-05
Amina/ social_icons_text_update

amina-deriv:


src/sass/_common/common.scss
```diff
@@ -772,7 +772,7 @@ li strong.margin-right {
         background-color: $COLOR_GRAY;
     }
     .name {
-        text-transform: capitalize;
+        text-transform: capitalize !important;
```
akmal-binary, line 775: Hi, I don't think this change would fix the mentioned issue with translation. Can we check with translation team whether these social network names are translated?

amina-deriv: Hi  @akmal-binary, We are not translating social icon text in our sites. On checking with the QAteam, the actual issue is that for few languages, the social icon text starts with a lowercase letter.

akmal-binary: Alright, could you also mention this on the redmine card, because later it might cause confusions among QA team.

akmal-binary (dismissed):

src/sass/_common/common.scss
amina-deriv: sure. Thank you.

ridho-binary: thanks @akmal-binary  for headsup , i was actually confused , i thought we are going to translate the social media button 😅 .
@amina-deriv i will update the remd card description as totally miss-leading 😄

akmal-binary (approved):

PR #6684, opened 2021-07-15
Hubert / e-mail input wouldn't show error after making valid e-mail address invalid again.

hubert-deriv:
When the user fills in a valid e-mail, the error message will disappear, indicating the validity of the e-mail. Although, when the user removes some characters again, or adds invalid characters, the error wouldn't show up again.


src/javascript/app/common/form_validation.js
```diff
@@ -363,15 +364,15 @@ const Validation = (() => {
...
         }
 
         return all_is_ok;
     };
 
-    const clearError = (field) => {
+    const clearError = (field, localized_message) => {
         if (field.$error && field.$error.length) {
-            field.$error.setVisibility(0);
+            field.$error.html(localized_message).setVisibility(0);
```
akmal-binary, line 375: I'm a bit confused... why do we need to set the error message before the message is cleared?

hubert-deriv: the showError function uses the setVisibility functions on the .html() contents that is equal to localized_message. In order for the error message to be properly cleared, the clearError function needs to be the same.

If the setVisibility function is set on the $error element itself, the error message will stay invisible if it gets triggered, because the showError function is set on the contents of this element '.html(localized_message)', not the element itself like clearError.

src/javascript/app/common/form_validation.js
```diff
@@ -314,7 +315,7 @@ const Validation = (() => {
         }
 
         let all_is_ok = true;
-        let message   = '';
+        const message = '';
```
akmal-binary, line 318: Do we still need this message if it is always going to be empty? 😅

src/javascript/app/common/form_validation.js
```diff
@@ -346,13 +347,17 @@ const Validation = (() => {
             }
 
             if (!field.is_ok) {
-                message = options.message || ValidatorsMap.get(type).message;
+                let message_template;
+                field.$error.text(options.message || ValidatorsMap.get(type).message);
                 if (type === 'length') {
-                    message = template(message, [options.min === options.max ? options.min : `${options.min}-${options.max}`]);
+                    message_template = template(message, [options.min === options.max ? options.min : `${options.min}-${options.max}`]);
```
akmal-binary, line 353: Comparing to the previous version, on this stage the message already had a value `options.message || ValidatorsMap.get(type).message` but now you will always pass an empty constant. Doesn't it change the message content?

src/javascript/app/common/form_validation.js
```diff
@@ -346,13 +347,17 @@ const Validation = (() => {
...
+                    field.$error.text(message_template);
```
akmal-binary, line 354: I see you're setting this in every condition below as well, could you call it once below the entire `if statement` instead, on line 362?

src/javascript/app/common/form_validation.js
hubert-deriv: Removed this now

src/javascript/app/common/form_validation.js
hubert-deriv: The message constant is not needed anymore, I removed it. Instead, I just pass an empty string in the template functions now.

src/javascript/app/common/form_validation.js
hubert-deriv: Fixed

src/javascript/app/common/form_validation.js
```diff
@@ -346,22 +346,23 @@ const Validation = (() => {
...
+                    message_template = template('', [options.min === options.max ? options.min : `${options.min}-${options.max}`]);
                 } else if (type === 'min') {
-                    message = template(message, [options.min]);
+                    message_template = template('', [options.min]);
                 } else if (type === 'not_equal') {
-                    message = template(message, [options.name1, options.name2]);
+                    message_template = template('', [options.name1, options.name2]);
                 }
                 all_is_ok = false;
+                field.$error.text(message_template);
```
akmal-binary, line 358: I think the initial idea about passing message to `showError` function as a parameter was fine. In that case we could move this line inside the function and just execute it before line 384. Do you think this would work?

hubert-deriv: I think that would work indeed, I moved the line now into there.

akmal-binary (dismissed):

akmal-binary (approved):

PR #6695, opened 2021-07-21
Mahdi | Mamali / Adding new trading password to binary

mahdiaryayi-fs:


akmal-binary (approved):

balakrishna-binary (approved):

matin-binary (approved):

akmal-binary (approved):

src/javascript/app/pages/user/metatrader/metatrader.js
```diff
@@ -289,6 +289,20 @@ const MetaTrader = (() => {
                 }
 
                 const req = makeRequestObject(acc_type, action);
+                if (action === 'new_account' && MetaTraderUI.shouldSetTradingPassword()) {
+                    const resp = await BinarySocket.send({ mt5_login_list: 1 });
```
reza-fs, line 293: (suggestion/question) Possible to handle the error case here? also using spread to define a better naming for response? 
```suggestion
                    const {mt5_login_list, error} = await BinarySocket.send({ mt5_login_list: 1 });
                    if (error) // needs handling? 
```

src/javascript/app/pages/user/metatrader/metatrader.ui.js
```diff
@@ -143,6 +140,47 @@ const MetaTraderUI = (() => {
...
+    const populatePasswordManager = () => {
+        const $mng_passwd  = $container.find('#frm_manage_password');
+        const $step_1 = $mng_passwd.find('.step-1');
+        const $step_2 = $mng_passwd.find('.step-2');
+        const $button = $mng_passwd.find('#password_change_button');
+        const $confirm_button = $mng_passwd.find('#password_change_confirm_buttons .btn_ok');
+        const $cancel_button = $mng_passwd.find('#password_change_confirm_buttons .btn_cancel');
+        const setStep = (step) => {
+            switch (step) {
+                case 1:
```
reza-fs, line 157: (Suggestion) 
Possible to map these numbers in a constant object for a better understanding of the current step? e.g.:
```
case STEPS.SERVER_SELECTION: 
```

src/javascript/app/pages/user/metatrader/metatrader.js
mahdiaryayi-fs: @reza-fs spread handling is applied but I can't think of an error handling in this phase. because this is just getting MT5 login list without any user-generated or dynamic payload and so it is almost error-proof!
By the way, we are not handling errors in many other places `mt5_login_list` is called and I think it is OK

mustofa-binary (approved):

PR #6699, opened 2021-07-25
Sara|Hojjat / update readme 

hojjat-fs:


README.md
```diff
@@ -1,103 +1,151 @@
...
+
+## Pre-installation
+
+Before running or contribute to this project, you need to have the setup of the following packages in your environment
 
 - Ruby, RubyGems
 - Sass (`sudo gem install sass`)
-- Node.js (10.14.2 or higher is recommended) and NPM (see <https://nodejs.org/en/download/package-manager/>)
-- Go to project root, then run `npm install`
+- Node.js (10.14.2 or higher is recommended)
```
ali-fs, line 27: for node > 12, we have an error for installing node-sass because our node-sass version is not up to date and doesn't support node>12. Do we need to mention this here? if we use node 12, we don't need to install sass separately with gem.  or also we can think about upgrading our sass version

README.md
```diff
@@ -1,103 +1,151 @@
...
-=============================
+5.  **To start developing:**
+
+    ```sh
+    npm run start
+    ```
+
+6.  **Open the source code and start editing!**
+
+    Your site is now running at `http://localhost:3000`!
```
ali-fs, line 63: isn't it `https://localhost:443`?

src/templates/README.md
```diff
@@ -1,93 +1,106 @@
...
+
+```js
+{ array.map[() => ())}
+```
+
+## Rendering conditionally
+
 - Use `{condition && <el/> ...</el>}` syntax to conditionally render an element.
-- Use `<el attr={value || undefined}` to conditionally render an attribute, React omits attributes with null or undefined values.
+- Use `<el attr={value || undefined}` to conditionally render an attribute, React
```
ali-fs, line 15: check this please

src/templates/README.md
```diff
@@ -1,93 +1,106 @@
...
+
+```js
+it.dangreouslyRenderHtml('<a href="https://binary.com"></a>');
+```
+
+## Props, arguments
+
 - Do NOT use `<Element attributeName={true} />`; just use `<Element attributeName />`.
+- Use destructuring to capture parameters if needed.
+  omits attributes with null or undefined values.
```
ali-fs, line 32: do you mean `React omits attributes with null or undefined values.`?

README.md
```diff
@@ -1,103 +1,152 @@
...
+
+5.  **To start developing:**
+
+    ```sh
+    npm run start
+    ```
+
+6.  **Open the source code and start editing!**
+
+    Your site is now running at `http://localhost:443`!
```
ali-fs, line 64: without `https`, it won't redirect to ssl.
```suggestion
    Your site is now running at `https://localhost:443`!
```

ali-fs (approved):

PR #6726, opened 2021-08-02
Prince/payment method script

prince-deriv:


package.json
```diff
@@ -8,7 +8,8 @@
     "serve": "sudo -p 'enter password for sudo:' grunt serve",
     "start": "sudo -p 'enter password for sudo:' grunt start",
     "eslint": "eslint --fix \"src/**/*.jsx\" src/",
-    "build": "grunt release_test_link --production"
+    "build": "grunt release_test_link --production",
+    "build:payment_methods": "node scripts/payment_method"
```
mahdiaryayi-fs, line 12: IMO, the `build` command should be responsible for building **everything** (like it is a general pattern in most npm packages). so I recommend renaming the previous build and creating a new general one:

```json
    "build:production": "grunt release_test_link --production",
    "build:payment_methods": "node scripts/payment_method",
    "build": "npm run build:production && npm run build:payment_methods",
```

prince-deriv: @mahdiaryayi-fs will change it into this

PR #6728, opened 2021-08-03
Amina/logout client on reset binary.com password from all existing active sessions.

amina-deriv:


src/javascript/app/pages/user/reset_password.js
```diff
@@ -49,7 +50,10 @@ const ResetPassword = (() => {
                 localized_message: localize('You have a new binary password.'),
                 localized_title  : localize('Binary password'),
                 ok_text          : localize('Done'),
-                onConfirm        : () => Login.redirectToLogin(true),
+                onConfirm        : () => {
+                    Login.redirectToLogin(true);
+                    Client.sendLogoutRequest(true);
```
akmal-binary, line 55: Do we still need `Login.redirectToLogin` above? Because `sendLogoutRequest` can take a redirectTo url as a second parameter, so we can pass it there.

amina-deriv: Hi @akmal,
Thanks for bringing it up. It seems on passing **true**   to `Client.sendLogoutRequest();` , redirect to login page is done inside the code.  So looks like its good to go with **Client.sendLogoutRequest(true);. Could you please verify?

reza-fs (dismissed):

PR #6751, opened 2021-08-18
Maryia/Removed editable fields from Contact CS message

maryia-deriv:


src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -129,6 +129,16 @@ const PersonalDetails = (() => {
         }
     };
 
+    const modifyContactCSMessage = (immutable_fields) => {
+        if (!/tax/.test(immutable_fields.toString())) {
+            $('.rowCustomerSupport').text((index, text) => text.replace(localize('To change your name, date of birth, country of residence, email, or tax information'),
+                localize('To change your name, date of birth, country of residence, or email')));
```
mustofa-binary, line 135: I dont think this will cater the case for translated text, since we are checking the html text

mustofa-binary (changes requested):

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -129,6 +129,16 @@ const PersonalDetails = (() => {
...
+        }
+        if (!/name/.test(immutable_fields.toString())) {
+            $('.rowCustomerSupport').text((index, text) => text.replace(localize('To change your name, date of birth'), localize('To change your date of birth')));
```
mustofa-binary, line 138: else if?

mustofa-binary (changes requested):

src/javascript/app/pages/user/account/settings/personal_details.js
maryia-deriv: I expect both 'if' to fire if both of them are truthy. 'Else if' will make the second part be skipped. That's why I used 'if's only.

src/javascript/app/pages/user/account/settings/personal_details.js
maryia-deriv: After the text is translated by translators, it's going to work because both arguments of the .replace() method are first computed and then applied. The main thing here to make it work is to ask translators to translate the parts of the Contact CS message, which I passed into localize(), in the exact same way they translated this parts inside the entire Contact CS message before.

akmal-binary: @maryia-deriv Does this replace only a part of the sentence? I'm not sure the way our translators translate but I think there might be cases when the order of the words is altered in other languages, so it won't match with a part of it.

akmal-binary: Perhaps, we can translate the entire sentences fully and just provide the relevant one beforehand, instead of modifying DOM on the fly.

src/javascript/app/pages/user/account/settings/personal_details.js
mustofa-binary: it seems that both if's are changing $('.rowCustomerSupport') element. therefore if second if is true, it will overwrite first if

maryia-deriv: I expect it as well. that's why the sequence of 'if's is important here.

src/javascript/app/pages/user/account/settings/personal_details.js
maryia-deriv: It would work well If translators simply copied & pasted their previous translations from the full initial string to the new parts of it, and it would be enough to make translation the same for the parts of the sentence in the first localize() argument of each .replace() method only. Translation of the second localize() argument wouldn't impact the result in this case.
However, I changed my PR the way you suggested by adding more options for the Contact CS string in order to not modify DOM on the fly.

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -129,6 +129,18 @@ const PersonalDetails = (() => {
         }
     };
 
+    const selectContactCSMessage = (immutable_fields) => {
+        if (/name/.test(immutable_fields.toString()) && /tax/.test(immutable_fields.toString())) {
```
mustofa-binary, line 133: thanks for the change
nit suggestion: we can make these regex text into variables e.g `const has_name = /name/.test(immutable_fields.toString())`

maryia-deriv: This is so simple but so important. Don't know how I missed it in the first place. Thank you for pointing at it! All set :)

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -129,6 +129,20 @@ const PersonalDetails = (() => {
         }
     };
 
+    const selectContactCSMessage = (immutable_fields) => {
+        const has_name = /name/.test(immutable_fields.toString());
+        const has_tax_info = /tax/.test(immutable_fields.toString());
+        if (has_name && has_tax_info) {
+            $('.rowCustomerSupport').not('.full-cs-msg').setVisibility(0);
```
akmal-binary, line 136: A question: why don't we hide all the messages by default and then make one visible in this function? Currently, if this logic somehow fails it's going to show all of the messages at the same time 😄

mustofa-binary: all of these  are invisible by default, she put class invisible on each of these elements, @maryia-deriv should these be `setVisibility(1)`?

akmal-binary: True, they are invisible, but here it just makes them invisible again 😕

maryia-deriv: **The current flow is the following:**
1) First, all elements for authenticated account only (marked with the `.RealAccAuth` className) are made invisible including unnecessary Contact CS strings because if there are any other elements marked with the `.RealAccAuth` className, they also must be shown; _(line 183 in personal_details.js)_
2) Then, I hide all unnecessary Contact CS strings (among those marked with the `.rowCustomerSupport` className) while leaving the relevant Contact CS string only. _(line 184 in personal_details.js)_
**The reason for doing that:**
I suppose that the author of the initial code designed the `.RealAccAuth` className to mark any elements that must be visible for authenticated accounts only.
However, for some reason, the .RealAccAuth className is now used for the Contact CS string only. If you think that there will be no additional elements in the future that must be also visible only for authenticated accounts, I can remove the `.RealAccAuth` className or the `.rowCustomerSupport` className because at the moment both of them are used by Contact CS string only. And in such a case, knowing that the Contact CS string is the only element that must visible for authenticated accounts only, I will be able to first hide unnecessary Contact CS strings, and then show the relevant one. What do you suggest?

akmal-binary: Why can't we just remove the line 183? I see that `.RealAccAuth` only refers to these paragraphs and nothing else, so it's just an extra manipulation with the DOM by making them invisible by default, showing them all, and then removing each one except the correct element.

maryia-deriv: @akmal-binary Thank you. I have removed a redundant .RealAccAuth className according to your suggestion)

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -16,7 +16,7 @@ const getPropertyValue = require('../../../../../_common/utility').getPropertyVa
 const PersonalDetails = (() => {
     const form_id                = '#frmPersonalDetails';
```
akmal-binary, line 17:
```suggestion
    const form_id              = '#frmPersonalDetails';
```

akmal-binary: Can we align all equal signs on top of each other like it was before? In binary, we do this for better code readability across all variable assignments. The number of spaces is based on the longest variable name in the group, the one you changed previously in this case.

src/javascript/app/pages/user/account/settings/personal_details.js
```diff
@@ -357,7 +371,7 @@ const PersonalDetails = (() => {
                     $.scrollTo($('h1#heading'), 500, { offset: -10 });
                     $(form_id).setVisibility(0);
                     $('#missing_details_notice').setVisibility(0);
-                    $('.rowCustomerSupport').setVisibility(0);
+                    $(row_customer_support).setVisibility(0);
```
akmal-binary, line 374: Nit: we could stick to the previous version too since we just access this variable once. It's not an issue but if we directly use classname here, we reduce the code by one line.

maryia-deriv: @akmal-binary Agree) Done.

src/javascript/app/pages/user/account/settings/personal_details.js
maryia-deriv: @akmal-binary That's weird, I didn't touch this line at all.. Prettifier did it perhaps. Will change it back now.

akmal-binary (dismissed):

mustofa-binary (approved):

PR #6753, opened 2021-08-24
Mahdiyeh/ Fix: guide on 4th step

mahdiyeh-fs:


mahdiaryayi-fs (approved):

src/javascript/app/common/guide.js
```diff
@@ -87,8 +89,13 @@ const Guide = (() => {
     const setEvents = () => {
         $(`${opt.guideBtnID} strong`).click(() => {
             const enjoyhint_instance = new EnjoyHint({});
+            const contractList = $(opt.contractList);
```
akmal-binary, line 92: Can we use snake_case for these constants? `contract_list` and `close_confirmation`

mahdiyeh-fs: since guideBtnID is in the camelCase format, i keep it the same

akmal-binary: My suggestion would be to make all three consistent like `guide_btn_id`, `contract_list_id`, and `close_confirmation_id`. Do you think this would be better?

mahdiyeh-fs: yeah, right, will handle that, thanks

akmal-binary (approved):

mahdiaryayi-fs (approved):

matin-binary (approved):

PR #6783, opened 2021-09-14
nikolai.berestevich/text/Terms-and-conditions update

nikolaiberestevich-deriv:
Terms-and-conditions'  update to avoid  customer  to use 2nd acc

### DIFF
--- a/src/templates/static/legal/tac_sections/tac.jsx
+++ b/src/templates/static/legal/tac_sections/tac.jsx
@@ -95,6 +95,23 @@ const Tac = () => (
         <p>{it.L('If the Company (in its sole discretion) is not satisfied that a client is complying with these terms, or if a client does not comply with a reasonable request for information made by the Company, the Company may cancel the client\'s account and withhold any funds accumulated therein.')}</p>
 
         <h2 data-anchor='fraud-and-money-laundering'>J. {it.L('Fraud and money laundering')}</h2>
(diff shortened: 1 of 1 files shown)


src/javascript/app/pages/trade/markets.jsx
```diff
@@ -79,7 +79,8 @@ class Markets extends React.Component {
             underlying_symbol = Object.keys(this.markets[market_symbol].submarkets[submarket].symbols).sort()[0];
         }
         const market_arr = Object.entries(this.markets).sort((a, b) => sortSubmarket(a[0], b[0]));
-        const market_obj = Object.fromEntries(market_arr);
+        // market_arr.reduce works as Object.fromEntries(market_arr) but some browsers don't support Object.fromEntries => trackjs errors
+        const market_obj = market_arr.reduce((acc, market) => ({ ...acc, [market[0]]: market[1] }),{});
```
mustofa-binary, line 83: .reduce is also a new es6 solution which probably won't work on some browsers

mustofa-binary: it is better to make the change in another PR

mustofa-binary (changes requested):

mustofa-binary (approved):

akmal-binary (approved):

PR #6785, opened 2021-09-14
Hubert / HOTFIX for deposit error message

hubert-deriv:
Initially, the error message was being created through the validation functionality in metatrader config. In this validation it is not possible to inject HTML in a string as it seems that this string gets spit out as text only. So here it is not possible to implement an href in the string. In order to put an href in this error message, I put the error message in the metatrader.jsx file and then made some functionality that displays the error if the user doesn't have enough funds.

### DIFF
--- a/src/javascript/app/pages/user/metatrader/metatrader.config.js
+++ b/src/javascript/app/pages/user/metatrader/metatrader.config.js
@@ -446,9 +446,7 @@ const MetaTraderConfig = (() => {
                             const deposit_input_value   = document.querySelector('#txt_amount_deposit').value;
                             const min_req_balance = Currency.getTransferLimits(Client.get('currency'), 'min', 'mt5');
 
-                            const is_balance_more_than_min_req = +deposit_input_value > +min_req_balance;
-
-                            return is_balance_more_than_min_req;
+                            return +deposit_input_value > +min_req_balance;
                         },
                         message: localize('Should be more than [_1]', Currency.getTransferLimits(Client.get('currency'), 'min', 'mt5')),
                     }],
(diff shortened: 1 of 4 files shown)


src/javascript/app/pages/user/metatrader/metatrader.config.js
```diff
@@ -440,28 +439,16 @@ const MetaTraderConfig = (() => {
...
                         func: () => {
-                            const balance         = Client.get('balance');
+                            const deposit_input_value   = document.querySelector('#txt_amount_deposit').value;
                             const min_req_balance = Currency.getTransferLimits(Client.get('currency'), 'min', 'mt5');
 
-                            const is_balance_more_than_min_req = +balance >= +min_req_balance;
+                            const is_balance_more_than_min_req = +deposit_input_value > +min_req_balance;
 
-                            return balance && is_balance_more_than_min_req;
+                            return is_balance_more_than_min_req;
```
reza-fs, line 451:
```js
return +deposit_input_value > +min_req_balance;
```

src/sass/app/metatrader.scss
```diff
@@ -340,6 +340,12 @@
                     font-weight: bold;
                     color: $COLOR_GRAY_SHADE;
                 }
+                #insufficient_funds {
+                    a {
```
reza-fs, line 344: possible to add a class name to avoid using general css statements on this?

src/javascript/app/pages/user/metatrader/metatrader.config.js
hubert-deriv: Fixed

src/sass/app/metatrader.scss
hubert-deriv: added a class for the anchor now

reza-fs (approved):

mustofa-binary (approved):

PR #6802, opened 2021-09-28
Mahdiyeh/Add: sonarcloud code coverage

mahdiyeh-fs:


.gitignore
```diff
@@ -9,3 +9,4 @@ scripts/CNAME
 .DS_Store
 *.swp
 npm-debug.log
+coverage
```
reza-fs, line 12: Possible to add new line here?

package.json
```diff
@@ -137,5 +138,27 @@
...
+    "reporter": [
+      "text-summary",
+      "lcov"
+    ],
+    "temp-directory": "./coverage/.nyc_output",
+    "all": true,
+    "exclude": "typings",
+    "extension": [
+      ".js",
+      ".ts",
```
reza-fs, line 152: We don't need `.ts` and `.tsx` as they are not supported yet in binary.

package.json
```diff
@@ -137,5 +138,27 @@
...
+      ".jsx",
+      ".tsx"
+    ],
+    "sourceMap": true,
+    "instrument": true,
+    "include": [
+      "src/javascript/**/*.{js, ts, jsx, tsx}",
+      "scripts/**/*.{js, ts, jsx, tsx}"
```
reza-fs, line 160: here as well

package.json
```diff
@@ -137,5 +138,27 @@
...
+    ],
+    "cache": false
```
reza-fs, line 162: (question) what does this value stands for?

.gitignore
mahdiyeh-fs: sure

package.json
mahdiyeh-fs: sure, it can be added whenever we migrate to typescript

PR #6814, opened 2021-10-01
 Nikita/Improvement :  No message is displayed for change in currency to switch the account if the user is currently in crypto account . 

nikita-deriv:


src/javascript/_common/base/client_base.js
```diff
@@ -386,8 +386,8 @@ const ClientBase = (() => {
 
     const hasSvgAccount = () => !!(getAllLoginids().find(loginid => /^CR/.test(loginid)));
 
-    const canChangeCurrency = (statement, mt5_login_list, is_current = true) => {
-        const currency             = get('currency');
+    const canChangeCurrency = (statement, mt5_login_list, loginid, is_current = true) => {
```
reza-fs, line 389: Please also update all the users of this function, I've seen this is being used in `metatrader.js` file

nikita-deriv: I updated function by adding `loginid`  as possible argument. If there is no `loginid` its not ruined the logic of using this function.

reza-fs (approved):

reza-fs (approved):

PR #6871, opened 2021-10-11
Pavel-Latyshou/Account creation form for MF still using Financial Account term

pavel-latyshou-deriv:


src/javascript/app/pages/user/new_account/real_account_opening.js
```diff
@@ -104,7 +104,7 @@ const RealAccountOpening = (() => {
     const setPageTitle = () => {
         getElementById('page_title').innerHTML =
             real_account_signup_target === 'maltainvest'
-                ? localize('Financial Account Opening')
+                ? localize('Real money account opening')
```
kevinw-binary, line 107: If it's the same then I think we don't need the `if else`

however we need to double confirm with the QA whom raised the issue

PR #6895, opened 2021-10-13
Mahdiyeh/ Fix:remove MT5 password option from change password page when the passwor…

mahdiyeh-fs:
.

### DIFF
--- a/src/javascript/app/pages/user/account/change_password.js
+++ b/src/javascript/app/pages/user/account/change_password.js
@@ -125,8 +125,8 @@ const ChangePassword = (() => {
         $change_password_loading.setVisibility(1);
 
         BinarySocket.wait('get_account_status').then((response) => {
-            const account_status = response.get_account_status.status;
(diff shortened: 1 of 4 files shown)


src/javascript/app/pages/user/account/change_password.js
```diff
@@ -119,14 +120,19 @@ const ChangePassword = (() => {
...
         forgot_binary_pw_btn_id         = '#forgot_binary_pw_btn';
         forgot_trading_pw_btn_id        = '#forgot_trading_pw_btn';
+        $frm_trading_password_id        = $('#frm_trading_password');
 
         $change_password_loading.setVisibility(1);
 
-        BinarySocket.wait('get_account_status').then(() => {
+        BinarySocket.wait('get_account_status').then((response) => {
+            const account_status = response.get_account_status.status;
+            const has_social_signup      = hasSocialSignup();
```
matin-binary, line 129: let's settle the extra spacing here 😄

PR #6946, opened 2021-11-09
Al-amin/Update/banner for redirection notification in binary for eu uk clients

alamin-deriv:


src/javascript/app/common/deriv_banner.js
```diff
@@ -11,29 +13,34 @@ const DerivBanner = (() => {
...
-            let affiliate_token;
+        BinarySocket.wait('authorize', 'website_status', 'landing_company').then(() => {
 
-            if (affiliate_cookie) affiliate_token = affiliate_cookie.t;
-            else {
-                const queryString = window.location.search;
-                const urlParams = new URLSearchParams(queryString);
-                affiliate_token = urlParams.get('t');
-            }
+            if (!isEuCountry()) {
```
reza-fs, line 18: Is it possible to reverse this condition? So instead of wrapping everything inside `if (!isEuCountry())`, we could have something like: 
```js
if (isEuCountry()) {
    return;
}

if (!is_deriv_banner_dismissed) {
...
```

This'll help reducing the indentation and possibly help with cognitive complexity of the logic.

alamin-deriv: Yeah sure, thanks

reza-fs (dismissed):

kevinw-binary (approved):

PR #6953, opened 2021-11-12
nikolai.berestevix/fix/mx , mf and Belgium text error

nikolaiberestevich-deriv:
Error message for mlt,mf acc and Belgium  is changed to  "Unfortunately, trading options isn't possible in your country"


src/javascript/app/pages/trade/process.js
```diff
@@ -22,15 +22,19 @@ const getVisibleElement = require('../../../_common/common_functions').getVisibl
...
 const Process = (() => {
+    const mlt_fx_countries_list = ['au','lv','bg','lt','hr','cy','cz','nl','dk','pl','ee','pt','fi','ro','sk','si','hu','se','ie','be'];
     /*
      * This function process the active symbols to get markets
      * and underlying list
      */
     const processActiveSymbols = () => {
         BinarySocket.send({ active_symbols: 'brief' }).then((response) => {
-            if (response.active_symbols && response.active_symbols.length) {
+            if ((ClientBase.get('landing_company_shortcode') === 'malta' || 'maltainvest') || (mlt_fx_countries_list.indexOf(ClientBase.get('landing_company_shortcode')) > -1)){
```
mamali-fs, line 35: This `if` condition doesn't look right at all, 'maltainvest' is not going to be compared with `ClientBase.get('landing_company_shortcode')`. 
Also, landing_company_shortcode cannot be compared to `mlt_fx_countries_list` because it only returns values such as `svg`, `malta`, `iom`, etc.

mamali-fs (changes requested):

nikolaiberestevich-deriv (commented):
I changed check

habib-deriv (approved):
LGTM

PR #6967, opened 2021-11-18
hirad/fix--IOM game close banner and popup will both be shown

hirad-rewok:


src/javascript/app/common/game_close_popup.js
```diff
@@ -25,7 +25,7 @@ const ClosePopup = (() => {
         BinarySocket.wait('authorize', 'website_status', 'landing_company').then(() => {
             const is_uk_residence = (Client.get('residence') === 'gb' || State.getResponse('website_status.clients_country') === 'gb');
             const is_iom_client = (Client.get('residence') === 'im' || State.getResponse('website_status.clients_country') === 'im');
-            const client_account = Client.get('landing_company_shortcode') === 'malta' || Client.get('landing_company_shortcode') === 'iom';
```
matin-binary, line 28: Do we need to add another condition instead of replacing `malta`?

hirad-rewok: We are showing banner and the pop up for VRTC account of IOM, as the landing_company_shortcode that I get from BE is virtual, I changed the condition to virtual. I do not think if the landing_company_shortcode for demo accounts will ever be malta. Considering these conditions, Should I add another condition or not ?

matin-binary (approved):

matin-binary (approved):

matin-binary (dismissed):

matin-binary (dismissed):

matin-binary (dismissed):

matin-binary (approved):

PR #6972, opened 2021-11-22
Prince/ Deriv static landing page html to  binary

prince-deriv:


yashim-deriv:
I think we can exclude the src map from the files 🙏🏼

yashim-deriv (approved):
LGTM

kevinw-binary (dismissed):

PR #7009, opened 2021-12-02
Nuri/hotfix - Remove date on move-to-deriv redirection banner

njazuli-deriv:
Remove date on move-to-deriv redirection banner


yashim-deriv (approved):
LGTM

PR #7028, opened 2021-12-19
Amina/ style fix on popup

amina-deriv:


src/sass/app/websocket_pages.scss
```diff
@@ -325,6 +339,13 @@
                 .button span {
                     width: calc(50% - 55px);
                 }
+                #btn_cancel, #btn_ok {
+                    width: calc(100% - 25px);
+
+                    span {
```
hojjat-fs, line 345: can we use a class name selector instead of a tag name selector for efficiency?

https://css-tricks.com/efficiently-rendering-css/

amina-deriv: Done. Thanks for the suggestion @hojjat-fs

hojjat-fs (dismissed):

hojjat-fs (approved):

PR #7034, opened 2022-01-03
Maryia/Fix: to fit POI field label text (in RU testable at staging) into 1 line

maryia-deriv:


src/sass/_common/common.scss
```diff
@@ -218,6 +218,9 @@ label {
         display: block;
         text-align: right;
     }
+    .fields > & > div:first-child {
```
akmal-binary, line 221: Is there any way we can omit `&` from this selector or simplify it? It is hardly readable at the moment.

---
For your reference: https://css-tricks.com/the-sass-ampersand/#using-the-with-and

maryia-deriv: @akmal-binary Unfortunately I can't totally omit `&` in this selector because I need to apply the style to specifically the first child divs of .form-row (not of .fields) which in turn should be a child of .fields.
I can make it more readable though 🙂

akmal-binary: Thanks @maryia-deriv , but what I see now is that it will target this element: `.form-row .fields .form-row div:first-child` since you are putting this whole selector inside `.form-row` parent class. Is that what you're trying to achieve? It looks kinda weird to me because it repeats `form-row` class twice inside. I think you can either put the whole style outside of the parent selector or just target `div:first-child` if it won't effect any other cases.

maryia-deriv: @akmal-binary Unfortunately it will affect other cases if I target div:first-child only. The reason why I put it inside .form-row parent class is that I target div which is inside it but its selector points out that its .form-row parent class must be a child of .fields that's why this solution works.
I agree with you however that it's better to put it outside to avoid any confusion here.

mahdiyeh-fs: suppose it should be `fields .form-row div:first-child`

maryia-deriv: @mahdiyeh-fs yeah, it's already what I have)

akmal-binary: @maryia-deriv, if I understand correctly now, what you're trying to generate in the end would be `.fields .form-row div:first-child` right? In that case you can just modify the current style like: 
```
.form-row {
    padding: 10px 0;
    .required_field_label {
        font-size: 13px;
    }
    div:first-child {

       .fields & {
            // YOUR STYLE HERE
       }

       label {
        display: block;
        text-align: right;
       }
      }
}
```

akmal-binary: Would that be sufficient for your case?

maryia-deriv: @akmal-binary Thank you for your suggestion) 
Unfortunately this solution doesn't remove the possible confusion that you initially intended to avoid saying that if we put .fields inside .form-row, someone might be confused and consider that .fields is a child of .form-row.
So as I can see, your suggestion is basically has the same readability as my previous one:
```
.form-row {
    padding: 10px 0;
    .required_field_label {
        font-size: 13px;
    }
    div:first-child label {
        display: block;
        text-align: right;
    }
    .fields .form-row div:first-child {
        padding: 0; // needed to fit in text in Russian
    }
}
```
with the only difference that mine has less lines.

What do you think?)

akmal-binary: @maryia-deriv I don't see any confusion regarding your previous solution you shared above, the only thing is it will simply not work because it will generate the mentioned selector `.form-row .fields .form-row div:first-child` which will not select any element on the page by right 🙂 What I'm saying is since we use SCSS, we should preserve the Sass structure, which as I understand from the expected classname in your case will be achieved in my suggested example. We can also proceed with the current version since the selector is outside `.form-row` which is fine, but the sass structure and flexibility are ignored in this case.

akmal-binary: You can use this converter to see the difference in the output css: https://jsonformatter.org/scss-to-css

maryia-deriv: @akmal-binary You are absolutely right. I guess the fact that my code actually works even despite the css is compiled not in the right way made me miss this.. Thank you for checking!)

mahdiyeh-fs (dismissed):

akmal-binary (dismissed):

matin-binary (approved):

yauheni-deriv (approved):

maryia-deriv:
duplicate of https://github.com/binary-com/binary-static/pull/7070

PR #7055, opened 2022-02-11
Yauheni/update TnC for payment agents

yauheni-deriv:
update T&Cs for Payment Agents according to attached pdf file in task #52818


hubert-deriv (approved):

hubert-deriv:
Accidentally approved with my private account 'bert96' aswell. Can ignore that.

hojjat-fs (approved):

PR #7065, opened 2022-03-03
yauheni/ API token scope update

yauheni-deriv:
The token scope definition is made more prominent in accordance to task #57518


src/sass/_common/common.scss
```diff
@@ -1,1130 +1,1315 @@
 /* DO NOT include any rules for specific pages in this file. */
 /* theme elements */
 
-html, body {
```
hubert-deriv, line 4: I think your linter accidentally differently formatted this CSS file, I believe this should somehow be reverted, otherwise I believe it can cause conflicts for other developers.

src/templates/_common/components/forms.jsx
```diff
@@ -135,10 +136,21 @@ export const FormRow = ({
             return 'gr-12';
         } else if (has_geovalidator) {
             return 'gr-3 gr-12-m';
+        } else if (label_row_cell_width) {
+            return `gr-${12 - label_row_cell_width} gr-12-m ${sub_row_class || ''}`;
         }
         return `gr-8 gr-12-m ${sub_row_class || ''}`;
     };
 
+    const getLebelRowClassName = () => {
```
hubert-deriv, line 145: I think you did small typo here with the word `Lebel`, it should be `Label`

src/templates/_common/components/forms.jsx
```diff
@@ -155,7 +167,7 @@ export const FormRow = ({
             className={`gr-row form-row center-text-m ${is_two_rows ? 'two-rows' : ''} ${row_class || ''}`}
             id={row_id}
         >
-            <div className={`${is_two_rows ? `gr-12 ${label_row_class}` : `gr-4 gr-12-m ${label_row_class}`}`}>
+            <div className={getLebelRowClassName()}>
```
hubert-deriv, line 170: Same here, make sure you change this function aswell from `Lebel` to `Label`

src/templates/_common/components/forms.jsx
```diff
@@ -227,10 +240,20 @@ export const SubmitButton = ({
...
+                return `gr-${12 - left_alignment_size} gr-push-${left_alignment_size} gr-12-m gr-push-0-m`;
+            }
+            return '';
+        };
+
         return (
             <div className={`${is_centered ? 'center-text' : 'gr-row'} ${className || ''}`} id={id}>
                 { !is_centered ?
-                    <div className={`${!is_left_align ? 'gr-8 gr-push-4 gr-12-m gr-push-0-m' : ''} center-text-m`}>
+                    <div className={`${getClassName()} center-text-m`}>
```
hubert-deriv, line 256: I wanted to ask, what does this function exactly do? When I look in the file, I see this function nowhere imported/passed or created. How does it work?

hubert-deriv (changes requested):

src/templates/_common/components/forms.jsx
yauheni-deriv: > 

Thanks for your notes. I think both names are suitable, because I created a new attribute "label_row_cell_width" which is passed to parent component and used into this new function. I think this function is not only for setting width but also for generating common classes. How do you think ?

hubert-deriv: > 

Sure! sounds good to me

hubert-deriv (approved):

matin-binary (changes requested):
- Whole `scope-item` needs to be clickable, please help to change.
- `Admin` scope is not clickable
- `Scopes` in the table at the bottom are missing

yauheni-deriv:
> * Whole `scope-item` needs to be clickable, please help to change.
> * `Admin` scope is not clickable
> * `Scopes` in the table at the bottom are missing

src/sass/_common/common.scss
```diff
@@ -947,33 +947,142 @@ label + input {
...
+
+        label {
+            margin: 0 0 0 13px;
+            text-align: left;
+        }
+    }
+    .scope-types {
+        display: flex;
+        flex-direction: row;
+        flex-wrap: wrap;
```
hubert-deriv, line 980: You can use grid here to more easily divide elements in an equal size into 2 columns. Then you don't have to hardcode width and height on your scope items (see comment I made further below).

You could do: 
`
display: grid;
`
`grid-template-columns: 1fr 1fr;`

hubert-deriv (approved):

matin-binary (approved):

PR #7068, opened 2022-03-09
Amina/show-poa-suspect-reasons-on-onfido-suspect-status

amina-deriv:
On POI suspected status - Display the reasons for the POI rejection as per API

The requirement was on POI-suspect status,  on Deriv app we display the reasons for the rejection. But in Binary.com , we display the message the 'POI verification failed.' We need to show the reasons for rejection as in Deriv app.


src/javascript/app/pages/user/account/authenticate.js
```diff
@@ -1370,8 +1370,6 @@ const Authenticate = (() => {
                 showCTAButton('document', 'pending');
                 break;
             case 'suspected':
-                $('#unverified').setVisibility(1);
-                break;
             case 'rejected':
                 if (Number(submissions_left) < 1) {
                     $('#limited_poi').setVisibility(1);
```
amina-deriv, line 1375: Screen Shot 2022-03-09 at 5 51 09 PM

In Deriv App , we have same logic for poi "suspected" and poi "rejected" cases. we are displaying the rejection reasons in both status. so making the same logic here in Binary

yashim-deriv (approved):
LGTM

matin-binary (approved):

PR #7097, opened 2022-04-12
MaryiaF/fix: Cannot read properties of null (reading 'offsetWidth')

maryiafrantsava-binary:


src/javascript/app/pages/trade/digit_trade.js
```diff
@@ -133,13 +133,13 @@ const DigitDisplay = (() => {
...
 
     const init = (id_render, proposal_open_contract) => {
-        initTable(id_render, calculateTableHeight(proposal_open_contract), proposal_open_contract);
-        DigitTicker.init(
-            'digit_ticker_container',
-            proposal_open_contract.contract_type,
-            proposal_open_contract.shortcode,
-            proposal_open_contract.tick_count,
-            proposal_open_contract.status
+        initTable(id_render, calculateTableHeight(proposal_open_contract), proposal_open_contract).then(
```
maryiafrantsava-binary, line 136: The error occurs because we are accessing a property of a tree element that is not yet in the tree. Presumably, the function of creating an element is processed later (initTable()), and the function that accesses the element tree is processed earlier (init()).

matin-binary (approved):

PR #7108, opened 2022-04-26
MaryiaF/refactor: subscribe proposal_open_contract globally without s…

maryiafrantsava-binary:
…pecific contract_id, use same approach as in Deriv

During open contact mode runtime subscribe request happens with proposal_open_contract globally, without specific contract_id.

### DIFF
--- a/src/javascript/_common/base/socket_base.js
+++ b/src/javascript/_common/base/socket_base.js
@@ -180,9 +180,6 @@ const BinarySocketBase = (() => {
 
         if (isReady() && !availability.is_down && config.isOnline()) {
             is_disconnect_called = false;
-            if (!getPropertyValue(data, 'passthrough') && !getPropertyValue(data, 'verify_email')) {
-                data.passthrough = {};
-            }
 
             binary_socket.send(JSON.stringify(data));
             config.wsEvent('send');
--- a/src/javascript/app/pages/trade/purchase.js
+++ b/src/javascript/app/pages/trade/purchase.js
@@ -227,7 +227,7 @@ const Purchase = (() => {
                 proposal_open_contract: 1,
                 subscribe             : 1,
             };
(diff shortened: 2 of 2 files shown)


src/javascript/_common/base/socket_base.js
```diff
@@ -180,9 +180,6 @@ const BinarySocketBase = (() => {
 
         if (isReady() && !availability.is_down && config.isOnline()) {
             is_disconnect_called = false;
-            if (!getPropertyValue(data, 'passthrough') && !getPropertyValue(data, 'verify_email')) {
-                data.passthrough = {};
```
maryiafrantsava-binary, line 184: I suppose we don’t need to pass an empty object in the request, as in Deriv-app

akmal-binary (approved):

matin-binary (approved):

src/javascript/app/pages/trade/purchase.js
```diff
@@ -225,7 +225,6 @@ const Purchase = (() => {
         if (show_chart) {
             const request = {
                 proposal_open_contract: 1,
-                contract_id           : receipt.contract_id,
```
balakrishna-binary, line 228: global subscription should be made only one time. It should not be done for each purchase of contract

maryiafrantsava-binary: @balakrishna-binary thank you, let me fix it👌

maryiafrantsava-binary: @balakrishna-binary I've fixed

matin-binary:
Closing this PR as the changes are not needed. 
`proposal_open_contract ` needs to be global only in Portfolio page which is already being handled.
On dashboard, there is only a single purchase at a time so no need for global `proposal_open_contract`.

PR #7113, opened 2022-05-10
fix: update cfd retail percentage

mitra-deriv:


habib-deriv (dismissed):

munsei-binary:
why update  % of clients losing money on CFDs need 2462 files changes?

prince-deriv:
> why update % of clients losing money on CFDs need 2462 files changes? @mitra-deriv

@munsei-binary  because the design of the landing page came from the build files from deriv.com it's been a norm for months now.

PR #7141, opened 2022-05-25
Hirad/ TrackJs null tick_display_value error

hirad-rewok:


src/javascript/app/pages/trade/tick_trade.js
```diff
@@ -69,9 +69,10 @@ const TickDisplay = (() => {
                 title        : show_contract_result ? '' : contract.display_name,
                 tooltip      : {
                     formatter() {
-                        const tick = contract.tick_stream.find((data) => data.tick === this.y).tick_display_value;
+                        const tick = contract.tick_stream ? contract.tick_stream.find((data) => data.tick === this.y) : '';
```
mahdiyeh-fs, line 72: assigning string to the tick is redundant since we will check the `tick` to have value in the next line, also the tick type is object not a string.

hirad-rewok: Removed the redundant condition.

mahdiyeh-fs (approved):

matin-binary (approved):

PR #7171, opened 2022-08-02
Minor edit 4

thibault-deriv:
For testing "job execution unauthorized: block-unregistered-user"

For this purpose, commit as someone else (account having problems to start CircleCI jobs).


thibault-deriv:
Author and committer are habib-deriv but triggered by me at the end so no problem CircleCI job is running in org and status is reported and green.

trigger

PR #7178, opened 2022-08-04
Suisin/Update Error Message for new account

suisin-deriv:


src/javascript/app/common/account_opening.js
```diff
@@ -77,9 +77,11 @@ const AccountOpening = (() => {
     const showResponseError = (response) => {
         getElementById('loading').setVisibility(0);
         getElementById('real_account_wrapper').setVisibility(1);
+        const error_message = response.error.message;
+        const client_error_message = response.error.code.message_to_client;
```
matin-binary, line 81: it is better to check if `message_to_client` exists before accessing it.

src/javascript/app/common/account_opening.js
```diff
@@ -293,9 +295,10 @@ const AccountOpening = (() => {
     const handleNewAccount = (response, message_type) => {
         if (response.error) {
             const error_message = response.error.message;
+            const client_error_message = response.error.code.message_to_client;
```
matin-binary, line 298: same here

src/javascript/app/common/account_opening.js
suisin-deriv: orite 👌

src/javascript/app/common/account_opening.js
```diff
@@ -77,9 +77,11 @@ const AccountOpening = (() => {
     const showResponseError = (response) => {
         getElementById('loading').setVisibility(0);
         getElementById('real_account_wrapper').setVisibility(1);
+        const error_message = response.error.message;
+        const client_error_message = response.error.code.message_to_client;
         const $notice_box = $('#client_message').find('.notice-msg');
         $('#submit-message').empty();
-        $notice_box.text(response.msg_type === 'sanity_check' ? localize('There was some invalid character in an input field.') : response.error.message).end()
+        $notice_box.text(response.msg_type === 'sanity_check' ? localize('There was some invalid character in an input field.') : error_message === null ? client_error_message : error_message).end()
```
yauheni-kryzhyk-deriv, line 84: I'd like to suggest you for better readability to move this code  `error_message === null ? client_error_message : error_message` into new variable.
I believe  It is not good practice  to use ternary operator more then once in one place in terms of readability.

and the other thing is: can we use here just these expression: `(error_message || client_error_message)` ?

matin-binary (approved):

yauheni-kryzhyk-deriv (approved):

PR #7206, opened 2022-08-23
Meenu/updated url for binary to deriv row users

meenakshi-deriv:


src/javascript/app/common/redirect_banner.js
```diff
@@ -12,14 +13,30 @@ const RedirectBanner = (() => {
...
+                handleRedirectROW();
             }
             
         });
 
     };
 
     const handleRedirect = () => {
-        window.location.href = '/move-to-deriv/';
+        window.location.href = 'https://www.binary.bot/movetoderiv.html';
```
habib-deriv, line 25: /move-to-deriv/

mitra-deriv: Hi @habib-deriv @meenakshi-deriv the new one is  `/binary-to-deriv/` which we need to append to the url. We won’t need any redirection I think. 🤔 

`binary.com/binary-to-deriv/`

src/javascript/app/common/redirect_banner.js
```diff
@@ -12,14 +13,30 @@ const RedirectBanner = (() => {
...
+    };
+    const handleRedirectROW = () => {
+
+        // Check if param have ?binary-com-lp
+        const redirectBinary = new URLSearchParams(window.location.search);
+
+        if (redirectBinary.has('binary-com-lp') || Cookies.get('binary-com-show')){
```
mitra-deriv, line 32: I think we don’t need this `binary-com-lp` check here anymore cause the changes will be in binary itself now.
We just need to append `/binary-to-deriv/` once the client country is rest of the world.

@habib-deriv please correct me if I'm wrong 🙏

PR #7220, opened 2022-09-06
Cloudflare pages: Fix nasty redirection and regex (2)

thibault-deriv:
This is a fixed version of #7189 that was reverted in #7214 

The order of regex substitution was not good, and a `.` in the query string (`amount_crypto=0.00015`) was confusing the logic:
e.g. : `https://staging.binary.com/en/trading.html?currency=USD&...&amount_crypto=0.00015`

# Changes
- We are hosting backups on Cloudflare pages. Cloudflare pages webserver is rewriting URLs to remove `.html` extensions while still flexibly serving the html file hence has a very little impact in general. Forcing `.html` suffix is producing an infinite loop. And this feature seems not needed in general for routing/redirect logic, just a convenience.

The QA was done on a derived version of this pull request via URL: [test.cf-pages-binary-static.binary.com](https://test.cf-pages-binary-static.binary.com/)


habib-deriv (approved):
LGTM

PR #7278, opened 2022-10-27
thisyahlen/fix view disabled button

thisyahlen-deriv:


src/javascript/app/pages/user/view_popup/view_popup.js
```diff
@@ -56,6 +56,11 @@ const ViewPopup = (() => {
         }
 
         getContract();
+
```
mahdiyeh-fs, line 59: there is a same if statement above, why we need to disable it and then enable it again? doesn't removing above statement fix the issue?

thisyahlen-deriv: True also, but it disables the button at the back of the statement. Then it should enable back with the cleanup.

PR #7299, opened 2022-12-13
[FE Deriv.com squad] NikitK/ Redirect Binary affiliates & IB page to Deriv.com affiliates & IB page

NikitK-deriv:
redirect from binary to deriv affiliate page

### DIFF
--- a/src/javascript/static/pages/affiliate_ib_landing.js
+++ b/src/javascript/static/pages/affiliate_ib_landing.js
@@ -1,8 +1,11 @@
 const Url         = require('../../../javascript/_common/url.js');
 const TabSelector = require('../../_common/tab_selector');
+const getLanguage = require('../../_common/language.js').get;
 
 const AffiliateIBLanding = (() => {
+    const current_language = getLanguage().toLowerCase();
     const onLoad = () => {
+        window.location.replace(`https://deriv.com/${current_language}/partners/affiliate-ib/`);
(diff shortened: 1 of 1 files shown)


src/javascript/static/pages/affiliate_ib_landing.js
```diff
@@ -1,8 +1,11 @@
 const Url         = require('../../../javascript/_common/url.js');
 const TabSelector = require('../../_common/tab_selector');
+const getLanguage = require('../../_common/language.js').get;
 
 const AffiliateIBLanding = (() => {
+    const current_language = getLanguage().toLowerCase();
```
mitra-deriv, line 6: In binary Chinese and traditional Chinese having ` _ ` but in deriv.com we have `-` i.e `zh-cn` in deriv but `zh_cn` so we need to also replace that. 🙏 

```suggestion
    const current_language = getLanguage().toLowerCase().replace(/_/g, '-');
```

mitra-deriv (changes requested):

src/javascript/static/pages/affiliate_ib_landing.js
NikitK-deriv: Goot catch, thanks @mitra-deriv 🙏

mitra-deriv (approved):

habib-deriv (approved):
LGTM

PR #7312, opened 2023-01-06
fix: Change Rise/Fall contract from displaying Higher/Lower

henry-deriv:


hirad-rewok (approved):

src/javascript/app/pages/user/view_popup/view_popup.js
```diff
@@ -120,24 +121,15 @@ const ViewPopup = (() => {
...
             MULTDOWN    : localize('Multiplier Down'),
-        });
-
-        return {
-            get: () => {
-                if (!contract_type_display) {
-                    contract_type_display = initContractTypeDisplay();
-                }
-                return contract_type_display;
-            },
```
matin-deriv, line 131: You need to follow the standard for the project and this is how we define the basic functionalities of each component. you need to revert this back.
Also it is breaking other places in which `get` method is being used.

henry-deriv: Same as the other pull request for DSmartTrader 🙏.

matin-deriv (approved):

hirad-rewok (approved):

matin-deriv (approved):

PR #7328, opened 2023-01-31
kate/78896/ TrackJS: a.markets.forex is undefined

kate-deriv:
The issue comes from Japan accounts (and only one from Norway): `this.markets` (line 126) is an object which has NO forex key, the only available is synthetics (see the screenshot of console in the card). So, the simple check with `if ` for undefined was implemented.

**Note 1**: I couldn't use optional chaining (`delete this.markets?.forex.submarkets?.smart_fx`) because of an error from babel. The option below also couldn't be used, as I got an error from eslint:
`this.markets.forex && delete this.markets.forex.submarkets.smart_fx;`

**Note 2**:  In local host there was an error about the missing key in `<React.Fragment>` after using map method, so I decided to fix it by adding an index.


maryia-deriv (approved):
👍🏻

habib-deriv (approved):

maryia-matskevich-deriv (approved):
LGTM

maryia-deriv (approved):

